PR #54, opened 2019-10-03
URL to image was broken

Swader:

PR #52, opened 2019-10-02
"Learn" section update

laboon:
Results of review and updating the "Learn" section of the wiki.

This includes fixing typos, updating some things that we have clarified with Kusama, adding Kusama links, clarifying a few examples, adding some words/phrases to the glossary... lots of minor things.  

If anyone has a suggestion for how we can better organize and categorize the Learn section, I'd be open to it.

Note - this does NOT cover the change from Alexander's "triple-account system" to Kusama's session keys.  I am going to do that as a separate PR because it's a) it's more in-depth and thus I want to make sure we spend time reviewing it and b) this PR is already pretty big.

PR #56, opened 2019-10-03
Image src fix

Swader:


ansonla3:
I suggest using markdown ```` instead of HTML for consistency

Swader:
Let's hold off though, maybe #52 renders #54 and #56 obsolete.

laboon:
This was also overridden by #52 - the image didn't really belong on this page in the first place

PR #108, opened 2019-10-23
disable Spanish &  Korean

ansonla3:


lsaether:
Why?

ansonla3:
> Why? 

There is not much content that has been translated to Spanish & Korean. We also did a revamp which means that the contents (format) would not be consistent with the latest one especially Korean has been already translated a lot that follows the old wiki. It would require someone who understands Korean to do a format matching one by one first in Crowdin. 
We can enable it again (say 50% of the contents have been translated to a particular language)

lsaether (approved):

lsaether (changes requested):
Is there anyone in community that wants to do it?

Swader:
We could ask Melea, he's very active and a validator / nominator, and produces some blogs now and then too. If he won't be interested directly, perhaps he could find us someone who might be interested.

Swader:
Ok, let's kill this, it'll be a while before we find volunteers. Will this conflict with #181?

PR #110, opened 2019-10-24
A better page on SPREE 

### LINKED ISSUE #64: Add more SPREE information
Information on SPREE is kind of light (although since it is not yet implemented, this is to be expected).

lsaether:
closes #64


docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
 
-> This appendix would be opt-in for each parachain: parachains would be able to "tell" the runtime that they're happy to use this appendix.
+SPREE modules are fragments of logic (in concrete terms they are blobs of WebAssembly code) that are uploaded onto
+Polkadot through a governance mechanism or by parachains. Once the blob is uploaded to Polkadot, all other parachains
+can decide to opt-in to the logic. The SPREE module would retain its own storage independent of the parachain, but would
+be able to be called through an interface with the parachain. Parachains will send messages to the SPREE module synchronously.
 
-> I'm not sure about the name though :) - there's not really anything "smart" about them and they're not really "protocols". Really they're a Shared Protected Runtime Appendix is more like it...
+SPREE modules are important to the overall ICMP architecture because they give guarentee to the code that will be
+executed on destination parachains. While ICMP guarentees the delivery of a message, it does not guarentee what code
```
Swader:
```suggestion
executed on destination parachains. While ICMP guarantees the delivery of a message, it does not guarantee what code
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+will be executed, i.e. how the receiving parachain will interpret the message. While ICMP accomplishes trustless message
+passing, SPREE is the trustless interpration of the message and a key part to the usefulness of ICMP.
```
Swader:
```suggestion
passing, SPREE is the trustless interpenetration of the message and a key part to the usefulness of ICMP.
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
 
-The description was later simplified in a presentation at Tokyo DOT Day:
+SPREE modules are like recipes in cookbooks. For example, if we give an order to a cook to make a souffle, and we’re
```
Swader:
```suggestion
SPREE modules are like recipes in cookbooks. For example, if we give an order to a cook to make a soufflé, and we’re
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+decently confident in the ability of the cook, we have a vague idea of what will be made but no actually surety how it 
+will be made. However, let’s say that a cook has the “Souffle Maker’s Manual” on their bookshelf and has committed
```
Swader:
```suggestion
will be made. However, let’s say that a cook has the “Soufflé Maker’s Manual” on their bookshelf and has committed
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+themselves to only make souffles from this book. Now we can also consult the same book that the cook has, and we have a
+precise understanding of what will happen when we tell the cook to make a souffle. In this example, “make a souffle” was
```
Swader:
```suggestion
precise understanding of what will happen when we tell the cook to make a soufflé. In this example, “make a soufflé” was
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+the message in ICMP and the cookbook was the SPREE module.
 
-- Parachains can opt-in to special runtime logic fragments (like smart contracts)
-- Own storage, own ICMP endpoint
-- All instances across parachains have identical logic
-- Executes alongside parachain
-- Protected: storage can not be altered by parachain logic; messages can not be faked from them by parachains
+In concrete terms, SPREE modules could be useful for various functionally on Polkadot. One suggested use case of SPREE
```
Swader:
```suggestion
In concrete terms, SPREE modules could be useful for various functionality on Polkadot. One suggested use case of SPREE
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+the incrementing and decrementing of balances of various assets based on a unique identifier.
 
 ## Why?
 
-Sending messages across parachains in ICMP only ensures that the message will be delivered but does not specify the code that will be executed, or how the message will be interpreted by the receiving parachain. There would be ways around this such as requesting a verifiable receipt of the execution from the receiving parachain, but in the naked case the other parachain would have to be trusted. Having shared code which exist in appendices that the parachain can opt-in to supporting resolves the need for trust and makes the execution of the appendices completely trustless.
+Sending messages across parachains in ICMP only ensures that the message will be delivered but does not specify the code
+that will be executed, or how the message will be interpreted by the receiving parachain. There would be ways around
+this such as requesting a verifiable receipt of the execution from the receiving parachain, but in the naked case the
+other parachain would have to be trusted. Having shared code which exist in appendices that the parachain can opt-in to
```
Swader:
```suggestion
other parachain would have to be trusted. Having shared code which exists in appendices that the parachain can opt-in to
```

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+supporting resolves the need for trust and makes the execution of the appendices completely trustless.
```
Swader:
```suggestion
resolves the need for trust and makes the execution of the appendices completely trustless.
```

Swader: Resolves the need for trust + makes ... completely trustless = unnecessary repetition?

docs/learn-spree.md
```diff
@@ -4,36 +4,61 @@ title: SPREE
...
+- These fragments have their own storage and own ICMP endpoint.
+- All instances across parachains have identical logic.
+- It executes alongside parachain.
+- Protected: storage can not be altered by parachain logic; messages can not be faked from them by parachains.
 
-> A parachain would be able to upload a "runtime appendix" to the relay chain. It would live on the relay chain and be a WebAssembly blob, not dissimilar to a parachain's normal validation function. However, it would only ever be executed by parachain collators.
+## Origin
 
-> Storage/state and ICMP would be independent of the parachain itself. However, the parachain would be able to pass it messages synchronously for it to interpret on its own terms.
+On 28 March, 2019 u/Tawaren, a member of the Polkadot community, made a post on [r/dot](polkadot reddit) called
```
Swader:
```suggestion
On 28 March, 2019 u/Tawaren, a member of the Polkadot community, made a post on [r/dot][polkadot reddit] called
```

Swader (changes requested):
Much clearer. I would also clarify storage a bit more, and name some more examples of SPREE modules. For example, if a SPREE module has shared storage for all the parachains, that means state bloat. How does this affect validators - do all validators need to maintain all the SPREE modules, or do they only maintain those into which the parachains they are devoted to have opted into? Likewise, is there a concept of pruning / state rent in SPREE? If we have a balance SPREE module, then:

a) Where can one inspect the history of transactions affecting that SPREE module, i.e. will the current polkadot explorer support this and
b) When it grows too much, can the relay chain somehow shed this history if it's not needed? And to that effect, will all SPREE modules be permanent storage, or can some of them be transient too with "pure" functions or temporary payloads only?

lsaether:
SPREE modules shouldn't be shared storage for all parachains, but one storage root for each instance created for each parachain that opts into it. 

Forseeably the only storage that needs to be kept on the relaychain would be parachains * SPREE modules bits for whether or not a particular parachain opts into a particular SPREE module. And then a storage root for this SPREE module, but it could be possible to use a single SPREE storage root per parachain. Collators would be expected to keep data, so validators wouldn't need to do any extra work besides their normal validation duties against the SPREE module now too.

a) Yes.
b) Probably there's some limit of SPREE modules that can be active at any one time and some of them would be replaced by new ones over time. Is that what you mean by transient or with temporary payloads?

ansonla3:
would it be better to have a diagram to explain the whole operation (like run through an example) how it works to give the readers to have a better understanding in general?

Swader:
Yes, that's what I meant, being able to reliably get rid of past implementations.

@ansonla3 I like that idea. I think SPREE are going to be a big selling point that needs to be clarified, so going more in depth would be great.

lsaether:
We can go more in-depth on a subsequent PR but I think already this page is an improvement. Can you re-review @Swader ?

Swader (approved):

FatemeShirazi (approved):
I think it looks ok, but since it is not implemented yet we should indicate at the beginning of the page that this is only a tentative plan and details might change.

lsaether:
Thanks @FatemeShirazi, I will update that!

PR #120, opened 2019-10-29
Add validator upgrade instructions

joepetrowski:
Does what the title says it does.

### DIFF
--- a/docs/maintain-guides-how-to-upgrade.md
+++ b/docs/maintain-guides-how-to-upgrade.md
@@ -18,7 +18,7 @@ Session keys are stored in the client and used to sign validator operations. The
 
 ### Database
 
-Validators keep a database with all of their votes. If you start another machine with a validator on it, it is important to sync the database with your other machine or you risk equivocating.
+Validators keep a database with all of their votes. If two machines have the same Session keys but different databases, they risk equivocating. For this reason, we will generate new Session keys each time we change machines.
 
 [More info about equivocation.](learn-staking#slashing)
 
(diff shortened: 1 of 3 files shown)


lsaether (commented):
How does one use the same database? Should they just copy over the `db` directory to the new machine?

docs/maintain-guides-how-to-upgrade.md
```diff
@@ -0,0 +1,51 @@
...
+1. Submit a `set_key` extrinsic from your Controller account with your new Session keys.
+1. Take note of the Session that this extrinsic was executed in.
+
+It is imperitive that your Validator A keep running in this Session. `set_key` only takes effect in the next Session.
+
+### Session `N+1`
+
+Validator B is now acting as your validator. You can safely take Validator A offline.
+
+1. Stop Validator A.
```
andresilva, line 42: Maybe add a note here about making sure that any block in the current session `N + 1` has been finalized. Otherwise `Validator A` might still be active in the GRANDPA authority set (since it hasn't finalized the end session block and thus hasn't transitioned).

A message like:

```
2019-10-28 21:44:13 Applying authority set change scheduled at block #450092
2019-10-28 21:44:13 Applying GRANDPA set change to new set with 20 authorities
```

Should also show up in the node logs.

andresilva (approved):
overall instructions lgtm, just a comment regarding finality.

joepetrowski:
> How does one use the same database? Should they just copy over the db directory to the new machine?

Yeah exactly, in a datacenter you should be able to provision storage separate from compute. So just start a new server with the same storage resource.

andresilva:
In this case it isn't that much important to use the same database, we're waiting across a session change, nodes `A` and `B` should have different keys so there's no possibility for an equivocation. This is the safest way to handle any upgrades, just wait until the session finishes to decommission the node. If we were to do this process on the same session we would either have to:
- Have some downtime as we take the node down in order to copy the db (since otherwise the db could be rewritten with a newer vote in the meantime);
- Risk equivocating due to race conditions in the process above.

lsaether (approved):

andresilva (commented):
lgtm

PR #125, opened 2019-10-31
Added an updated page on randomness

### LINKED ISSUE #73: Randomness page needs clarifications
The randomness page needs more work, especially in comparison with Ethereum 2.0's RANDAO+VDF approach and why the quadratic compute doesn't hit them, while advertising much higher numbers on much weaker hardware in terms of nodes.

Swader:
This closes #73.


docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
 sidebar_label: Randomness
 ---
 
-## VRFs
+Randomness in Proof of Stake blockchains is important for a fair and unpredictable distribution of validator responsibilities. Because computers are deterministic (the same input always produces the same output), they are bad at random numbers. Additionally, general random numbers as found in common applications are pseudo-random as they depend on a sufficiently random _seed_ provided by the user or another type of _oracle_, like a [weather station for atmospheric noise](https://www.random.org/randomness/), [your heart rate](https://mdpi.altmetric.com/details/47574324), or even [lava lamps](https://en.wikipedia.org/wiki/Lavarand).
```
lsaether:
```suggestion
Randomness in Proof of Stake blockchains is important for a fair and unpredictable distribution of validator responsibilities. Computers are bad at random numbers because they are deterministic devices (the same input always produces the same output). Another complication is that general random numbers that are commonly used in applications are pseudo-random and they depend on a sufficiently random _seed_ provided by the user or another type of _oracle_, like a [weather station for atmospheric noise](https://www.random.org/randomness/), [your heart rate](https://mdpi.altmetric.com/details/47574324), or even [lava lamps](https://en.wikipedia.org/wiki/Lavarand).
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
...
 
-The VRF used in Polkadot is roughly the same as the one used in Ouroboros Praos. Ouroboros randomness is secure for block production and works well for BABE. However, Polkadot requires a stronger source of randomness since it would be viable for an attacker to manipulate this randomness and effect validator-parachain selection.
+There are two main approaches to blockchain randomness in production today: RANDAO and VRF. Polkadot uses VRF.
 
-There are proposed solutions that have been considered for stronger randomness, but each has its own weakness (see [the section](https://research.web3.foundation/en/latest/polkadot/BABE/sortition/) in the research wiki for a fuller explanation). For this reason, Polkadot uses randomness which is based on a VRF.
+## VRF
 
-## VDFs
+A verifiable random function is a mathematical operation which for some input produces as output a random number, and proof that this random number was generated by the submitter. This proof can be verified by any challenger to make sure the random number roll was indeed valid.
```
lsaether:
```suggestion
A verifiable random function is a mathematical operation that takes some input and produces a random number along with a proof of authenticity that this random number was generated by the submitter. The proof can be verified by any challenger to ensure the random number generation is valid.
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
...
 
-[Verifiable Delay Functions](https://vdfresearch.org/) are computations which take a prescribed duration of time to complete, even on parallel computers. They produce unique output which can independently and efficiently verified in a public setting.
+The VRF used in Polkadot is roughly the same as the one used in Ouroboros Praos. Ouroboros randomness is secure for block production and works well for BABE. Where they differ is that Polkadot's VRF does not depend on a central clock (the problem becomes - whose central clock?), rather, it depends on its own past results to determine present and future results, and it uses slot numbers as a clock emulator, estimating time. Here's how it works in detail.
```
lsaether:
```suggestion
The VRF used in Polkadot is roughly the same as the one used in Ouroboros Praos. Ouroboros randomness is secure for block production and works well for BABE. Where they differ is that Polkadot's VRF does not depend on a central clock (the problem becomes - whose central clock?), rather, it depends on its own past results to determine present and future results, and it uses slot numbers as a clock emulator, estimating time.

Here's how it works in detail:
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
...
 
-## Other approaches to randomness on blockchains
+Slots are discrete units of time six seconds in length. Each slot can contain a block, but doesn't have to. The aim is to have block times at six seconds too, so in theory every slot should have a block. Slots make up epochs - specifically, 2400 slots make one epoch, which makes epochs four hours long.
```
lsaether:
```suggestion
Slots are discrete units of time six seconds in length. Each slot can contain a block, but may not. Slots make up epochs - on Kusama, 2400 slots make one epoch, which makes epochs four hours long.
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
...
 
-- [RANDAO](https://github.com/randao/randao) - RANDAO requires each validator to prepare by performing a series of hashes on some seed. Validators then publish the final hash during a round and the random number is derived from every participant's entry into the game. As long as one honest validator participates, the randomness is considered secure (non-economically viable to attack). It was not chosen as the randomness method for the VRF used in Polkadot because revealing a hash from every block producer for each slot costs quadratic bandwidth, or at least quadratic computation.
+In every slot, each validator "rolls a die". They execute a function (the VRF) which takes as input the following:
+
+- the "secret key", a key specifically made for these die rolls and which gets regenerated in every new slot
```
lsaether:
```suggestion
- **The "secret key"**, a key specifically made for these die rolls and which gets regenerated in every new slot.
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
...
+- the hash of VRF values from the blocks in the epoch before last (N-2), so past randomness has an effect on the current pending randomness (N)
```
lsaether:
```suggestion
- **The hash of VRF values from the blocks in the epoch before last (N-2)**, so past randomness has an effect on the current pending randomness (N).
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,51 @@ title: Randomness
...
+- the slot number
```
lsaether:
```suggestion
- **The slot number.**
```

Swader:
Thank you @lsaether!

docs/learn-randomness.md
```diff
@@ -4,22 +4,53 @@ title: Randomness
 sidebar_label: Randomness
 ---
 
-## VRFs
+Randomness in Proof of Stake blockchains is important for a fair and unpredictable distribution of validator responsibilities. Computers are bad at random numbers because they are deterministic devices (the same input always produces the same output). Another complication is that general random numbers that are commonly used in applications are pseudo-random and they depend on a sufficiently random _seed_ provided by the user or another type of _oracle_, like a [weather station for atmospheric noise](https://www.random.org/randomness/), [your heart rate](https://mdpi.altmetric.com/details/47574324), or even [lava lamps](https://en.wikipedia.org/wiki/Lavarand).
```
laboon:
```suggestion
Randomness in Proof of Stake blockchains is important for a fair and unpredictable distribution of validator responsibilities. Computers are bad at random numbers because they are deterministic devices (the same input always produces the same output).   What people usually call random numbers on a computer, such as in a gaming application, are actually _pseudo-random_ - that is, they depend on a sufficiently random _seed_ provided by the user or another type of _oracle_, like a [weather station for atmospheric noise](https://www.random.org/randomness/), [your heart rate](https://mdpi.altmetric.com/details/47574324), or even [lava lamps](https://en.wikipedia.org/wiki/Lavarand), from which it can generate a series of seemingly-random numbers.  But given the same seed, the same sequence will always be generated.
```

laboon: Just wanted to clarify this a bit.

docs/learn-randomness.md
```diff
@@ -4,22 +4,53 @@ title: Randomness
...
+
+In every slot, each validator "rolls a die". They execute a function (the VRF) which takes as input the following:
+
+- **The "secret key"**, a key specifically made for these die rolls and which gets regenerated in every new slot.
+- **The hash of VRF values from the blocks in the epoch before last (N-2)**, so past randomness has an effect on the current pending randomness (N).
+- **The slot number.**
+
+![](assets/VRF_babe.png)
+
+The output is two values: a `RESULT` and a `PROOF`.
```
laboon:
```suggestion
The output is two values: a `RESULT` (the random value) and a `PROOF` (a proof that the random value was generated correctly).
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,53 @@ title: Randomness
...
+
+The `RESULT` is then compared to a _threshold_ defined in the implementation of the protocol (specifically, in the Polkadot Runtime Environment currently being built by Parity). If the value is less than the threshold, then the validator who rolled this number is a viable block production candidate for that slot. The validator then attempts to create a block and submits this block into the network along with the previously obtained `PROOF` and `RESULT`. 
+
+The fishermen - the nodes watching the network for collator and validator wrongdoing - will be verifying relay chain blocks. Since an illegal roll will generate an illegal block, and since fishermen will have access to the `RESULT` and `PROOF` in every block produced by a validator, it'll be easy for them to automatically report cheating validators.
```
laboon:
```suggestion
The fishermen - nodes watching the network for collator and validator wrongdoing - will be verifying relay chain blocks. Since an illegal roll will generate an illegal block, and since fishermen will have access to the `RESULT` and `PROOF` in every block produced by a validator, it'll be easy for them to automatically report cheating validators.
```

docs/learn-randomness.md
```diff
@@ -4,22 +4,53 @@ title: Randomness
...
+
+- **The "secret key"**, a key specifically made for these die rolls and which gets regenerated in every new slot.
+- **The hash of VRF values from the blocks in the epoch before last (N-2)**, so past randomness has an effect on the current pending randomness (N).
+- **The slot number.**
+
+![](assets/VRF_babe.png)
+
+The output is two values: a `RESULT` and a `PROOF`.
+
+The `RESULT` is then compared to a _threshold_ defined in the implementation of the protocol (specifically, in the Polkadot Runtime Environment currently being built by Parity). If the value is less than the threshold, then the validator who rolled this number is a viable block production candidate for that slot. The validator then attempts to create a block and submits this block into the network along with the previously obtained `PROOF` and `RESULT`.
```
laboon: There can - and will - be more PREs (created by e.g. Soramitsu and Chainsafe).  Maybe get rid of the Parity reference, although they are the only ones who have actually produced one so far?

laboon (changes requested):
Couple of suggestions

docs/learn-randomness.md
Swader: You and your double spaces :D

docs/learn-randomness.md
Swader:
```suggestion
The `RESULT` is then compared to a _threshold_ defined in the implementation of the protocol (specifically, in the Polkadot Runtime Environment). If the value is less than the threshold, then the validator who rolled this number is a viable block production candidate for that slot. The validator then attempts to create a block and submits this block into the network along with the previously obtained `PROOF` and `RESULT`. 
```

Swader: Reason was that I asked around where the threshold is defined and _no one_ knows, not even the Parity devs I could reach. But yeah, sounds better without it.

lsaether: https://github.com/paritytech/substrate/blob/master/core/consensus/babe/src/authorship.rs#L30

Swader: Nice!

Swader: Is this specced anywhere? How will other implementers know?

lsaether: It should be in the BABE paper / research and also in the PRE specification.

PR #126, opened 2019-11-01
Closes 85 - BABE explained with aura angle / primary-secondary block …

### LINKED ISSUE #85: When there are empty blocks, AURA kicks in. Documentation needed.
During the shuffling, some blocks have no slot leader candidates. These slots used to remain empty under BABE, but with AURA which runs in constant time to catch up with those, they are handled. This has some ICMP implications too, given that queues can get filled up if a block is skipped.

We need a document about AURA. I'll see what I can do about that once I've digested the paper unless someone is already more familiar.

Swader:
…production

Closes #85 

@joepetrowski


lsaether (approved):
LGTM

docs/learn-consensus.md
```diff
@@ -34,14 +34,32 @@ This is a way of getting the benefits of probabilistic finality (the ability to
...
 
-### GRANDPA: Finality Gadget
+### BABE
+
+BABE (Blind Assignment for Blockchain Extension) is the block production mechanism that runs between the validator nodes and determines the authors of new blocks. BABE is comparable as an algorithm to Ouroboros Praos, with some key differences in chain selection rule and slot time adjustments. BABE assigns block production slots to validators according to stake and using the Polkadot [randomness cycle](learn-randomness).
+
+Validators in Polkadot will participate in a [lottery](learn-randomness) in every slot which will tell them whether or not they are the block producer candidate for that slot. Slots are discrete units of time 6 seconds in length, but because of this randomness mechanic, it can happen that multiple validators are candidates in a slot. Other times, when no validators made a successful roll, it might happen that a slot is empty which can get in the way of Polkadot's ambitious 6-second block time. How, then, do we make sure that block time is indeed constant? How do we deal with each of these scenarios?
```
joepetrowski:
```suggestion
Validators in Polkadot will participate in a [lottery](learn-randomness) in every slot that will tell them whether or not they are the block producer candidate for that slot. Slots are discrete units of time, nominally 6 seconds in length. Because of this randomness mechanism, multiple validators could be candidates for the same slot. Other times, a slot could be empty, resulting in inconsistent block time.
```

docs/learn-consensus.md
```diff
@@ -34,14 +34,32 @@ This is a way of getting the benefits of probabilistic finality (the ability to
...
+
+#### Multiple Validators per Slot
+
+When multiple validators are block producer candidates in a given slot, both will produce a block and broadcast it to the network. At that point, it's a race. The validator whose block reaches most of the network first wins. Depending on network topology and latency, both chains will continue to build in some capacity, until finalization kicks in and amputates a fork. See Fork Choice below for how that works.
```
joepetrowski:
```suggestion
When multiple validators are block producer candidates in a given slot, all will produce a block and broadcast it to the network. At that point, it's a race. The validator whose block reaches most of the network first wins. Depending on network topology and latency, both chains will continue to build in some capacity, until finalization kicks in and amputates a fork. See Fork Choice below for how that works.
```

docs/learn-consensus.md
```diff
@@ -34,14 +34,32 @@ This is a way of getting the benefits of probabilistic finality (the ability to
...
+
+#### No Validators in Slot
+
+When no validators have rolled low enough in the randomness lottery to qualify for block production, a slot can remain seemingly blockless. We avoid this by running a secondary round robin style validator selection algorithm in the background. The validators selected to produce blocks through this algorithm always produce blocks, but these _secondary_ blocks are ignored if the same slot also produces a primary block from a [VRF-selected](learn-randomness) validator. Thus, a slot can have either a _primary_ or a _secondary_ block, and no slots are ever skipped.
```
joepetrowski:
```suggestion
When no validators have rolled low enough in the randomness lottery to qualify for block production, a slot can remain seemingly blockless. We avoid this by running a secondary, round-robin style validator selection algorithm in the background. The validators selected to produce blocks through this algorithm always produce blocks, but these _secondary_ blocks are ignored if the same slot also produces a primary block from a [VRF-selected](learn-randomness) validator. Thus, a slot can have either a _primary_ or a _secondary_ block, and no slots are ever skipped.
```

docs/learn-consensus.md
```diff
@@ -34,14 +34,32 @@ This is a way of getting the benefits of probabilistic finality (the ability to
...
+
+### GRANDPA: Finality gadget
 
 GRANDPA (GHOST-based Recursive ANcestor Deriving Prefix Agreement) is the finality gadget that is implemented for the Polkadot relay chain.
 
 It works in a partially synchronous network model as long as 2/3 of nodes are honest and can cope with 1/5 Byzantine nodes in an asynchronous setting.
 
 A notable distinction is that GRANDPA reaches agreements on chains rather than blocks, greatly speeding up the finalization process, even after long-term network partitioning or other networking failures.
 
+In other words, as soon as 2/3 or more validators attest to a block, all blocks leading up to that one are immediately finalized at once.
```
joepetrowski: "immediately" and "at once" are tautologous :)
```suggestion
In other words, as soon as more than 2/3 of validators attest to a chain containing a certain block, all blocks leading up to that one are finalized at once.
```

docs/learn-consensus.md
```diff
@@ -52,15 +70,16 @@ The [Rust implementation](https://github.com/paritytech/substrate/blob/master/sr
 
 For even more detail, see the [GRANDPA research page](http://research.web3.foundation/en/latest/polkadot/GRANDPA/) on the W3F Research pages.
 
-### BABE: Block Production
+### Fork Choice
 
-BABE (Blind Assignment for Blockchain Extension) is the block production mechanism that runs between the validator nodes and determines the authors of new blocks.
+Bringing BABE and GRANDPA together, the fork choice of Polkadot becomes clear. When there are competing forks, the following is how priority is determined for both BABE to produce new blocks on top of, and for GRANDPA to finalize:
```
joepetrowski: I'm not crazy about this section. It's not wrong per se, but I think a little misguided. One of the requirements of using GRANDPA is that whatever block production engine is building on it MUST follow the fork choice rule of GRANDPA.
```suggestion
Bringing BABE and GRANDPA together, the fork choice of Polkadot becomes clear. BABE must always build on the chain that has been finalized by GRANDPA. When there are forks after the finalized head, BABE provides probabilistic finality by building on the chain with the most primary blocks.
```

joepetrowski (commented):
Close, just a few things I would change to be more precise.

Swader:
Thank you @joepetrowski for your valuable feedback!

PR #150, opened 2019-11-10
Changed the explanation of the NTBias Diagram

wbarobinson:
I believe (and please correct me if I am wrong) the explanation that the diagram is describing a 34% nay requirement for rejection is backwards. With a negative turnout bias, the whole point is that with few people it takes a a huge percentage to reject (or vice-versa, a small amount of ayes to accept). So when there is only 25% percent turnout in a negative turnout bias situation, the ayes only need 34% of the vote to carry the council's proposal through.


lsaether:
#152

PR #152, opened 2019-11-10
Update learn-governance.md

wbarobinson:
I believe (and please correct me if I am wrong) the explanation that the diagram is describing a 34% nay requirement for rejection is backwards. With a negative turnout bias, the whole point is that with few people it takes a a huge percentage to reject (or vice-versa, a small amount of ayes to accept). So when there is only 25% percent turnout in a negative turnout bias situation, the ayes only need 34% of the vote to carry the council's proposal through.

### DIFF
--- a/docs/learn-governance.md
+++ b/docs/learn-governance.md
@@ -167,9 +167,9 @@ In contrast, when it has 75% turnout, the tally of "aye" votes has to reach 54%,
 
 Suppose there is a unanimous proposal proposed by the council, ``Negative Turnout Bias`` would be used, so that means the proposal is passed by default. Hence, more token holders have to participate in voting to prevent it from passing if they do not like this proposal.
 
-Referring to the above image, when the referenda only has 25% turnout, the tally of "nay" votes has to reach 34% for it to reject.
+Referring to the above image, when the referenda only has 25% turnout, the tally of "aye" votes has to reach 34% for it to accept.
 
-In short, when turnout rate is low, a super-majority is required to pass the proposal, which means a higher threshold of "aye" (yes) votes have to be reached, but as turnout increases towards 100%, it becomes a simple-majority.
+In short, when turnout rate is low, a super-majority is required to reject the proposal, which means a lower threshold of "aye" (yes) votes have to be reached, but as turnout increases towards 100%, it becomes a simple-majority.
 
 All three tallying mechanisms - majority carries, super-majority approve, and super-majority against - equate to a simple majority carries system at 100% turnout.
 


docs/learn-governance.md
```diff
@@ -167,9 +167,9 @@ In contrast, when it has 75% turnout, the tally of "aye" votes has to reach 54%,
 
 Suppose there is a unanimous proposal proposed by the council, ``Negative Turnout Bias`` would be used, so that means the proposal is passed by default. Hence, more token holders have to participate in voting to prevent it from passing if they do not like this proposal.
 
-Referring to the above image, when the referenda only has 25% turnout, the tally of "nay" votes has to reach 34% for it to reject.
+Referring to the above image, when the referenda only has 25% turnout, the tally of "aye" votes has to reach 34% for it to accept.
```
Swader, line 170: I think the image needs to be fixed and labeled better, it does not make sense to me at all.

But I think this edit isn't right. The proposal passes by default, so people in this scenario are fighting to reject it. This should say:

```suggestion
Referring to the above image, when the referendum only has 25% turnout, the tally of "nay" votes must not exceed 66% or the referendum is rejected. When the turnout is 75%, the tally of "nay" votes is tolerated until 54%, anything beyond and the proposal fails.
```

@laboon does this sound about right?

Swader (changes requested):
I am confused by some of this wording. And if I am, I am sure many more will be. Let's fix this and dumb it down, and maybe fix the image too with some labels.

wbarobinson:
I believe that it is a mistake to reclass the negative turnout bias as something that needs to be rejected, rather than accepted. For myself, while learning about your governance structure, it removed clarity, rather than adding. Think of it this way: in all cases you are looking to accept a change, but in the NTB case, that change will be accepted by default. So now it is a sliding scale, as somewhat indicated by the diagram (which is certainly in need of more labelling).


The scale goes from hardest (low turnout positive turnout bias); to hard (high turnout positive turnout bias); to even (total turnout PTB, total turnout NTB, basic majority for when majority of the council proposes); to easy (high turnout NTB); to very easy (low turnout NTB).

docs/learn-governance.md
```diff
@@ -167,9 +167,9 @@ In contrast, when it has 75% turnout, the tally of "aye" votes has to reach 54%,
 
 Suppose there is a unanimous proposal proposed by the council, ``Negative Turnout Bias`` would be used, so that means the proposal is passed by default. Hence, more token holders have to participate in voting to prevent it from passing if they do not like this proposal.
```
lsaether, line 168:
```suggestion
When the council proposes a new proposal through unanimous consent, the referendum would be put to a vote using "Negative Turnout Bias." In this case it is easier to pass this proposal with low turn-out and requires a super-majority to reject. As more token holders participate in voting the bias approaches a plain majority carries. 
```

lsaether:
The governance page needs some TLC.

docs/learn-governance.md
laboon: Much better

docs/learn-governance.md
laboon: I think it's correct, but it's rather confusingly worded.  I'd keep it as aye votes instead of changing it to nay votes.

docs/learn-governance.md
```diff
@@ -167,9 +167,9 @@ In contrast, when it has 75% turnout, the tally of "aye" votes has to reach 54%,
...
 
-Referring to the above image, when the referenda only has 25% turnout, the tally of "nay" votes has to reach 34% for it to reject.
+Referring to the above image, when the referenda only has 25% turnout, the tally of "aye" votes has to reach 34% for it to accept.
 
-In short, when turnout rate is low, a super-majority is required to pass the proposal, which means a higher threshold of "aye" (yes) votes have to be reached, but as turnout increases towards 100%, it becomes a simple-majority.
+In short, when turnout rate is low, a super-majority is required to reject the proposal, which means a lower threshold of "aye" (yes) votes have to be reached, but as turnout increases towards 100%, it becomes a simple-majority.
```
laboon, line 172: 👍

laboon (approved):
see comments

laboon:
I think it makes sense for Logan to go over the whole page after this merges in.  It's kind of hard to understand (and looking at `git blame`, the fault definitely lies behind the keyboard here..)

lsaether:
subsumed by #156  thank you @wbarobinson

docs/learn-governance.md
Swader: Am I misunderstanding that even with zero aye votes the proposal would pass?

laboon: Let me rephrase - your correction is correct but confusingly worded.  Can you reword your change so that it still references the "aye" votes and not the "nay" votes? 

With zero aye votes, no vote should ever be able to pass - this would mean either 100% against (which will always reject) or no voters at all (which, at least with the current code, means automatic rejection - see https://github.com/paritytech/substrate/blob/6611308111143362e210a0e72f302aab534f98f7/srml/democracy/src/vote_threshold.rs#L82)

PR #177, opened 2019-11-30
inject Kusama network data

ansonla3:
With this, it is able to inject the Kusama / Polkadot network data to the markdown files during rendering. 
`fetchNetworkData.js` is the script for getting & processing those data and `createVariableInjectionPlugin` in `siteConfig.js` will do the injection.

**Example:** 
If you want to inject the total of number KSM to the `index.md`, you can add the following injection code anywhere in the md file: 
{@inject: VARIABLE_DEFINED_IN_JSON}

### DIFF
--- a/website/fetchNetworkData.js
+++ b/website/fetchNetworkData.js
@@ -1,0 +1,45 @@
+const { ApiPromise, WsProvider } = require('@polkadot/api');
+const fs = require('fs');
+
+const filename = 'networkData.json' || process.env.NETWORK_DATA_FILE_NAME;
+const KSM_DECIMAL = 1000000000000;
+let numOfProposal, numOfValidator, totalIssuance;
+
+(async function () {
+  // Initialise the provider to connect to the local node
+  const provider = new WsProvider('wss://kusama-rpc.polkadot.io/');
+
+  // Create the API and wait until ready
+  const api = await ApiPromise.create({ provider });
+
+  // Retrieve the chain & node information information via rpc calls
+  [numOfProposal, numOfValidator, totalIssuance] = await Promise.all([
+    api.query.treasury.proposalCount(),
+    api.query.staking.validatorCount(),
+    api.query.balances.totalIssuance(),
+  ]);
+
+  processData();
+  process.exit();
+  
+})();
+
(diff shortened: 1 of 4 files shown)


ansonla3:
Downside: create a new tag that means it shows the value in a new line, so if we want to add some value in between text, we need to see if there is a workaround solution.

Swader:
Brilliant, I'll play with this!

lsaether:
Anson and I discussed this offline. There is a minor annoyance with how the plug-in adds a newline after each injected value.

My suggestion is to ice this one and open a new PR that will just add a Parameters page which pulls all of the parameters and displays them on a newline (probably using a subheader). We can reference this page from other places in the wiki.

Swader:
Sounds like a decent compromise.

Swader:
One solution is to:

1. Remove remarkable-embed (it's getting deprecated anyway)
2. Add a script that's included with every page (https://docusaurus.io/docs/en/site-config#scripts-array)
3. That script calls a node of ours and pulls data from it, then replaces placeholders in text with values, or defaults if values not found.

laboon:
Is this still something we're planning on doing?  It's been sitting in this state for two months now..

Swader:
Yes! Let me play with this.

Swader:
So the script to do the replacements is ready. I'm working with Federico on Watchtower, the graphql endpoint for nodes, so that the script can pull data from that rather than from nodes directly.

ansonla3:
> So the script to do the replacements is ready. I'm working with Federico on Watchtower, the graphql endpoint for nodes, so that the script can pull data from that rather than from nodes directly.

apart from changing the way of getting the data, what's different from the current way of doing it?

Swader:
Nothing really, it's just faster to fetch and a nice demo of graphql integration, good for a tutorial or workshop. And right now it needs Polkadot-JS and issues RPC calls through it, while with a graphql endpoint it'll be issuing regular HTTP calls, so possibly faster. But yeah, I could deploy the script as is and replace source later. I'll try and do that in a PR soon.

Swader:
Closing this in favor of #348

PR #191, opened 2019-12-16
fix broken math in governance

ansonla3:
It seems that the plugin `remarkable-katex` is not able to show the math correctly. 
We can use https://codecogs.com/latex/eqneditor.php to create those formulas and paste the URL.


Swader:
We should debug the cause of the failure, as there will be other places where this math is used and might fail unless fixed at the source. Can we update the plugin / see why it's failing?

laboon:
Yeah, @lsaether is the Wiki King, but I don't like having external links every time we want to display an equation.  Besides adding an external dependency, it will also slow down page loading.

If we can't get the plugin to work, I'd recommend we just generate images locally and include them.

lsaether (approved):

lsaether:
Let's remove the plugin in a follow-up PR.

PR #213, opened 2020-01-15
Fee is a percentage

laboon:
Fixes a line where I missed updating fees when we moved from absolute value to percentage.

thanks for noticing @joepetrowski


joepetrowski (approved):

laboon:
👍

PR #215, opened 2020-01-21
Deprecate Alexander

laboon:
Remove any information on how to nominate/validate on Alexander, or anything that refers to Alexander in anything other than past tense.

Instead of deleting the pages, I have listed notification of the testnet being deprecated and forwarding to Kusama.

*NOTE: Alexander will not be _officially_ deprecated until the new testnet comes up (should be within a week).  Please do not merge until then!*


laboon:
Alexander is no longer available as an option from polkadot.js.org, so I think it's safe to merge whenever people are ready.

laboon:
Kusama Westend has started but I can't connect to it yet.  But I think that shouldn't stop us from removing Alexander.

PR #229, opened 2020-01-29
Polkadot mainnet release 2019 to early 2020

laboon:
Several places mention Polkadot being released at the end of 2019, fixing that to say early 2020.

Also added in Westend testnet to list of networks which was released in Jan 2020, deprecating Alexander.


Swader (approved):
It would be better if we had any links to include with Westend, but we can add that later

laboon:
Sadly and a bit strangely, there doesn't seem to have been an official announcement on Medium or anything.  It just started going. =/ . 

I'll ask Emily if she knows of anything (or if not, can produce something)

PR #198, opened 2019-12-28
added link checker to CI

fgimenez:
Closes https://github.com/w3f/infrastructure/issues/62

PR #232, opened 2020-01-29
Fixed 404s

Swader:
This fixes some links that were still breaking the CI in #198. However, we need to re-trigger the crowdin bot, and I'm not sure how to do that right now. This is needed because those links remain invalid in the auto-generated translation files.


laboon:
👍

PR #233, opened 2020-01-30
Comprehensive identity guide, closes #203

### LINKED ISSUE #203: Comprehensive Identity Guide
The current identity guide is based off the initial implementation and only covers basic functionality. It should be overhauled to include:

- Usage of sub accounts
- Multi-sig registrars
- Step-by-step user guide

Swader:
Still pending multi-sig registrar accounts, will add that later.


docs/learn-identity.md
```diff
@@ -4,27 +4,63 @@ title: Identity
...
+
+You can have a maximum of 100 custom fields.
+
+### Format Caveat
+
+Please note the following caveat: because the fields support different formats, from raw bytes to various hashes, a UI has no way of telling how to encode a given field it encounters. The PolkadotJS UI currently encodes the raw bytes it encounters as UTF8 strings, which makes these values readable on screen. However, given that there are no restrictions on the values that can be placed into these fields, a different UI may be equipped to expect them all to be IPFS hashes only. This means any field stored as raw bytes will become unreadable by that specific UI. As field standards crystallize, things will become easier to use but for now, every custom implementation of displaying user information will likely have to make a conscious decision on the approach to take, or support multiple formats and then attempt multiple encodings until the output makes sense.
 
 ## Registrars
 
 Registrars are appointed from a specified "Registrar Origin", which will start as the [council](learn-governance). Either the council or a public referendum could appoint a new registrar origin to appoint registrars.
```
laboon: I know this is outside of what you wrote, but we should figure this out - should we capitalize "council"?

Same with DOTs, looks like marketing is going with "dots" (and DOT as just the three-letter code à la BTC or BCH)

docs/learn-identity.md
```diff
@@ -4,27 +4,63 @@ title: Identity
...
+
+The easiest way to add the built-in fields is to click the gear icon next to one's account and select "Set on-chain identity".
+
+![Gear icon provides the option to set identity](/img/identity/01.jpg)
+
+A popup will appear, offering the default fields.
+
+![Identity field setup popup](/img/identity/02.jpg)
+
+To add custom fields beyond the default ones, use the Extrinsics UI to submit a raw transaction by first clicking "Add Item" and adding any field you desire. The example below adds a field `steam` which is a user's [Steam](https://store.steampowered.com) username. The first value is the field name in bytes ("steam") and the second is the account name in bytes ("theswader"). The display name also has to be provided, otherwise the Identity pallet would consider it wiped if we submitted it with the "None" option still selected. That is to say, every time you make a change to your identity values, you need to re-submit the entire set of fields: the write operation is always "overwrite", never "append".
```
laboon:
```suggestion
To add custom fields beyond the default ones, use the Extrinsics UI to submit a raw transaction by first clicking "Add Item" and adding any field name you like. The example below adds a field `steam` which is a user's [Steam](https://store.steampowered.com) username. The first value is the field name in bytes ("steam") and the second is the account name in bytes ("theswader"). The display name also has to be provided, otherwise the Identity pallet would consider it wiped if we submitted it with the "None" option still selected. That is to say, every time you make a change to your identity values, you need to re-submit the entire set of fields: the write operation is always "overwrite", never "append".
```

The word "desire" has strong connotations, generally romantic.

docs/learn-identity.md
```diff
@@ -4,27 +4,63 @@ title: Identity
...
+
+Note that custom fields are not shown in the UI by default:
+
+![Only built-in fields are shown](/img/identity/04.jpg)
+
+The rendering of such custom values is, ultimately, up to the UI/dapp makers. In the case of PolkadotJS, the team prefers to only show official fields for now. If you want to check that the values are still stored, use the [Chain State UI](https://polkadot.js.org/apps/#/chainstate) to query the active account's identity info:
+
+![Raw values of custom fileds are available on-chain](/img/identity/05.jpg)
+
+It is up to your own UI or dapp to then do with this data as it pleases. The data will remain available for querying via RPC calls, so you don't have to rely on the PolkadotJS UI.
```
laboon:
```suggestion
It is up to your own UI or dapp to then do with this data as it pleases. The data will remain available for querying via the Polkadot API, so you don't have to rely on the PolkadotJS UI.
```

I know that the Polkadot API wraps RPC calls, but I think this makes more sense since they'd probably call the API.

docs/learn-identity.md
```diff
@@ -4,27 +4,63 @@ title: Identity
...
+
+You can have a maximum of 100 custom fields.
+
+### Format Caveat
+
+Please note the following caveat: because the fields support different formats, from raw bytes to various hashes, a UI has no way of telling how to encode a given field it encounters. The PolkadotJS UI currently encodes the raw bytes it encounters as UTF8 strings, which makes these values readable on screen. However, given that there are no restrictions on the values that can be placed into these fields, a different UI may be equipped to expect them all to be IPFS hashes only. This means any field stored as raw bytes will become unreadable by that specific UI. As field standards crystallize, things will become easier to use but for now, every custom implementation of displaying user information will likely have to make a conscious decision on the approach to take, or support multiple formats and then attempt multiple encodings until the output makes sense.
```
laboon:
```suggestion
Please note the following caveat: because the fields support different formats, from raw bytes to various hashes, a UI has no way of telling how to encode a given field it encounters. The PolkadotJS UI currently encodes the raw bytes it encounters as UTF8 strings, which makes these values readable on screen. However, given that there are no restrictions on the values that can be placed into these fields, a different UI may interpret them as, for example, IPFS hashes or encoded bitmaps. This means any field stored as raw bytes will become unreadable by that specific UI. As field standards crystallize, things will become easier to use but for now, every custom implementation of displaying user information will likely have to make a conscious decision on the approach to take, or support multiple formats and then attempt multiple encodings until the output makes sense.
```

docs/learn-identity.md
```diff
@@ -4,27 +4,63 @@ title: Identity
...
 Registrars can set a fee for their services and limit their attestation to certain fields. For example, a registrar could charge 1 DOT to verify one's legal name, email, and GPG key. When a user requests judgement, they will pay this fee to the registrar who provides the judgement on those claims. Users set a maximum fee they are willing to pay and only registrars below this amount would provide judgement.
 
-## Information
+### Becoming a registrar
+
+To become a registrar, submit a pre-image and proposal into Democracy, then wait for people to vote on it. For best results, write a post about your identity and intentions beforehand, and once the proposal is in the queue ask people to second it so that it gets ahead in the referendum queue.
+
+Here's how to submit a proposal to become a registrar:
+
+Go to Democracy, select Submit preimage, and input the information for this motion - notably which account you're nominating to be a registrar in the `identity.setRegistrar` function.
```
laboon:
```suggestion
Go to the Democracy tab, select "Submit preimage", and input the information for this motion - notably which account you're nominating to be a registrar in the `identity.setRegistrar` function.
```

docs/learn-identity.md
```diff
@@ -39,16 +75,48 @@ When a registrar provides judgement, they can select up to six levels of confide
...
+
+This will make your identity go from unjudged:
+
+![An unjudged identity](/img/identity/07.jpg)
+
+To "waiting":
+
+![A pending identity](/img/identity/09.jpg)
+
+At this point, direct contact with the registrar is required. Each registrar will have their own set of procedures to verify your identity and values, and only once you've satisfied their requirements will the process continue.
```
laboon: Should we add something which explains how to contact a registrar or find/select one in the first place?

docs/learn-identity.md
```diff
@@ -39,16 +75,48 @@ When a registrar provides judgement, they can select up to six levels of confide
...
+
+Once the registrar has confirmed the identity, a green checkmark should appear with the appropriate confidence level:
```
laboon:
```suggestion
Once the registrar has confirmed the identity, a green checkmark should appear next to your account name with the appropriate confidence level:
```

docs/learn-identity.md
```diff
@@ -39,16 +75,48 @@ When a registrar provides judgement, they can select up to six levels of confide
...
+
+![A confirmed identity](/img/identity/10.jpg)
+
+_Note that changing even a single field's value after you've been verified will un-verify your account and you need to start the judgement process anew, but you can still change fields while the judgement is going on - it's up to the registrar to keep an eye on the changes._
```
laboon:
```suggestion
_Note that changing even a single field's value after you've been verified will un-verify your account and you will need to start the judgement process anew.  However, you can still change fields while the judgement is going on - it's up to the registrar to keep an eye on the changes._
```

website/i18n/en.json
```diff
@@ -177,10 +177,6 @@
         "title": "Treasury",
         "sidebar_label": "Treasury"
       },
-      "learn-UI": {
-        "title": "Polkadot UI",
-        "sidebar_label": "Polkadot UI"
-      },
```
laboon, line 183: Why did you get rid of this?

docs/learn-identity.md
```diff
@@ -4,27 +4,63 @@ title: Identity
 sidebar_label: Identity
 ---
 
-Polkadot provides a naming system that allows participants to ask for verification of off-chain information from registrars. Users can ask for verification of multiple attributes (e.g. name, website, image). 
+Polkadot provides a naming system that allows participants to add personal information to their on-chain account and subsequently ask for verification of this information by [registrars](#registrars).
+
+## Setting an Identity
+
+Users can register some default fields like legal name, display name, website, Twitter handle, Riot handle, etc. along with extra, custom fields for which they would like attestations (see [Judgements](#judgements)). Users must reserve funds in a bond to store their information on chain - 10 per identity, and 2.5 KSM per each field beyond the legal name. These funds are locked and not spent - they are returned when the identity is cleared. Each field can store up to 32 bytes of information, so the data must be less than that. When inputting the data manually through the [Extrinsics UI](https://polkadot.js.org/apps/#/extrinsics), a [UTF8 to bytes](https://onlineutf8tools.com/convert-utf8-to-bytes) converter can help.
```
laboon:
```suggestion
Users can register some default fields like legal name, display name, website, Twitter handle, Riot handle, etc. along with extra, custom fields for which they would like attestations (see [Judgements](#judgements)). Users must reserve funds in a bond to store their information on chain - 10 KSM per identity, and 2.5 KSM per each field beyond the legal name. These funds are _locked_, not spent - they are returned when the identity is cleared. Each field can store up to 32 bytes of information, so the data must be less than that. When inputting the data manually through the [Extrinsics UI](https://polkadot.js.org/apps/#/extrinsics), a [UTF8 to bytes](https://onlineutf8tools.com/convert-utf8-to-bytes) converter can help.
```

laboon (changes requested):

website/i18n/en.json
Swader: Accidental. That page was removed before because it seemingly wasn't linked in any sidebar (it was causing problems in the link checker), so I cleaned up the JSON here too. Was it supposed to stick around? I thought it was orphaned for a reason (outdatedness).

docs/learn-identity.md
Swader: Done

Swader:
@laboon I've added a way to identify and contact a registrar and incorporated other feedback, please check

lsaether (approved):
lgtm

website/i18n/en.json
laboon: No, I didn't know it was already removed and it didn't seem relevant to the rest of the PR, so I asked.  May as well just get rid of it now.

laboon:
Looks great!

laboon (approved):

PR #235, opened 2020-01-30
Removed some dead/outdated links

Swader:
I would just as soon remove this page entirely because it's packed with resources that are incredibly hard to keep up to date.


lsaether:
It does serve a purpose for being a single collection of all the links, but historically this was from before wiki times when it was used as the only central guide post.

lsaether (approved):

PR #248, opened 2020-02-06
Added explainer on transactions and metering

### LINKED ISSUE #241: Add tx fee and weight information
```
dieter
Do we have any easy to read write-ups that cover Polkadot's tx fee model including how this works in parachains vs. parathreads and the major differences between this for Polkadot and Ethereum?
I feel like this would be good to have if we dont
emily
it's mentioned here: https://polkadot.network/polkadot-parachain-slots/
under "cost of a parachain" 
Bruno | W3F
We can focus on it during Feb's wiki blitz
dieter
emily
it's mentioned here: https://polkadot.network/polkadot-parachain-slots/
Ya but this doesn't really talk about tx costs in addition to parachain/parathread costs
emily
but it easily could be a full article. It also doesn't mention parathreads. For parathreads there is this: https://medium.com/polkadot-network/parathreads-pay-as-you-go-parachains-7440d23dde06
In reply to emily
dieter
Ya but this doesn't really talk about tx costs in addition to parachain/parathread costs
yes agreed.
Also, a lot has changed since this was written
dieter
ya what i really mean is how Polkadot implements this https://substrate.dev/docs/en/next/development/module/fees
i think parachain cost and parathread costs are fairly well explained and people seem to understand generally how they work
Bruno | W3F
There's a size limit and a weight limit in every block. Whichever gets filled up first is what makes a block full. Weights are currently defined but not very well thought through and not even used right now, they're due for a revamp. Max size is 5MB. The concept is exactly the same as in eth with gas, only the block size doesn't auto-scale on demand in Polkadot, and the gas price is in the runtime, not in the VM, but they're basically identical things.
If this warrants a full post that goes into specifics, we can focus on it, sure.
dieter
I feel like the post would be valuable - I don't think this talked about anywhere
it can be short - even that bit was really helfpul for me
or maybe it doesnt need a post but just a short wiki entry
Bruno | W3F
Okay, I agree that this little bit isn't mentioned anywhere and can be a missing puzzle piece. I'll add it to the wiki into issues, we'll handle it.
```

Swader:
Closes #241


lsaether:
Git is showing irrelevant commits on this PR. Can you rebase to master before we merge this in?

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
+---
+id: learn-transactions
+title: Transaction Types and Metering
+sidebar_label: Transaction Types and Metering
+---
+
+Depending on the pallets installed in a [Substrate](https://substrate.dev)-based chain like Polkadot or Kusama, different transactions are possible within the system. For example, one can propose referenda in the Governance pallet, verify identities in the Identities pallet, send tokens through the Balances pallet, shout remarks into the chain through the System pallet, and more.
```
lsaether:
```suggestion
Depending on the pallets installed in a [Substrate](https://substrate.dev)-based chain like Polkadot or Kusama, different transaction types are possible within the system. For example, one can propose referenda in the Governance pallet, verify identities in the Identities pallet, send tokens through the Balances pallet, shout remarks into the chain through the System pallet, and more. 
```

Would it be better to change the examples into a markdown list?

Swader: I feel like that would demand more examples. This kind of prose-y exemplification lets me get away with less imo

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
+---
+id: learn-transactions
+title: Transaction Types and Metering
+sidebar_label: Transaction Types and Metering
+---
+
+Depending on the pallets installed in a [Substrate](https://substrate.dev)-based chain like Polkadot or Kusama, different transactions are possible within the system. For example, one can propose referenda in the Governance pallet, verify identities in the Identities pallet, send tokens through the Balances pallet, shout remarks into the chain through the System pallet, and more. 
+
+Batching these transactions into blocks cannot work based on size like it does in a blockchain like Bitcoin's because smaller transactions might require more computation and vice versa - it's all about their computational effect. These transactions also do not happen on a smart contract platform (as smart contracts are not a first-class citizen of Substrate-based chains), so there is no concept of gas to meter transactions with.
```
lsaether:
```suggestion
One option for batching transactions into a block is based on the size (in bytes) of the transactions. This approach works well for chains without complex computational logic like Bitcoin, but causes problems for Polkadot due to the variance of computational effect of each transaction type. These transactions also do not happen on within smart contract platform or virtual machine (as smart contracts are not the first-class computational model of Substrate-based chains), so there is no concept of gas metering to limit the block size.
```

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+
+Transactions are the consequence of extrinsics. Transactions often have *weight*.
+
+A block in Kusama is limited to [1 billion weight](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/lib.rs#L53) or to [5 MB](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/lib.rs#L55), whichever limit is hit first. These parameters _may_ change for Polkadot mainnet.
+
+Most transactions in the runtime code [have a weight declaration](https://github.com/paritytech/substrate/blob/922b36c8eb6d971fbe2b72d69099a291a368ea25/frame/sudo/src/lib.rs#L122):
+
+![The sudo function has a weight of 50000](/img/tx/01.jpg)
+
+The example above shows that the `sudo` function in the `sudo` pallet has a weight of 50000. Some transactions are free, meaning they carry no weight and only pay the base transaction fee as calculated by the system.
```
lsaether:
```suggestion
The example above shows that the `sudo` function in the `sudo` pallet has a weight of `50000`. Some transactions are free, meaning they carry no weight and only pay the base transaction fee as calculated by the system.
```

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+
+The weight and size limits apply only to blocks on the Relay Chain, and not to parachains or parathreads (see below).
+
+## Parachain Transactions
+
+Once a [parachain](learn-parachains) has won the [auction](learn-auction) for a parachain slot and attached itself to the Relay Chain (RC), that parachain is guaranteed free inclusion of blocks into the relay chain for as long as the slot lease lasts. This means extrinsics initiated by the validators who are including a parachain's block (sent to them by collators of that parachain) in the RC have priority - they are always included. The remaining weight and size is used by RC operations as described above.
+
+The collators of a parachain can but do not have to pay transaction costs, and those costs - if any - are internal to the chain. The economics of a parachain are up to the parachain to define, and whether or not the collators are incentivized to send proofs to the RC validators and process the parachain transactions is entirely up to the designers. For example, a parachain can but does not have to use the weight system.
```
lsaether:
```suggestion
The collators of a parachain can but do not have to pay transaction costs, and those costs - if any - are internal to the chain. The economics of a parachain are up to the parachain to define, and whether or not the collators are incentivized to send proofs to the Relay Chain validators and process the parachain transactions is entirely up to the designers. For example, a parachain can but does not have to use the weight system.
```

docs/learn-transactions.md
Swader: Yarr there be typos in this edit

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+---
+
+Depending on the pallets installed in a [Substrate](https://substrate.dev)-based chain like Polkadot or Kusama, different transactions are possible within the system. For example, one can propose referenda in the Governance pallet, verify identities in the Identities pallet, send tokens through the Balances pallet, shout remarks into the chain through the System pallet, and more. 
+
+Batching these transactions into blocks cannot work based on size like it does in a blockchain like Bitcoin's because smaller transactions might require more computation and vice versa - it's all about their computational effect. These transactions also do not happen on a smart contract platform (as smart contracts are not a first-class citizen of Substrate-based chains), so there is no concept of gas to meter transactions with.
+
+To seemingly complicate matters further, there are parachain transactions (within the parachain, as well as parachain-to-relay-chain), parathread transactions, and smart contract transactions in a smart contract pallet like the EVM pallet or the Ink pallet. This document will explain how it all fits together and how we meter transactions in Polkadot.
```
lsaether, line 11: Where do we class XCMP messages?

Swader: I don't think those count here because other chains don't have something similar. This document is meant to only compare transactions to their implementations in other chains that people might be familiar with. That said, if XCMP messages change something here drastically, let's include them. Do we know *anything* about their fees / inclusions / types / frequency, anything concrete we could state here?

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+
+The economics of parathread inclusion are still being fleshed out, so the 50 parathread blocks per slot number may change along with the specific inclusion fee.
+
+Transactions occurring on the parathread itself are, as with parachains, specific to that environment and are not metered by the Relay Chain in any way. Collators can but do not have to be incentivized to process the transactions, and the chain itself can but does not have to have any metering built-in.
+
+## Smart Contract Transactions
+
+Some chains (threads or full parachains) will include a smart contract pallet. This would let third parties develop applications on that chain without having to have native code commit access to that chain's runtime. Transactions within that smart contract pallet might need to be metered somehow as well.
```
lsaether:
```suggestion
Some chains (threads or full parachains) will include a smart contract pallet. This would let third parties develop applications on that chain without having to have native code commit access to that chain's runtime. Transactions within that smart contract pallet will need to be metered or otherwise access-controlled to prevent spam and block fullness.
```

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+
+Just like with parathreads and parachains, the smart contract transactions within a parathread or parachain are entirely local and do not affect any fee or metering calculation on the Relay Chain. The chain implementing the smart contract pallet can meter the transactions using its own design, or it can decide that no metering is necessary at all, or anything in between. The transaction fees and block sizes of the chain are up to the chain and its collators and a block produced by a thread or chain with a smart contract pallet will be like any other thread's or chain's block - as long as the state transition function of the chain offers the `execute_block` function for validators to execute and validate, the chain's internals are no business of the relay chain.
+
+## Short version
+
+There are four types of transactions:
+
+1. On-relay-chain transactions (from various pallets like governance, identity, etc.), metered by weight and block size
```
lsaether:
```suggestion
1. Relay Chain transactions (from various pallets like governance, identity, etc.), metered by weight and block size.
```

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+2. Parachain block transactions, always free
```
lsaether:
```suggestion
2. Parachain block transactions, always free and granted by placement in the parachain slot.
```

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+3. Parathread block transactions, basic fee when no other threads in pool want to write, auction when there's high demand for writes
```
lsaether:
```suggestion
3. Parathread block transactions, basic fee when no other threads in pool want to write, auction when there's high demand for writes.
```

docs/learn-transactions.md
```diff
@@ -0,0 +1,62 @@
...
+4. Internal parathread / parachain transactions (including smart contracts): completely local and dependent on whatever the chain designers decide, these do not bleed into Relay Chain economics in any way
```
lsaether:
```suggestion
4. Internal parathread / parachain transactions (including smart contracts): completely local and dependent on whatever the chain designers decide, these do not bleed into Relay Chain economics in any way.
```

lsaether (changes requested):
I would switch the example from `sudo` to `transfer` or something that will be included in the Relay Chain runtime.

Swader:
@lsaether yeah probably a good idea, I'll look for another good example.

docs/learn-transactions.md
Swader: So it turns out XCMP channels will "stay open" and require a DOT deposit of some kind (Gav's words). Since this is still all up in the air, I'd rather avoid the topic.

Swader:
@joepetrowski you asked to be tagged when this is ready

Swader:
Closed in favor of #264

PR #252, opened 2020-02-08
do not require check-links for deploying

### LINKED ISSUE #251: Remove translated docs from link-checking
@fgimenez can you please set up the CI so that it doesn't take translated docs into account? As it is, all our PRs will auto-fail when we add new pages, and it doesn't make sense to wait for the crowdin bot to do a PR after every submit of ours. Since the crowdin PR is automated, it will by default inherit good and bad links from the main docs, so it doesn't make much sense to check those anyway.

fgimenez:
Closes https://github.com/w3f/polkadot-wiki/issues/251

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,8 +57,6 @@ workflows:
             tags:
               only: /.*/
       - deploy-website:
-          requires:
-            - check-links
           filters:
             branches:
               only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       - run:
           name: check links
           command: |
-            pylinkvalidate.py --parser=html5lib -P http://localhost:3000
+            pylinkvalidate.py --parser=html5lib -P http://localhost:3000/docs/en -d 0
 
   deploy-website:
     docker:


Swader:
Great! Is it hard to update the CI setup to not check translated docs for links so we can solve both things at once?

fgimenez:
If there's a concrete path to be ignored then it can be done pretty easily, for example `docs/zh-CN/` ?

Swader:
Anything NOT under /en/ should be ignored. So docs/en/ yes, docs/anything-else/ no.

fgimenez:
Ok pls take a look, now the crawler starts in /docs/en and only reaches 1 level (bc there are links to, for instance, /docs/zh-CN from /docs/en), let me know wdyt

Swader:
Looks good to me!

Swader (approved):
Let's get this in - will it also break due to failing CI before this non-fail CI is merged in?

fgimenez:
> Let's get this in

Ok, done. There were still some 404 linked from `/docs/en/`, pls check https://circleci.com/gh/w3f/polkadot-wiki/1308

> will it also break due to failing CI before this non-fail CI is merged in?

All the branches forked off master from now on will have the fi in place, the existing branches need to rebase and push to refresh `.circleci/config.yml`

Swader:
Thanks!

PR #254, opened 2020-02-09
fix link-checker base

### LINKED ISSUE #253: CI checking redirection
https://circleci.com/gh/w3f/polkadot-wiki/1312?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@fgimenez the 404 failures here are because the link checker is looking at `docs/en` and not `docs/en/` - in deployment, docs/en redirects to `docs/en/` and this makes relative links work (because they get appended to `/en/`) but in the CI checking the redirect does not happen on this index page, and so it 404s on links that are relative to the wrong directory.

This only happens on the home page of the wiki which is the only one without a page name. Can we set the CI up so that it does not check `/docs/en` or so that it redirects `docs/en` to `/docs/en/`?

fgimenez:
Closes #253

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       - run:
           name: check links
           command: |
-            pylinkvalidate.py --parser=html5lib -P http://localhost:3000/docs/en -d 0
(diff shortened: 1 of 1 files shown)


Swader:
Perfect

PR #257, opened 2020-02-12
Existential deposit

### LINKED ISSUE #220: Define existential deposit and explain reaping better
what's the minimum amount of KSM one should have in order to transfer funds, vote, nominate etc.? I vaguely recall something like 0.01 KSM mentioned at some point.

Correct, the existential deposit is 0.01 KSM.  If an account has less than that, it is "reaped" (its index can be used by another address, i.e., your account is gone).  

Beyond that you need enough for transaction fees for what you want to do. They are lowish now but variable.

OK, makes sense, thanks. I was trying to make sure that someone with only 0.01 KSM cannot do anything practically other than keeping an account.

Correct, the existential deposit is 0.01 KSM.  If an account has less than that, it is "reaped" (its index can be used by another address, i.e., your account is gone).  
With most cryptocurrencies, I generate my keys offline, I get my account hash. What the gosh darn heck does it mean to "reap" an account that doesn't have money?

What if I sent 100 KSM to it, and then sent 100 KSM from it. The record of that account would still exist in the blocks that contained those transactions. Right?

If I get reaped, I can always send more money to it and get un-reaped, right? Is there something I'm missing.

It just means it's not stored in the database as an indexed active account. So it's like removing the data referencing that account. You can revive it with a deposit, but your index will be different. When you generate a key, you just generate an access code to open the account that'll eventually exist from that seed
So by "cleaning up" these low balance accounts, the database of the nodes shrinks. Think of it like your OS automatically clearting the Thumbs (windows) or .DS_Store files from empty folders and deleting the folders. It doesn't mean you can't ever use the folder again, you just have to recreate them (and account creation fee exists - it's something like 0.001 I think).

Swader:
Closes #220 by clarifying accounts in general, including existential deposits.


Swader:
@fgimenez this PR should have failed, do you know why the checker may have failed to do its job here?. I have included 3 invalid links: a local one, an external one that does not resolve, and an external 404. None of them were detected.

Swader:
@fgimenez it looks like it might need increased depth - right now it only crawls the homepage and the links it can find there, but a hops count of 0 makes it stay on that seed page. By increasing to 1, it should follow every link leading out of those linked sites too. That should be enough.

docs/learn-accounts.md
```diff
@@ -0,0 +1,89 @@
...
+
+Other than the extension and the default UI, Polkadot and Kusama addresses can also be created with the [Subkey tool](https://github.com/paritytech/substrate/tree/master/bin/utils/subkey). Subkey is intended for users comfortable with using the command line and can seem intimidating, but is quite approachable. Follow the instructions in [this page](https://substrate.dev/docs/en/ecosystem/subkey). When used properly, Subkey is the **most secure** available method of creating an account.
+
+There is also the very secure [Parity Signer](https://www.parity.io/signer/) but it currently only supports Kusama addresses, not Polkadot or generic Substrate.
+
+Alternatively, you might find some wallets on the [Polkaproject.com page](http://www.polkaproject.com/) but bear in mind that these are **unaudited** and not officially affiliated with Web3 Foundation or the Polkadot project unless otherwise stated.
+
+Hardware wallet integration from [Ledger](https://ledger.com) is coming soon.
+
+This is an [invalid external link](https://foobarbaznonexistent.com.uk) for testing our CI.
```
lsaether: needs to be removed

docs/learn-accounts.md
```diff
@@ -0,0 +1,89 @@
...
+
+This is an [invalid internal link](learn-something-that-doesnt-exist) for testing our CI.
```
lsaether: needs to be removed

docs/learn-accounts.md
```diff
@@ -0,0 +1,89 @@
...
+
+This is an [external 404 link](https://bitfalls.com/something-non-existent) for testing our CI.
```
lsaether: needs to be removed

docs/learn-accounts.md
```diff
@@ -0,0 +1,89 @@
...
+If slots 0-2 are populated by addresses A, B and C respectively, and I add an existential deposit to address X, that address will automatically be put into slot 3. Henceforth, you can send me money by just sending to `[3]` rather than remembering my whole address.
+
+```js
+[0][1][2][3][4][5][6]...
+[A][B][C][X][ ][ ][ ]...
+```
+
+But what if an account gets reaped as explained above? In that case, the index is emptied. In other words, the slot frees up again. If someone creates a new account, they may end up using the same index another address was using before.
+
+Because of this unreliability of indices, they will **not be present** in Polkadot mainnet.
```
lsaether: Has this been confirmed? Last I heard it was "may" not be present.

lsaether (commented):
Looks good after being cleaned up of CI checking.

docs/learn-accounts.md
Swader: Last I heard was "probably won't be there". I say risk it and say "won't be" and fix later if needed.

lsaether: I would say keep it to "may" since if it does end up being included we don't want to be wrong. And we won't have to remember to come back to fix it.

Swader: Ok

fgimenez:
> @fgimenez it looks like it might need increased depth - right now it only crawls the homepage and the links it can find there, but a hops count of 0 makes it stay on that seed page. By increasing to 1, it should follow every link leading out of those linked sites too. That should be enough.

@Swader Sorry for the late response, the problem with depth 1 is that it might enter the translations and find other unrelated broken links, is that ok? I can propose in a separate PR and we can test.

Swader:
@fgimenez we should try. Like this it doesn't make sense, it doesn't do anything useful. Can you set up a bigger depth but make it NOT crawl URLs which do not have /en/?

fgimenez:
> @fgimenez we should try.

ok i'll prepare a PR

> Can you set up a bigger depth but make it NOT crawl URLs which do not have /en/?

we would need to specify the list, afaik regex-like selectors are not allowed, let's check in the PR

fgimenez:
> @Swader Sorry for the late response, the problem with depth 1 is that it might enter the translations and find other unrelated broken links, is that ok? I can propose in a separate PR and we can test.

@Swader proposed in #261 PTAL

lsaether:
Is this blocked on the link checker or can we merge it?

Swader:
Let me fix the 404s and move them into another PR, then we can merge this in.

PR #258, opened 2020-02-12
Start folding in the Kusama guide

### LINKED ISSUE #16: Kusama Section
Evaluate integrating all Kusama content into the wiki in it's own "Kusama" section. This would make it easier to maintain as a single repository but may clutter up the Polkadot wiki unnecessarily.

lsaether:
closes #16 

- [x] index
- [x] faucet
- [x] timeline (including transition to PoS)
- [x] community
- [x] try
- [x] claims

## Migrating URLs

The biggest problem I see is that we've linked different pages of the Kusama guide in a bunch of different places. Specifically, the Claims guide and the validator guide will likely be spread out in different places on the web. 

### Redirects

We should catch everything that goes to `guide.kusama.network` or `guide.kusama.network/**/*` and point it to `wiki.polkadot.network/docs/en/kusama-index`.

## Sidebar

Now the sidebar includes a Kusama tab. Not sure if I'm happy with this as an end result. Especially when Polkadot launches and Kusama starts to diverge even more. We may want to host Kusama as a sub-wiki inside the Polkadot wiki instead of a "topic" like it is now.

image

EDIT:

image


Swader:
I'm a little hesitant to include it like this, because now we're missing the "try" and other actionable sections from Kusama Guide. The index might be an ideal place for those, if they are now under Polkadot Wiki's general maintain / build. 

The Kusama sub-category could also have links to existing maintain/build pages in Polkadot wiki, so they'd be listed twice in the sidebar, but users who expect them in Kusama could find them there, just like those looking for build/maintain in general.

lsaether:
@Swader This is pretty much ready. However when doing this I noticed the Builder's Starter Guide is out of date. Since this is already a pretty large PR, I think it's better to update that separately.

Swader:
Going to take a look tonight

Swader (approved):
This looks good to me - the separation in the top menu is apparent enough to be accessible but hidden away enough not to be intrusive. I love that we're able to directly share docs between the two "sub-wikis" (e.g. learn-identity) and I'd love it if we could write future guides with that in mind, to maximize reuse. Let's merge this in asap and kill the other kusama guide. In fact, those URLs might even be 1:1 redirectable to this new stuff now, they're not *that* different in most cases and there aren't *that* many... should we go for granular redirects, or just a blanket redirect to homepage?

lsaether:
It shouldn't be an issue to do both. We'll do a granular redirect of all the pages if we can and a blanket one to catch anything that slipped through.

PR #261, opened 2020-02-14
switch to linkchecker

fgimenez:
Uses https://github.com/linkchecker/linkchecker instead of pylinkvalidate, this allows us to ignore urls with regex.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   check-links:
     docker:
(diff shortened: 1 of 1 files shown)


Swader:
What's the best way to check if this works on my invalid links branch / PR?

fgimenez:
I've proposed these same changes to your branch here https://github.com/w3f/polkadot-wiki/pull/262

fgimenez:
Will continue working from #269

PR #262, opened 2020-02-14
add new linkchecker

fgimenez:


Swader:
Ok, this found 1 of the three - the local one. The external non-existent links were not detected though.

fgimenez:
The verbose execution is here https://circleci.com/gh/w3f/polkadot-wiki/1504 you can check the links that were filtered, for instance: 
```
URL        `https://foobarbaznonexistent.com.uk'
Name       `invalid external link'
Parent URL http://localhost:3000/docs/en/learn-accounts, line 104, col 15
Real URL   https://foobarbaznonexistent.com.uk
Info       The URL is outside of the domain filter, checked only
           syntax.
Result     Valid: filtered
```
Currently it is configured to ignore all the links that don't have `/en/` on them.

fgimenez:
After adding `--check-external` a few more errors are shown (all of them with `/en/` in the url)

Swader:
Very good - we're one step away from success! Now we need to modify the exclusion flag to ignore all without /en/ UNLESS they are absolute links. Possible?

fgimenez:
First attempt didn't work, on it

Swader:
@fgimenez let's continue in a new PR. I made #269

fgimenez:
Ok created #270 on top of #269, will continue there.

PR #264, opened 2020-02-15
Add page on transaction fees

### LINKED ISSUE #241: Add tx fee and weight information

joepetrowski:
Closes https://github.com/w3f/polkadot-wiki/issues/241

Supersedes https://github.com/w3f/polkadot-wiki/pull/248


docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+## Fee Calculation
+
+Fees on the Polkadot Relay Chain are calculated based on four parameters:
+
+- A base fee
+- A per-byte fee (also known as the "length fee")
+- A weight fee
+- A tip (optional)
+
+The base fee is a fixed fee applied to every transaction. The length fee is simply the product of a constant per-byte fee and the size of the transaction in bytes.
```
Swader:
```suggestion
The base fee is a fixed fee applied to every transaction. The length fee is the product of a constant per-byte fee and the size of the transaction in bytes.
```

Swader: The word simply is best avoided.

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+Weights are simply a fixed number designed to represent overall resource consumption of a transaction. Resources could be CPU cycles, memory usage, disk I/O, etc. Each transaction has a weight, which is multiplied by a per-weight fee to calculate the transaction's weight fee.
```
Swader:
```suggestion
Weights are a fixed number designed to represent overall resource consumption of a transaction. Resources could be CPU cycles, memory usage, disk I/O, etc. Each transaction has a weight, which is multiplied by a per-weight fee to calculate the transaction's weight fee.
```

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+Block producers prioritize transactions based on each transaction's total fee. Since a portion of the fee will go to the block producer, producers will include the transactions with the highest fees to maximize their reward.
+
+## Fee Adjustment
+
+Transaction volume on blockchains is highly irregular, and therefore transaction fees need a mechanism to adjust. However, users should be able to predict transaction fees.
+
+Polkadot uses a slow-adjusting fee mechanism with tips to balance these two considerations. In addition to block _limits,_ Polkadot also has a block fullness _target._ Fees increase or decrease for the next block based on the fullness of the current block relative to the target. The per-weight fee can change up to 30% in a 24 hour period. This rate captures long-term trends in demand, but not short-term spikes. To consider short term spikes, Polkadot uses tips on top of the base, byte, and weight fees. Users can optionally add a tip to the fee to give the transaction a higher priority.
```
Swader, line 48: This is incredibly useful, didn't know this. Where can one find out more about the target? Is it also subject to governance?

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+[Parathreads](learn-parathreads) will also make all the same transactions that a parachain might. In addition, the collators need to participate in an auction every block to progress their chain. The collators will need to have DOTs to participate in these auctions.
+
+## Other Resource Limitation Strategies
+
+Transaction weight must be computable prior to execution, and therefore can only represent fixed logic. Some transactions warrant limiting resources with other strategies. For example:
+
+- Bonds: Some transactions, like voting, may require a bond that will be returned later, or slashed if the voter tries anything malicious.
+- Burns: A transaction may burn funds internally based on its logic. For example, a transaction may burn funds from the sender if it creates new storage entries, thus increasing the state size.
```
Swader, line 64: I'd like more info about this - are all write tx burning part of the fee or just some? If some, which for example?

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Signed transactions
+- Unsigned transactions
+- Inherents
+
+This page only covered signed transactions, which is the way that most users will interact with Polkadot. Signed transactions come from an account that has funds, and therefore Polkadot can charge a transaction fee as a way to prevent spam.
+
+Unsigned transactions are for special cases where a user needs to make a transaction from an account that does not have funds. For example, when users claim their DOT tokens, their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```
Swader:
```suggestion
Unsigned transactions are for special cases where a user needs to make a transaction from an account that does not have funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network), their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```

Swader: Could a validator spam the network and bloat state with a barrage of unsigned heartbeats?

website/i18n/en.json
```diff
@@ -177,6 +177,10 @@
         "title": "Staking",
         "sidebar_label": "Staking"
       },
+      "learn-transactions": {
```
Swader:
```suggestion
      "learn-transaction-fees": {
```

Swader (changes requested):

docs/learn-transaction-fees.md
joepetrowski: Ah, breaking my own rule! Good catch.

docs/learn-transaction-fees.md
joepetrowski: Hmm, our docs are a bit light, but I guess not one place where it all comes together. Here are a few:
- Fees Page: https://substrate.dev/docs/en/next/development/module/fees#fee-multiplier
- Research: https://research.web3.foundation/en/latest/polkadot/Token%20Economics.html#-2.-slow-adjusting-mechanism
- Sub 0: https://www.youtube.com/watch?v=mN0ueyTWr_8

In Substrate, `NextFeeMultiplier` is abstract and is made concrete in the runtime (so it's up to the runtime to decide what this multiplier should be). 

Yes, this target and the function to calculate the `NextFeeMultiplier` are both in the runtime and therefore changeable with a runtime upgrade.

docs/learn-transaction-fees.md
joepetrowski: Nope, and actually we just removed the only example that existed. `transfer` used to take an extra fee if an account was created. Gav took it out with some other method (and the existential deposit is a form of bonding). I guess this could come out of the Polkadot RC wiki since we're not doing it anymore. Still is an option for designing a chain.

docs/learn-transaction-fees.md
joepetrowski: Nope, that's where this custom validation logic comes in. The transaction queue does a bunch of checks on these before even accepting them (like making sure that the validator hasn't already sent one). So literally for each unsigned tx that your chain has, it needs to have its own `validate_unsigned` function.

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Signed transactions
+- Unsigned transactions
+- Inherents
+
+This page only covered signed transactions, which is the way that most users will interact with Polkadot. Signed transactions come from an account that has funds, and therefore Polkadot can charge a transaction fee as a way to prevent spam.
+
+Unsigned transactions are for special cases where a user needs to make a transaction from an account that does not have funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network), their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```
joepetrowski:
```suggestion
Unsigned transactions are for special cases where a user needs to submit an extrinsic from a key pair that does not control funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network), their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```

docs/learn-transaction-fees.md
joepetrowski: As an example, this is the Balances pallet from a few weeks ago: https://github.com/paritytech/substrate/blob/d2bb42d314bbdcc48c0cf01cb59f7ab10294ebad/frame/balances/src/lib.rs#L989

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Signed transactions
+- Unsigned transactions
+- Inherents
+
+This page only covered signed transactions, which is the way that most users will interact with Polkadot. Signed transactions come from an account that has funds, and therefore Polkadot can charge a transaction fee as a way to prevent spam.
+
+Unsigned transactions are for special cases where a user needs to submit an extrinsic from a key pair that does not control funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network), their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```
joepetrowski:
```suggestion
Unsigned transactions are for special cases where a user needs to submit an extrinsic from a key pair that does not control funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network), their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These heartbeats must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
+---
+id: learn-transaction-fees
+title: Transaction Fees
+sidebar_label: Transaction Fees
+---
+
+Several resources in a blockchain network are limited, for example, storage and computation. Transaction fees prevent individual users from consuming too many resources. Polkadot uses a weight-based fee model as opposed to a gas-metering model. As such, fees are charged prior to transaction execution; once the fee is paid, nodes will execute the transaction.
+
+Web3 Foundation designed the Polkadot fee system with the following objectives:
```
lsaether:
```suggestion
[Web3 Foundation Research](https://research.web3.foundation/en/latest/polkadot/Token%20Economics.html#relay-chain-transaction-fees-and-per-block-transaction-limits) designed the Polkadot fee system with the following objectives:
```

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+- A weight fee
+- A tip (optional)
+
+The base fee is a fixed fee applied to every transaction. The length fee is the product of a constant per-byte fee and the size of the transaction in bytes.
+
+Weights are a fixed number designed to represent overall resource consumption of a transaction. Resources could be CPU cycles, memory usage, disk I/O, etc. Each transaction has a weight, which is multiplied by a per-weight fee to calculate the transaction's weight fee.
+
+Tips are an optional transaction fee that users can add to give a transaction higher priority.
+
+Together, these four fees constitute the inclusion fee. This fee is deducted from the sender's account prior to transaction execution. A portion of the fee will go to the block producer and the remainder will go to the [Treasury](learn-treasury). At genesis, this is set to 20% and 80%, respectively.
```
lsaether:
```suggestion
Together, these four fees constitute the inclusion fee. This fee is deducted from the sender's account prior to transaction execution. A portion of the fee will go to the block producer and the remainder will go to the [Treasury](learn-treasury). At Polkadot's genesis, this is set to 20% and 80%, respectively.
```

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Block Limits and Transaction Priority
+
+Blocks in Polkadot have both a maximum length (in bytes) and a maximum weight. Block producers will fill blocks with transactions up to these limits. A portion of each block - currently 25% - is reserved for critical transactions that are related to the chain's operation, such as misbehavior reports. Block producers will only fill up to 75% of a block with normal transactions.
```
lsaether: > ... critical transactions that are related to the chain's operation, such as misbehavior reports.

What can these transactions be besides the misbehavior reports?

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+[Parachains](learn-parachains) have a dedicated slot on the Relay Chain for execution, so their collators do not need to own DOTs in order to include blocks. The parachain will make some transactions itself, for example, opening or closing an [XCMP](learn-crosschain) channel, participating in an [auction](learn-auction) to renew its slot, or upgrading its runtime. Parachains have their own accounts on the Relay Chain and will need to use those funds to issue transactions on the parachain's behalf.
+
+[Parathreads](learn-parathreads) will also make all the same transactions that a parachain might. In addition, the collators need to participate in an auction every block to progress their chain. The collators will need to have DOTs to participate in these auctions.
+
+## Other Resource Limitation Strategies
+
+Transaction weight must be computable prior to execution, and therefore can only represent fixed logic. Some transactions warrant limiting resources with other strategies. For example:
+
+- Bonds: Some transactions, like voting, may require a bond that will be returned later, or slashed if the voter tries anything malicious.
```
lsaether:
```suggestion
- Bonds: Some transactions, like voting, may require a bond that will be returned later, or slashed if the voter tries anything malicious. Sometimes these are also referred to as _deposits_.
```

These are also referred to as deposits in the code ([for example the new indices pallet](https://github.com/paritytech/substrate/blob/master/frame/indices/src/lib.rs#L51)).

docs/learn-transaction-fees.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Signed transactions
+- Unsigned transactions
+- Inherents
+
+This page only covered signed transactions, which is the way that most users will interact with Polkadot. Signed transactions come from an account that has funds, and therefore Polkadot can charge a transaction fee as a way to prevent spam.
+
+Unsigned transactions are for special cases where a user needs to submit an extrinsic from a key pair that does not control funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network), their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These heartbeats must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```
lsaether:
```suggestion
Unsigned transactions are for special cases where a user needs to submit an extrinsic from a key pair that does not control funds. For example, when users [claim their DOT tokens](https://claims.polkadot.network) after genesis, their DOT address doesn't have any funds yet, so that uses an unsigned transaction. Validators also submit unsigned transactions in the form of "heartbeat" messages to indicate that they are online. These heartbeats must be signed by one of the validator's [session keys](learn-keys). Session keys never control funds. Unsigned transactions are only used in special cases because, since Polkadot cannot charge a fee for them, each one needs its own, custom validation logic.
```

lsaether: The sentence about unsigned transaction applies to the post-genesis claims. Pre-genesis claims do involve a transaction on Ethereum.

lsaether (approved):
Just a few comments from me, otherwise looks good.

docs/learn-transaction-fees.md
joepetrowski: Will add a new line. I see the difference between bond and deposit as:
- Bond: Returned after some chain event (like the election ends).
- Deposit: Returned after some user event (frees index).

PR #265, opened 2020-02-17
Added era points and uncle blocks to payout doc

### LINKED ISSUE #130: Document uncle rewards
https://research.web3.foundation/en/latest/polkadot/Token%20Economics/#payment-details

_(the exact values of the point system are up for discussion. There are no good reasons for those exact values)_

Validators get paid for different actions. Both to the block producer who includes a reference to an uncle block, and the author of the uncle block are paid.

We pay more to the former. The logic is that the system doesn't really benefit from uncle blocks, but it does benefit from having a record that an uncle block exists, so we can detect equivocations (i.e. the block producer deciding to create two blocks at the same height). I guess it made sense in PoW where uncle rewards were a consolation for wasted energy to no fault of the miner (latency, instead), but here the cost of block production is nothing.

Best location - learn-dot? learn-consensus? Leaning towards learn dot, plus a mention in learn-consensus with link.

Swader:
Closes #130

### DIFF
--- a/docs/maintain-guides-validator-payout.md
+++ b/docs/maintain-guides-validator-payout.md
@@ -4,13 +4,26 @@ title: Validator Payout Overview
 sidebar_label: Validator Payout Overview
 ---
 
+## Era Points
+
+For every era (a period of time approximately 6 hours in length in Kusama, and 24 hours in Polkadot), validators are paid proportionally to the amount of _era points_ they have collected. Era points are reward points earned for payable actions like:
+
+- issuing validity statements for [parachain](learn-parachain) blocks.
+- producing a non-uncle block in the relay chain.
+- producing a reference to a previously unreferenced uncle block.
+- producing a referenced uncle block.
+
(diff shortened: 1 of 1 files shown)


docs/maintain-guides-validator-payout.md
```diff
@@ -4,13 +4,26 @@ title: Validator Payout Overview
...
+## Era Points
+
+For every era (a period of time approximately 6 hours in length in Kusama, and 24 hours in Polkadot), validators are paid proportionally to the amount of _era points_ they have collected. Era points are reward points earned for payable actions like:
+
+- issuing validity statements for [parachain](learn-parachain) blocks.
+- producing a non-uncle block in the relay chain.
+- producing a reference to a previously unreferenced uncle block.
+- producing a referenced uncle block.
+
+_Note: An uncle block is a relay chain block that is valid in every regard, but which failed to become canonical. This can happen when two or more validators are block producers in a single slot, and the block produced by one validator reaches the majority of the network before the others. We call the lagging blocks uncle blocks._
```
lsaether:
```suggestion
_Note: An uncle block is a relay chain block that is valid in every regard, but which failed to become canonical. This can happen when two or more validators are block producers in a single slot, and the block produced by one validator reaches the next block producer before the others. We call the lagging blocks uncle blocks._
```

lsaether (approved):

Issue #117, opened 2019-10-28
Slashing across eras with NPoS


ansonla3:
-add a section to explain how slashing would happen across eras

Source:  https://research.web3.foundation/en/latest/polkadot/slashing/npos/

ansonla3:
Reference: https://github.com/paritytech/substrate/pull/3846

lsaether:
Will be added as part of changes in #268

PR #268, opened 2020-02-17
Added offence levels and chilling

### LINKED ISSUE #178: Clarify offense levels and chills
There are 4 levels of offence in Polkadot:

1. Offline
2. Repeated offlineness; equivocation, solo (e.g. during a failover-to-backup)
3. Equivocation, repeat and/or wide-scale
4. 3 but which threatens the network?

Levels further documented here: https://research.web3.foundation/en/latest/polkadot/slashing/amounts.html

These levels and the percentages of maximum slashes applied should be added to learn-staking.

Additionally, some of these result in a validator being kicked, like so:

```
When a validator is found guilty of a misconduct:

a) We remove them from the list of candidates in the next NPoS validator election (for all misconducts).

b) We immediately mark them as inactive in the current era (for misconducts of levels 2 and up).

c) We remove them from all the nominators' lists of trusted candidates (for misconduct of levels 3 and up).
```

Swader:
Closes #178


lsaether:
The slashing research document is older than the slashing implementation. I would rather have the wiki reflect slashing as its implemented over what was specified in the research. Correlated unresponsiveness for instance carries up to 5% slash iirc.

Swader:
Ok I suppose let's wait for @AlfonsoCev to find out more about the implementation and translate it back to us, unless we want to go digging.

lsaether:
Should we resolve #117 here too or keep that a separate issue?

Swader:
Let's handle that here too.

docs/learn-staking.md
```diff
@@ -104,7 +104,14 @@ Slashing will happen if a validator misbehaves (e.g. goes offline, attacks the n
 
 Validator pools with larger total stake backing them will get slashed more harshly than less popular ones, so we encourage nominators to shift their nominations to less popular validators to reduce the possible losses.
 
-Based on Polkadot's latest codebase, the following slashing conditions have been implemented:
+The following levels of offence are [defined](https://research.web3.foundation/en/latest/polkadot/slashing/amounts.html):
+
+* Level 1: isolated unresponsiveness, i.e. going offline for a long time. No slashing, only [_chilling_](#chilling).
+* Level 2: concurrent unresponsiveness or isolated equivocation. Slashing up to 1% of the staked amount.
+* Level 3: misconducts unlikely to be accidental, but which do not harm the network's security to any large extent. Examples include concurrent equivocation or isolated cases of unjustified voting in [GRANDPA](learn-consensus). Slashing up to 10%.
+* Level 4: misconduct that poses a serious security or monetary risk to the system, or mass collusion. Slashing 100% of the stake behind the offenders.
```
laboon: These don't exactly line up with the way it's implemented.  We might want to keep the levels but not give the slashing percentages.  This way it's still "conceptually" correct without giving any false info.

docs/learn-staking.md
```diff
@@ -137,10 +144,16 @@ GRANDPA and BABE equivocation slashing penalty is calculated as below:
...
 
 > Notice:
-If a validator is reported for any one of the offences they will be removed from the validator set and they will not be paid while they are kicked out.
+If a validator is reported for any one of the offences they will be removed from the validator set ([chilled](#chilling)) and they will not be paid while they are kicked out.
 
 If you want to know more details about slashing, please look at our [research page](https://research.web3.foundation/en/latest/polkadot/slashing/amounts.html).
 
+### Chilling
+
+Chilling is the act of kicking a validator out of the active validator set, also disqualifying them from the set of electable candidates in the next NPoS cycle. This carries an implied penalty of being unable to earn rewards for the duration of the next era, and requires them to earn back any nominations they may have had before being kicked out.
```
laboon: I think "kicking" is too strong a word here, since you can "chill" yourself.  Maybe "removing"?

docs/learn-staking.md
```diff
@@ -137,10 +144,16 @@ GRANDPA and BABE equivocation slashing penalty is calculated as below:
...
+
+Because every offence regardless of level triggers a chill, and every chill triggers an automatic re-election of a new active validator set, the offending validator is immediately removed from the current set and the next, and someone else takes their place. To re-join, they need to acknowledge the slash and re-announce their candidature.
```
laboon, line 161: I thought this changed when we had the large-scale slashing, and we had all the nominators chilled?  This caused some problems with keeping validators in the active set.

laboon (changes requested):

docs/learn-staking.md
Swader: If I'm reading [this](https://github.com/paritytech/substrate/blob/d940c02440f80c5e0ded092c4dbc9f048312d8a6/frame/staking/src/slashing.rs#L327) correctly, every offence chills, but some also slash?

docs/learn-staking.md
Swader: Changed

Swader:
I have no idea what happened with the rebase here. I'll make a new PR and kill this. Meanwhile, could you check the Slashing Across Eras section in the learn-staking file of this PR please @laboon @lsaether @ansonla3

PR #270, opened 2020-02-18
Update CI

fgimenez:


Swader:
Looking much better! Just missing external 404 detection now!

```
This is an [invalid external link](https://foobarbaznonexistent.com.uk) for testing our CI.

This is an [invalid internal link](learn-something-that-doesnt-exist) for testing our CI.

This is an [external 404 link](https://bitfalls.com/something-non-existent) for testing our CI.
```

Right now only the second one is detected.

fgimenez:
I've come up with this regex for the ignore-url requirements `^(?:(?!http))(?:(?!\/en\/).)*$` it works fine when testing but through the linkchecker it doesn't ignore the required urls (those that don't have /en/ and don't start with http), will keep digging.

Swader:
This seems to work now!

Swader (approved):

PR #292, opened 2020-02-27
Fix kusama hijack

### LINKED ISSUE #289: Kusama section hijacks sidebar entries
When a major section includes a page ID also included by another major section, the former will lose that page in its menu. It gets overridden. See learn-identity. It's only visible in Kusama, not in Learn for Polkadot. This is probably the same for some other pages too.

lsaether:
closes #289

### DIFF
--- a/docs/mirror-build-build-with-polkadot.md
+++ b/docs/mirror-build-build-with-polkadot.md
@@ -1,0 +1,334 @@
+---
+id: mirror-build-build-with-polkadot
+title: Polkadot Builders Starter's Guide
+sidebar_label: Polkadot Builders Starter's Guide
+---
+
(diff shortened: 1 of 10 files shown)


Swader:
Can we add this to the CI so the mirroring is part of the build process and the mirrors are never out of date automatically? cc @fgimenez

laboon:
pinging @fgimenez ^

fgimenez:
Sorry, i totally missed this, so what should be done in CI?

lsaether:
I think we'll be able to keep CI the same and use [husky](https://github.com/typicode/husky) to add a git hook that runs the script on every commit. It only registers a change in git if there is an actual diff to that file.

PR #302, opened 2020-03-04
Removed dead link test file

Swader:
Removing the dead links file I accidentally merged, and also checking if the checker noticed other links. If it did, I'll also fix all of those in this PR.


Swader:
I don't understand why this isn't looking at my changes and keeps failing links. It also still looks at chinese translations, and uses the no-trailing-slash as the seed page, which makes relative links invalid (i.e. if you start from /en/ links work, but checker starts from /en). @fgimenez ideas?

fgimenez:
> I don't understand why this isn't looking at my changes and keeps failing links. It also still looks at chinese translations, and uses the no-trailing-slash as the seed page, which makes relative links invalid (i.e. if you start from /en/ links work, but checker starts from /en). @fgimenez ideas?

@Swader Yes, this PR https://github.com/w3f/polkadot-wiki/pull/270 was not finished and was merged, you seem to get the same failures we were getting on it.

I'll try to look into it soon, will ping you.

PR #310, opened 2020-03-09
revert CI link checker

### LINKED ISSUE #309: Fix CI link detection
Right now our builds are "failing" because the link detection is picking errors for valid links. 

Related #270 was not finished but merged. 

@fgimenez is there some quick fix we can do in the mean time to get green builds on link detection job?

fgimenez:
Closes https://github.com/w3f/polkadot-wiki/issues/309

Shows a single error that seems legit. Now it only checks local links with `/en/` in the url.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       - run:
           name: check links
           command: |
-            linkchecker --ignore-url="^(?:(?!http))(?:(?!\/en\/).)*$" --check-extern http://localhost:3000/
+            linkchecker --ignore-url="^(?:(?!\/en\/).)*$" --check-extern http://localhost:3000/
 
   deploy-website:
     docker:


Swader:
So with this we lose external link checks but internal works?

fgimenez:
> So with this we lose external link checks but internal works?

Yes exactly, this is what we had before the changes for including external links checks.

lsaether (approved):
Fixed the error in 410fd4db5de27b823f123bcaccff23874a028d2b. I'm happy with this until we have a working solution for external links too.

Swader:
@fgimenez rather than hack with regex, can we simply make another pass with just external link checks? Is there anything preventing us from making two link checker passes?

fgimenez:
The problem is that if we tell it to ignore everything that doesn't start with https in a separate command then it won't crawl the site, these are the results:
```
Start checking at 2020-03-10 09:30:43+000

URL        `http://localhost:3000/'
Real URL   http://localhost:3000/
Check time 0.014 seconds
D/L time   0.000 seconds
Size       380B
Result     Valid: 200 OK

URL        `en/index.html'
Name       `link'
Parent URL http://localhost:3000/, line 12, col 58
Real URL   http://localhost:3000/en/index.html
Info       The URL is outside of the domain filter, checked only
           syntax.
Result     Valid: filtered

URL        `docs/en'
Parent URL http://localhost:3000/, line 5, col 5
Real URL   http://localhost:3000/docs/en
Info       The URL is outside of the domain filter, checked only
           syntax.
Result     Valid: filtered
```

You can see that it doesn't follow `en/index.html` nor `docs/en` because they are outside the domain filter.

I'll try to continue digging, looks like linkchecker doesn't apply the provided regexp directly and does some changes that break the specific config that we require. Will keep you posted.

PR #321, opened 2020-03-18
Staking lazy payouts

laboon:
I updated the staking page to include information on claiming.  Additionally I have added a slightly modified version of @joepetrowski 's gist on lazy payouts as a separate page to explain the rationale.

There was also one reference to Alexander on here, I just removed it.


docs/learn-lazy-payouts.md
```diff
@@ -0,0 +1,46 @@
+---
+id: learn-lazy-payouts
+title: Lazy Payouts
+sidebar_label: Lazy Payouts
+---
+
+# Lazy Payouts
+
+Polkadot and Kusama make all stakers claim their rewards for past eras by submitting a transaction. This naturally leads to spreading out reward distribution, as people make transactions at disparate times, rather than updating the accounts of all stakers in a single block. Even if everyone submitted a reward claim at the same time, the fact that they are individual transactions would allow the block construction algorithm to process only a limited number per block and ensure that the network maintains a constant block time. If all rewards were sent out in one block, this could cause serious issues with the stability of the network due to too much activity taking place in one block.
```
joepetrowski: "Activity" is just too vague, and the actual problem (I/O) is explained in the next paragraph.
```suggestion
Polkadot and Kusama make all stakers claim their rewards for past eras by submitting a transaction. This naturally leads to spreading out reward distribution, as people make transactions at disparate times, rather than updating the accounts of all stakers in a single block. Even if everyone submitted a reward claim at the same time, the fact that they are individual transactions would allow the block construction algorithm to process only a limited number per block and ensure that the network maintains a constant block time. If all rewards were sent out in one block, this could cause serious issues with the stability of the network.
```

joepetrowski (approved):

PR #327, opened 2020-03-19
Title is duplicated

Swader:
Removed duplicate title from document - when an H1 is the first element in a doc, then it's two identical H1s in a row, one inherited from the doc's metadata, and one that's put in manually.


Swader:
Quick layout fix so just going to merge this.

PR #331, opened 2020-03-20
changes

lsaether:


website/static/js/custom.js
```diff
@@ -83,10 +83,11 @@ docReady(function () {
     // Insert "Improve this wiki ribbon"
     let link = document.createElement('a');
     link.classList = 'github-fork-ribbon';
-    link.setAttribute('href', 'contributing');
-    link.setAttribute('title', 'Improve this wiki');
+    link.setAttribute('href', 'https://wiki.polkadot.network/docs/en/contributing');
```
Swader, line 86: There's no need for an absolute URL. If you use relative, like "contributing" it stays on the currently selected language. Yours would force English even if someone is on Korean lang.

PR #348, opened 2020-03-31
Inject values

### LINKED ISSUE #136: Reading blockchain parameters from the chain
There are several bits of information in the wiki which get outdated easily, whenever a parameter in Kusama (and later Polkadot) changes. E.g. the validator count going from 25, to 50, to 100. The council mandate duration going from 100 blocks, to 2 hours, to 1 year(?). The number of parachain slots, the reward distribution, and more.

My suggestion is making a simple script we could slap into the docusaurus output template which would, on each render or each CI run fetch these values from the live blockchain and replace template values with them. For example, instead of writing "Council members remain in position for 1 year unless voted out by a referendum" you'd write "Council members remain in position for {{COUNCIL_DURATION || 1 year}}" and the script would fetch council duration straight from the chain, translate to human readable value, and replace the template. If it doesn't find a value it defaults to "1 year".

This would also be a nice side project, a super-lightweight JS package for reading chain params, which would then be usable by others as well (block explorers, for example).

This is maybe a good target for Hackusama?

Swader:
Supersedes #177, closes #136

I also took the opportunity to update the README to include mirroring instructions.


Swader:
If this is accepted and merged, I plan to update the script to support basic logic.

For example, bonding duration returns 28, in eras. To show that in a human readable way we should do `bonding_duration / 4`, so the update would be simple piping to basic arithmetic:

```md
{{ bonding_duration|/4 }}
```

lsaether:
Looks alright to me, are there any template strings in the docs already? If not can you add them for the ones you've included in the dict?

Swader:
On it.

ansonla3:
LGTM. With some examples on how to use it would be great. And when running with `node inject.js`,  I got 
```
node inject.js 
Connecting to node wss://kusama-rpc.polkadot.io/
2020-04-07 05:36:24        API/INIT: RPC methods not decorated: state_queryStorageAt}
```
Is that right?

Swader:
This is hopefully the last time we see those pesky translation files forced into our PRs. I have moved the mirror script away from the start command - mirror manually if you have to, because right now the mirroring is done automatically in CI. I made the injector auto-run after building (hopefully this translates to CI - @fgimenez?) and the injector only replaces values in the built files - the sources will keep templates. For an example, see [docs/learn-faq.md](docs/learn-faq.md).

lsaether:
We shouldn't be merging in mirrors at all since they're all generated during the build process. Can you remove the ones that snuck in here?

EDIT: We can also keep the mirror script in the start command I just added a fix to the .gitignore in master that will now ignore all mirror files included the nested translated ones.

lsaether (changes requested):
Tested locally, looks good. Just need the comments above resolved and it's good to merge.

lsaether (approved):

PR #349, opened 2020-03-31
CI update, mirror before build and start. Closes #347

### LINKED ISSUE #347: CI: add mirror script to build process
Before build and deploy, `node mirror.js` should be called so source files are copied into their mirrors. This will keep mirrors up to date.

Swader:
Updated start and build script to mirror before executing. Also recommend moving mirror.js to website so it can interact with the packages installed by docusaurus itself and not duplicate dependencies outside of the folder (in case our custom scripts get a little more complex).


lsaether (approved):
Nice, thanks

PR #358, opened 2020-04-03
Offence levels and cross-era slashing

### LINKED ISSUE #117: Slashing across eras with NPoS

Swader:
Closes #117 and #178

Cross-era slashing basically copied from the PR, but not sure on how it affected the chilling paragraph above, would appreciate a review on that.


docs/learn-staking.md
```diff
@@ -26,7 +26,7 @@ A validator node is required to be responsive 24/7, perform its expected duties
...
+Once the nomination period ends, the NPoS election mechanism takes the nominators and their associated votes as input, and outputs a set of validators of the required size, that maximizes the stake backing of any validator, and that makes the stakes backing validators as evenly distributed as possible. The objectives of this election mechanism are to maximize the security of the network, and achieve fair representation of the nominators. If you want to know more about how NPoS works (e.g. election, running time complexity, etc.), please read [here](http://research.web3.foundation/en/latest/polkadot/NPoS.html).
```
laboon: Do you think it makes sense to link to the Phragmen page here?

docs/learn-staking.md
```diff
@@ -137,10 +144,26 @@ GRANDPA and BABE equivocation slashing penalty is calculated as below:
...
 
 > Notice:
-If a validator is reported for any one of the offences they will be removed from the validator set and they will not be paid while they are kicked out.
+If a validator is reported for any one of the offences they will be removed from the validator set ([chilled](#chilling)) and they will not be paid while they are kicked out.
 
 If you want to know more details about slashing, please look at our [research page](https://research.web3.foundation/en/latest/polkadot/slashing/amounts.html).
 
+### Chilling
+
+Chilling is the act of removing a validator from the active validator set, also disqualifying them from the set of electable candidates in the next NPoS cycle. This carries an implied penalty of being unable to earn rewards for the duration of the next era, and requires them to earn back any nominations they may have had before being removed.
```
laboon: We might want to add that this is not necessarily a penalty - a validator may choose to self-chill because, e.g., they know that a hurricane is coming which may disrupt their network connection.

docs/learn-staking.md
```diff
@@ -137,10 +144,26 @@ GRANDPA and BABE equivocation slashing penalty is calculated as below:
...
+
+Because every offence regardless of level triggers a chill, and every chill triggers an automatic re-election of a new active validator set, the offending validator is immediately removed from the current set and the next, and someone else takes their place. To re-join, they need to acknowledge the slash and re-announce their candidature.
```
laboon, line 252: I thought that this changed recently?  We had the issue where a huge group got chilled and all the nominators lost the ability to nominate them, causing a chain reaction which pushed down the amount of stake in the system.

docs/learn-staking.md
```diff
@@ -220,8 +245,8 @@ please see [here](https://research.web3.foundation/en/latest/polkadot/Token%20Ec
 
 ## Why not stake?
 
-* Tokens will be locked for 28 days after unbonding.
-* Tokens can be slashed as punishment if the validator is not running properly.
+* Tokens will be locked for about 12 weeks on Polkadot, seven days on Kusama
```
laboon: "12 weeks on Kusama" -> is this set in stone?  I thought most Kusama params were set to 4x as fast as Polkadot.

laboon (changes requested):

docs/learn-staking.md
Swader: I thought about it but decided against it, as it's not referencing just that part of the election specifically. Also, I think this document will need a further revamp soon, or a separate "basic" version built, because this is possibly the most essential doc we have for new ecosystem participants and it's complex beyond readability for most people not already familiar with most mechanics here. I think we should have learn-staking-basics that dumbs this down a lot and removes a lot of stuff like this paragraph.

docs/learn-staking.md
Swader: Good thinking

docs/learn-staking.md
Swader: This was not my addition, I added the seven days. Let me check.

docs/learn-staking.md
Swader: I have been unable to verify this yet. @AlfonsoCev can you shed some light perhaps?

docs/learn-staking.md
Swader: Checked in code, it's 28 eras so 28 days.

docs/learn-staking.md
kianenigma: I recommend looking into these unit tests in `frame/staking`

- `offence_deselects_validator_even_when_slash_is_zero`
- `nomination_slash_filter_is_checked`
- `zero_slash_keeps_nominators`

and what I can add is: 
- If a validator is slashed, the validator is chilled, but the nominators are not. And there is an additional filter that checks if a validator has been slashed recently, then the nominator's who voted for the slashed validator  _before the slashing event_ are actually dropped (only the vote is dropped, not the entire nominator). 

I don't quite know when this recovers and when the validators can become effective again and re-gain the lost votes. @rphmeier will know more. 

```rust
targets.retain(|stash| {
	<Self as Store>::SlashingSpans::get(&stash).map_or(
		true,
		|spans| submitted_in >= spans.last_nonzero_slash(),
	)
});
```

And I know from experience that the easiest way to unravel such details from the code is to add new unit tests that check the hypothesis.

Swader: Useful, thanks!

AlfonsoCev: I think we should clarify in the text that chilling and slashing are not necessarily linked. As Bill said, a validator can choose to chill if they need to stop their node for a while. In that case they continue being active in the current era and are inactive from the next era onward, and they don't lose any nominations. 

If I understand correctly, when there's an offense event chilling is just one of the consequences, being ignored for the remainder of the current era is another consequence, and losing nominations is another one (but this only for non-zero slashings).

rphmeier: > I don't quite know when this recovers and when the validators can become effective again and re-gain the lost votes

At the moment, it does not. The nominators must re-issue their votes on those validators to re-activate.

laboon (approved):

laboon:
👍 feel free to merge

PR #370, opened 2020-04-08
Update council elections

### LINKED ISSUE #367: Update council election
It should be Phragmén to elect the council instead of the approval voting.
https://wiki.polkadot.network/docs/en/learn-governance#how-to-be-a-council-member

lsaether:
closes #367 

This does not update the diagram for approval voting, since I'm not sure we have a diagram for the Phragmen elections.


laboon:
👍 

Phragmen's method is really a subset of the broader term "approval voting", I believe.  From a "user voting" perspective, a straight-up approval vote would look exactly the same.  So I think it's not a big issue for now.

laboon:
lgtm if you want to merge

PR #384, opened 2020-04-15
Add the secure validator guide to the sidebar

lsaether:


lsaether:
Merging right away since the article is just missing from the sidebar.

PR #401, opened 2020-04-23
Crowdfund section

### LINKED ISSUE #380: Add explainer for crowdfunding module
https://github.com/paritytech/polkadot/blob/master/runtime/common/src/crowdfund.rs

From Fabi at Parity - "Talking to parachain teams, it seems like they a lot of them want to make use of this feature. Probably worth getting an explainer up."

lsaether:
closes #380

### DIFF
--- a/docs/learn-parachains.md
+++ b/docs/learn-parachains.md
@@ -24,6 +24,47 @@ Transaction fees in a native parachain token can also be an implementation choic
 
 Parachains are not required to have their own token. If they do, is up to the parachain to make the economic case for their token, not Polkadot.
 
+## Crowdfunding parachains
+
+Polkadot allows for parachains to  crowdfund their slots in a decentralized and
+safe way. The logic for this is handled in the [crowdfunding pallet][]. 
+
+During a parachain auction, anyone can create a new crowdfunding campaign for
+a parachain slot. When a campaign is created, the range of slots (i.e. the
+duration of the lease) is specified. Up to four slots, for a total time duration
+of roughly two years, can be selected. The creator of the crowdfund becomes the
+owner of the campaign, and can later upload the parachain's code. When creating
+a campaign, a crowdfunding "cap" is also specified. The crowdfund will refuse
+to accept funds after the cap has been reached.
+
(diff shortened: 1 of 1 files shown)


docs/learn-parachains.md
```diff
@@ -24,6 +24,47 @@ Transaction fees in a native parachain token can also be an implementation choic
...
+Polkadot allows for parachains to  crowdfund their slots in a decentralized and
+safe way. The logic for this is handled in the [crowdfunding pallet][]. 
+
+During a parachain auction, anyone can create a new crowdfunding campaign for
+a parachain slot. When a campaign is created, the range of slots (i.e. the
+duration of the lease) is specified. Up to four slots, for a total time duration
+of roughly two years, can be selected. The creator of the crowdfund becomes the
+owner of the campaign, and can later upload the parachain's code. When creating
+a campaign, a crowdfunding "cap" is also specified. The crowdfund will refuse
+to accept funds after the cap has been reached.
```
Swader: It is unclear to me how it makes sense to cap a pool of funds that is supposed to participate in an auction. Does this not essentially reveal to your competition how much maximum money you have and, thus, how much is needed to reliably beat you?

Swader (changes requested):
Just an uncertainty

docs/learn-parachains.md
lsaether: Yeah, not sure whether it will be useful to place a cap on the funds in practice. The only situation I can see is when a parachain is confident they'll raise the minimum amount for a slot but doesn't want to extend too far over the minimum. It would be possible to calculate cap minimums depending on total supply and the total number of parachain slots open for auction.

For example, total supply is 10m and parachain slots is 4. Roughly 5m is locked in staking. So the minimum cap would be ~1.25M DOTs. A parachain could cap it there and still be guaranteed a slot.

Swader: Sounds good - let's include this in the text too?

lsaether: Added

Swader (approved):

PR #410, opened 2020-04-27
Added fathom

Swader:
Let's see how fathom treats us for analytics. It's not a free service because it doesn't sell people's data. It's the ethical tradeoff we'd have to agree with. It's $140 per year, or $14 per month. Steep, but let's see if it's worth it?

### DIFF
--- a/website/static/js/custom.js
+++ b/website/static/js/custom.js
@@ -89,4 +89,12 @@ docReady(function () {
     link.innerText = 'Improve this wiki';
     document.querySelector('body').appendChild(link);
     // -----------------
+
+    // Add fathom to footer
+    
+    let script = document.createElement('script');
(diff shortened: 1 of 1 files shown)


laboon:
Let me talk with finance, but I don't foresee a problem with such a small cost.

Swader:
Let's get this in first during the 1-week trial, we can upgrade if we like it. The trial is already ongoing.

laboon:
lgtm, let's try it

ansonla3:
LGTM, but curious what is the difference between google analytics?

Swader:
@ansonla3 GA is free but not private, fathom is private but not free. I.e. fathom is more into preserving privacy and more in line with our ethos of people owning their own data.

PR #422, opened 2020-05-02
Updated multisig image

Swader:


Swader:
Going to merge this since it's the weekend and the change is pretty jarring.

PR #427, opened 2020-05-04
Add flag for session key generation

xz-cn:
As of V0.7.32, in order to submit RPC calls like rotating session keys, additional flag `--unsafe-rpc-expose` is needed.


Swader:
Thanks!

PR #429, opened 2020-05-04
Use Prettier Markdown Formatting

joepetrowski:
Markdown pages in this repo are a complete mess. We started using Prettier in the Knowledge Base and it makes life much easier. Running it is a simple one-line command (see update in README, 3rd commit). For example,

- Automatically wraps lines at a fixed (100, to be consistent with Substrate) column width.
- Removes trailing spaces.
- Makes consistent spacing between header lines and plain text (no zero/double).
- Formats tables so they are aligned in monospace.

For example, this:
image

Becomes:
image


Swader:
Sounds good, we've discussed prettyfication before.

Swader (approved):

joepetrowski:
OK, going to merge since it will cause merge conflicts with other PRs. If people are really against it we can revert. But in Knowledge Base this has really been a nice addition and it's so easy to use.

PR #442, opened 2020-05-07
Enforce prettier

lsaether:


lsaether:
Merging as change is inconsequential and only to CI.

Swader:
I don't think it's necessary to prettify our build since that's not the MD we edit. We should be running prettier during pre-commit, not during deployment.

lsaether:
It runs now on CI so we don't get a green build if it hasn't been formatted.

We will have to yell at anyone who tries to PR changes that doesn't use prettier. The aim here was just to make it apparent if files aren't being formatted so we can fix it if we need too. We should only be merging green CI builds anyway

Swader:
Makes sense!

PR #449, opened 2020-05-13
Add compatibility and mortality notes

joepetrowski:
Needs review from @kirushik


docs/learn-accounts.md
```diff
@@ -97,6 +97,22 @@ and not officially affiliated with Web3 Foundation or the Polkadot project unles
...
+
+Seed phrases are not portable across all wallets. Some wallets only support certain signing curves
+and derivation paths. See the following table for compatibility.
+
+|                  | sr25519 |  ed25519  | secp256k1 |
+| :--------------- | :-----: | :-------: | :-------: |
+| Polkadot JS UI   |    X    |     X     |     X     |
+| Polkadot JS Apps |    X    |     X     |     X     |
+| Parity Signer    |    X    |    X\*    |           |
+| Ledger           |         | X&dagger; |           |
```
lsaether: Is this the Ledger app made by Zondax for Kusama? Or is it a generic ed25519 app on Ledger?

docs/learn-accounts.md
```diff
@@ -97,6 +97,22 @@ and not officially affiliated with Web3 Foundation or the Polkadot project unles
 
 Hardware wallet integration from [Ledger](https://ledger.com) is coming soon.
```
lsaether:
```suggestion
Full hardware wallet integration on [Ledger](https://ledger.com) is coming soon.
```

docs/learn-accounts.md
kirushik: Ledger app by Zondax currently only supports Ed25519+BIP32 hard derivations with `44'/434'` hardcoded prefix: https://github.com/Zondax/ledger-kusama/blob/master/app/src/coin.h#L27-L28.

They are planning to add sr25519, but AFAIU it will be covered by another round of grants by Web3 Foundation.

docs/learn-accounts.md
```diff
@@ -95,7 +95,23 @@ Alternatively, you might find some wallets on the
...
+|                  | sr25519 |  ed25519  | secp256k1 |
+| :--------------- | :-----: | :-------: | :-------: |
+| Polkadot JS UI   |    X    |     X     |     X     |
+| Polkadot JS Apps |    X    |     X     |     X     |
+| Parity Signer    |    X    |    X\*    |           |
+| Ledger           |         | X&dagger; |           |
+| Subkey           |    X    |     X     |     X     |
+
+- \* Planned
+- &dagger; Only with BIP32 derivations
```
kirushik: Maybe we need some other form more explicitly telling which kind of deriations (hard or soft and using which protocol) each of the implementations uses?
So far this table doesn't answers the question "OK, I have seed and a derivation path from Ledger, which tools I can use to properly import it"?
It seems that Polkadot JS UI would support the BIP32-derivation, but that (as far as I can tell) is not correct.

joepetrowski: Does Ledger have documentation about these paths? In this case, I think the Wiki should point people to the resources needed to figure out what is portable. There's no way we can keep up with every wallet's implementation. I would rather use this to warn people that portability is an issue and they need to look at X and Y resources to figure out if they can port a seed phrase from A to B.

kirushik: Makes sense. @jleni, do you have some Kusama/Polkadot-on-Ledger documentation we can point to from here?

docs/learn-accounts.md
jleni: The app is for Kusama, we currently support runtime 1062 and it is already integrated in Polkadot JS. It can be used for development/testing today. At the moment, Ledger is reviewing the app and preparing it for official publication.

It is not possible to have a generic Ed25519 as per guidelines. Apps need to be able to deserialize txs, so are chain specific.

Signature schemes:
- We support Ed25519
- Adding secp256k1 would be very easy. I don't know this could be useful. Does it make sense we add it?
- We are currently working on sr25519 support.

- Key Derivation:
We follow BIP32/BIP44 derivation, also standard in most Ledger apps. 
We actually enforce hard derivation only for the initial part, very much aligned with BIP44.

docs/learn-accounts.md
jleni: > Does Ledger have documentation about these paths? In this case, I think the Wiki should point people to the resources needed to figure out what is portable. There's no way we can keep up with every wallet's implementation. I would rather use this to warn people that portability is an issue and they need to look at X and Y resources to figure out if they can port a seed phrase from A to B.

I already highlighted mnemonic portability is a problem, if things are not clear. In my opinion, BIP32/BIP44 is a well standardized way to deriving keys. If every projects does their own unvetted derivation is can be very bad.

If key derivation is not clear and open and works like a blackbox, a bad derivation could actually drop a lot of the entropy or make bad things..

jleni: In my opinion, the best option is to make this clear, standardize and avoid too many options.

jleni: The derivation we use in the Ledger app for Ed25519 is the same that are used in other projects such as Tendermint-based chains, Cardano, etc. As a reference, Ledger has made public a simple python implementation that can be used to verify the derivation independently. https://github.com/LedgerHQ/orakolo/blob/master/src/python/orakolo/HDEd25519.py

For Ed25519 there are currently two options:
- Ed25519+BIP32 based on Khovratovich [Link](https://github.com/LedgerHQ/orakolo/blob/master/papers/Ed25519_BIP%20Final.pdf)
- SLIP-0010

As Cardano, Ledger opts to use Ed25519+BIP32 based on Khovratovich as default. This allows both hard and soft derivations unlike SLIP-0010. As already explained, the Kusama and Polkadot apps use Ed25519+BIP32 based on Khovratovich too.

By the way, there is a third thing that should be made explicit and this is the `mnemonic -> seed` step. We are using standard BIP39. If people use a different way to go from the wordlist to a seed, this can also affect mnemonic portability.

jleni: Ok, I summarize: I believe it very important to make explicit that there are actually multiple steps along the process. 
All these should be documented, not only for Ledger devices but for every other wallet.

`Path + Mnemonic -> Seed -> Signature Scheme -> Signature`

1. Paths: The Ledger app uses BIP44. (we used the registered SLIP-0044 values for Kusama & Polkadot [here](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) 
2. Mnemonic: The Ledger app uses BIP39
3. Derivation: The Ledger app uses Ed25519+BIP32 based on Khovratovich
4. Signature Scheme: We use Ed25519

I think these 4 points should be clarified for all wallets.

jleni (changes requested):

laboon:
Are we still waiting on a review here?

joepetrowski:
> Are we still waiting on a review here?

I still need to clean up a few things here. Leave it open for now.

laboon:
@joepetrowski Should we still leave this PR open?

joepetrowski:
> @joepetrowski Should we still leave this PR open?

Yes please, I actually just pointed someone to it yesterday to reference. I will clean it up and finish it soon.

lsaether:
superseded by #818

PR #453, opened 2020-05-13
Re-do Phragmén optimizations section based on new offchain worker / edge-reducing optimization

### LINKED ISSUE #443: Phragmen update
Phragmen is no longer sequential as we described earlier. From Kian (parahprased):

> In the last quarter of each session leading to a new era, some staking operations are disallowed. This is to make sure election solutions that are submitted in the meanwhile stay valid. Most notably:
> 
> - bond_extra, unbond, withdraw_unbonded, rebond
> 
> - validate, nominate and chill
> 
> - payouts
> 
> are disallowed.
> 
> Furthermore, off-chain workers run an optimization on the output before submitting it to reduce the size, without changing any validator's backing stake (compared to the good old sequential phragmen, which is what you were getting prior to this update), but this will further change how the nominator's stake is distributed:
> 
> So far, in sequential Phragmen (old Phragmen), if you vote for two validators and both of them are winners, almost always your stake is distributed among the two of them. One might get more stake, but both get something. reduce() in new Phragmen is very likely to remove as many edges (i.e. votes) as possible. So all in all, it is likely that if I vote for two validators, both of whom are in the winners set, all of my stake will actually be given to one of them.

laboon:
Re-did optimizations section explanation based on new edge-reducing optimization.

Did some re-organizing and went into more depth on rationale, as well.

Fixes #443

### DIFF
--- a/docs/learn-phragmen.md
+++ b/docs/learn-phragmen.md
@@ -610,12 +610,13 @@ an optimal manner.
 
 There are several further restrictions put in place to limit the complexity of the election and
 payout. As already mentioned, any given nominator can only select up to 16 validators to nominate.
-Conversely, a single validator can have only 64 nominators. A drawback to this is that it is 
-possible, if the number of nominators is very high or the number of validators is very low, that
-all available validators may be "saturated" and unable to accept more nominations. In this case,
-one may need a larger amount of stake to participate in staking, since nominations are 
-priority-ranked in terms of amount of stake. One can estimate the number of nominators for a 
-specific validator in the next era by running the [`offline-phragmen`](https://github.com/kianenigma/offline-phragmen) program.
+Conversely, a single validator can have only 64 nominators. A drawback to this is that it is
+possible, if the number of nominators is very high or the number of validators is very low, that all
+available validators may be "saturated" and unable to accept more nominations. In this case, one may
+need a larger amount of stake to participate in staking, since nominations are priority-ranked in
+terms of amount of stake. One can estimate the number of nominators for a specific validator in the
+next era by running the [`offline-phragmen`](https://github.com/kianenigma/offline-phragmen)
+program.
 
 ## External Resources
 
--- a/docs/learn-phragmen.md
+++ b/docs/learn-phragmen.md
@@ -614,9 +614,7 @@ Conversely, a single validator can have only 64 nominators. A drawback to this i
 possible, if the number of nominators is very high or the number of validators is very low, that all
 available validators may be "saturated" and unable to accept more nominations. In this case, one may
 need a larger amount of stake to participate in staking, since nominations are priority-ranked in
-terms of amount of stake. One can estimate the number of nominators for a specific validator in the
-next era by running the [`offline-phragmen`](https://github.com/kianenigma/offline-phragmen)
-program.
+terms of amount of stake. 
 
 ## External Resources
 
--- a/docs/learn-phragmen.md
+++ b/docs/learn-phragmen.md
@@ -614,7 +614,7 @@ Conversely, a single validator can have only 64 nominators. A drawback to this i
 possible, if the number of nominators is very high or the number of validators is very low, that all
 available validators may be "saturated" and unable to accept more nominations. In this case, one may
 need a larger amount of stake to participate in staking, since nominations are priority-ranked in
-terms of amount of stake. 
+terms of amount of stake.
 
 ## External Resources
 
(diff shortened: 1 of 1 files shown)


docs/learn-phragmen.md
```diff
@@ -526,18 +526,68 @@ their candidates selected will also not have any stake in any of the elected can
...
+incurs some diversification costs. If the single validator that a nominator has nominated goes
+offline or acts maliciously, then the nominator incurs a risk of a significant amount of slashing.
+Nominators are thus allowed to nominate up to 16 different validators. However, after the weighted
+edge-reducing algorithm is run, the number of validators per nominator is minimized. Nominators are
+likely to see themselves nominating a single active validator for an era.
+
+At each era change, as the algorithm runs again, nominators are likely to have a different validator
+than they had before (assuming a significant number of selected validators). Therefore, nominators
+can diversify against incompetent or corrupt validators causing slashing on their accounts, even if
+only nominate a single validator is nominated by them per era.
```
Swader:
```suggestion
only a single validator is nominated by them per era.
```

Swader: Consider changing to active voice

docs/learn-phragmen.md
```diff
@@ -526,18 +526,68 @@ their candidates selected will also not have any stake in any of the elected can
...
+
+### Rationale for Minimizing the Number of Validators Per Nominator
+
+Paying out rewards for staking from every validator to all of their nominators can cost a
+non-trivial amount of chain resources (in terms of space on chain and resources to compute). Assume
+a system with 200 validators and 1000 nominators, where each of the nominators has nominated 10
+different validators. Payout would thus require `1000 * 10`, or 10,000 transactions. In an ideal
+scenario, if every nominator selects a single validator, only 1,000 transactions would need to take
+place - an order of magnitude fewer. Empirically, network slowdown at the beginning of an era has
+been seen due to the large number of individual payouts by validators to nominators. In extreme
```
Swader: Consider changing from passive "has been seen"

docs/learn-phragmen.md
```diff
@@ -547,20 +597,20 @@ expensive proposition. Although obtaining an equal distribution is unlikely, the
...
 
-These optimizations will not be covered in-depth on this page. For more details, you can view the
-[Rust implementation in Substrate](https://github.com/paritytech/substrate/blob/master/frame/elections-phragmen/src/lib.rs)
-or the `seqPhragménwithpostprocessing` method in the
-[Python reference implementation](https://github.com/w3f/consensus/tree/master/NPoS). If you would
-like to dive even more deeply, you can review the
-[W3F Research Page on Sequential Phragmén Method](https://research.web3.foundation/en/latest/polkadot/NPoS/4.%20Sequential%20Phragm%C3%A9n%E2%80%99s%20method.html).
+There are several further restrictions put in place to limit the complexity of the election and
+payout. As already mentioned, any given nominator can only select up to 16 validators to nominate.
+Conversely, a single validator can have only 64 nominators.
```
Swader: Should we warn people about checking that the validators they're nominating aren't already saturated or they'd be wasting nominations?

Swader (changes requested):
Nits

docs/learn-phragmen.md
laboon: Do you know an easy way to do this?  We won't know that a validator is saturated until after the Phragmen election, and you have to nominate well before then.  One could make guesses ahead of time by running offline-Phragmen, I suppose.

I don't think it will be a big deal, at least at first, since the edge-reductions minimization reduces this problem considerably.  On Kusama, for instance, we can have 225 * 64 (14,400) nominating accounts.  I don't THINK we're getting close to that (but I am having trouble getting the kusama-validator-stats script to work right now...)

I can mention it here, though.

docs/learn-phragmen.md
laboon: Changed to "occurred"

docs/learn-phragmen.md
laboon: changed to ".. even if they only nominate a single validator per era."

laboon:
What do you think of the changes @Swader ?

Swader:
Offline phragmen is broken right now IIRC, so linking to it won't do much good. We should hold off on that until it's fixed. All good otherwise 👍

Swader (approved):

laboon:
Removed that reference and merging.  Thanks Bruno!

PR #455, opened 2020-05-14
Clarify address vs pubkey in Kusama claims

laboon:
There was some confusion from Syed

### DIFF
--- a/docs/kusama-claims.md
+++ b/docs/kusama-claims.md
@@ -82,7 +82,8 @@ Secret phrase `lobster flock few equip connect boost excuse glass machine find w
   Address (SS58): JL1eTcbzuZP99FjeySkDrMygNREPdbhRyV7iD5AsV4fDRcg
 ```
 
-The `Address (SS58)` field is what you should use to claim your KSM tokens. Never share your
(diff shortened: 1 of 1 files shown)


docs/kusama-claims.md
```diff
@@ -82,8 +82,9 @@ Secret phrase `lobster flock few equip connect boost excuse glass machine find w
   Address (SS58): JL1eTcbzuZP99FjeySkDrMygNREPdbhRyV7iD5AsV4fDRcg
 ```
 
-The `Address (SS58)` field is what you should use to claim your KSM tokens. Never share your
-`Secret phrase` or `Secret seed`, as these can both control your funds.
+The `Address (SS58)` field is the address you will claim your KSM tokens. You will use the Public
```
Swader, line 85:
```suggestion
The `Address (SS58)` field is the address to which you will claim your KSM tokens. You will use the Public
```

Swader (changes requested):

laboon:
This is of course only if you are trying to do this without actually selecting an account.  Closing because this is a rare thing and I tried to do it too quickly.

PR #458, opened 2020-05-15
China redirect

### LINKED ISSUE #439: Geolocate a visitor and land them on the translated content or their zone

Swader:
Closes #439 

@ansonla3 could you build this locally please and see if it works?

It should:

- redirect you from `en` to `zh-CN` if you were never redirected before
- not redirect you if you're already on `zh-CN`
- not redirect you from `en` if you were already redirected once

### DIFF
--- a/website/static/js/custom.js
+++ b/website/static/js/custom.js
@@ -96,4 +96,14 @@ docReady(function () {
     script.setAttribute('src', 'https://mammal.polkadot.network/script.js');
     script.setAttribute('site', 'YOUKYIQA');
     document.querySelector('body footer').appendChild(script);
+
+    // Redirect if Chinese
+    fetch("https://wiki.polkadot.network/countrycheck").then(result => {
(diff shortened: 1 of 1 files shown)


ansonla3:
the logic seems right, but I found there is a redirection problem between switching languages. will take a look at that first.

Swader:
What kind of problem?

ansonla3:
have you tried to test with locally and switch the language to Chinese?

lsaether:
I found this post the other day, do you think it would be helpful here?

https://dev.to/bitdweller/stop-setting-the-language-of-your-website-based-on-my-location-31h0

Swader:
I agree with that post in the western world, but I don't think it applies to China. @ansonla3?

ansonla3:
@Swader this approach makes sense to me.

ansonla3:
I will close this PR since it has been resolved already.

Issue #434, opened 2020-05-05
Add Timeline Page for Polkadot launch


laboon:
- [ ] Explain terminology
- [ ] Describe steps and transitions 
- [ ] Explain what is possible in PoA phase
- [ ] Describe how we move from PoA to PoS
- [ ] Explain what is possible in PoS phase
- [ ] Include link from front page

lsaether:
The [FAQ](https://wiki.polkadot.network/docs/en/learn-faq#polkadot-launch) page is a start on some of these points.

laboon:
We can build on Logan's article here: https://polkadot.network/explaining-the-polkadot-launch-process/

PR #460, opened 2020-05-15
Launch page

Swader:
Mainly copy pasting with some more details thrown in and some time limits from the chain itself. Let me know if this is too discouraging to show to the public (i.e. it'll take a year to have a fully unlocked mainnet if we don't change the enactment delay and voting periods).

### DIFF
--- a/docs/learn-launch.md
+++ b/docs/learn-launch.md
@@ -41,8 +41,6 @@ Time-wise, the transition to Proof-of-Stake will largely depend on when a suffic
 validators have registered and are ready to take over the security of the network. This number can
 be as low as 50 but probably closer to 100, as it was with [Kusama](https://kusama.network).
 
-This phase will last for at least two months.
-
 ## Sudo to Governance
 
 After the chain has been running well with a sufficiently large validator set, the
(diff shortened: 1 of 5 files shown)


Swader:
Closes #434

docs/learn-launch.md
```diff
@@ -0,0 +1,83 @@
...
+election. This election will transition the network from PoA into its second phase,
+[Nominated Proof of Stake (NPoS)](learn-staking), where the network is
+[secured by the economic stake](learn-security) that is bonded to the validators by the
+[nominators](maintain-nominator) and the validators themselves.
+
+Time-wise, the transition to Proof-of-Stake will largely depend on when a sufficient number of
+validators have registered and are ready to take over the security of the network. This number can
+be as low as 50 but probably closer to 100, as it was with [Kusama](https://kusama.network).
+
+This phase will last for at least two months.
```
lsaether: Why?

Swader: It took a month on kusama IIRC so just doubling that. You don't think so?

lsaether: I think it could happen more quickly this time around since there's already a large validator community built on Kusama, many of whom will be ready to set up similar infrastructure for Polkadot.

Swader: I'm fairly confident more than 100 will signal in the first week already, but I think Gav will want to play it safe and not jump on that signal. You're right though, it's unrealistic to estimate *anything* in this phase, I'll remove this part.

lsaether (approved):

PR #472, opened 2020-05-20
Remove faucet

laboon:
This goes through and either removes references to the Kusama faucet or speak of it in the past tense.  It also adds an explanation (based on the approved tweet wording) of why it was removed.


laboon:
Additionally, David reminded me that KSMs are not given out through the Grants process.

PR #505, opened 2020-05-28
Fixes to secure websocket

igorline:
Fixed root path for nginx to default location otherwise websockets connection will be failing due to 404


Swader:
Wonderful, thank you! Looking at this guide in full again today to fix some other problems people have reported but this is an amazing start!

PR #528, opened 2020-06-03
Kirsten add wallet

krichard410:
Created a Wallet page for Polkadot Wiki


docs/wallets.md
```diff
@@ -0,0 +1,41 @@
+---
```
laboon: This probably belongs to build so should be named build-wallets.md

Please add a link to this page on the sidebar ./website/sidebars.json

docs/wallets.md
```diff
@@ -0,0 +1,41 @@
...
+title: Polkadot Wallets
+sidebar_label: Polkadot Wallets
+---
+
+## Polkadot Wallets
+
+The integration of a [wallet](https://wiki.polkadot.network/docs/en/glossary#wallet) and Polkadot
+allows for simple and easy access to private keys and signing transactions. Below are wallets
+supported by Polkadot.
+
```
laboon: We should mention polkadot.js.org, Parity Signer, and polkadot.js plugin here and have them at the top.

docs/wallets.md
```diff
@@ -0,0 +1,41 @@
...
+### Live Wallets
+
+- [Polkawallet](https://polkawallet.io/) - Mobile wallet available on IOS and Android
```
laboon: Let's put these all in a table.

We don't want to endorse any specific wallet, but should include as much detail for people to make up their mind as possible.

You should also double-check that these work before posting them.

docs/wallets.md
```diff
@@ -0,0 +1,41 @@
...
+- [NGRAVE](https://www.ngrave.io/) - Hardware wallet
+- [Trust Wallet](https://trustwallet.com/) - Mobile app available on IOS and Android
+- [AToken](https://www.atoken.com/) - Mobile app available on IOS and Android
+- [Qbao Network](https://qbao.fund/) - Mobile app available on IOS and Android
+- [Node Factory](https://nodefactory.io/) - Metamask wallet plugin
+
+### Wallets still being Explored
+
+- [Token Pocket](https://www.tokenpocket.pro/en/) - Mobile and desktop wallet
+- [Atomic Wallet](https://atomicwallet.io/) - Mobile and desktop wallet
```
laboon: If any of the wallets say that we haven't gotten in touch with them recently, let's not add them quite yet.

laboon (changes requested):

docs/wallets.md
```diff
@@ -0,0 +1,41 @@
...
+id: wallets
+title: Polkadot Wallets
+sidebar_label: Polkadot Wallets
+---
+
+## Polkadot Wallets
+
+The integration of a [wallet](https://wiki.polkadot.network/docs/en/glossary#wallet) and Polkadot
+allows for simple and easy access to private keys and signing transactions. Below are wallets
+supported by Polkadot.
```
Swader: Wallets support Polkadot, not the other way around :)

docs/wallets.md
```diff
@@ -0,0 +1,41 @@
...
+  [GitHub](https://github.com/GetSpeckle/speckle-browser-extension)
+- [AirGap](https://airgap.it/) - Desktop and mobile wallet. Development found on
+  [GitHub](https://github.com/airgap-it/airgap-wallet)
+- [NGRAVE](https://www.ngrave.io/) - Hardware wallet
+- [Trust Wallet](https://trustwallet.com/) - Mobile app available on IOS and Android
+- [AToken](https://www.atoken.com/) - Mobile app available on IOS and Android
+- [Qbao Network](https://qbao.fund/) - Mobile app available on IOS and Android
+- [Node Factory](https://nodefactory.io/) - Metamask wallet plugin
+
+### Wallets still being Explored
```
Swader: What does "still being explored" mean? Also be mindful of the caps.

Swader (changes requested):
Good list. Agreed with Bill that we should have some kind of table with features, but that might be biting more than we can collectively chew given how fast things change and how much maintenance that would need. Maybe just a table for all of them with a "status" column, differentiating between live, in development, abandoned, unknown. Then we can think about other columns later, like governance, chain support, identity, etc.

krichard410:
> Good list. Agreed with Bill that we should have some kind of table with features, but that might be biting more than we can collectively chew given how fast things change and how much maintenance that would need. Maybe just a table for all of them with a "status" column, differentiating between live, in development, abandoned, unknown. Then we can think about other columns later, like governance, chain support, identity, etc.

> Thanks! Would including the 'abandoned/unknown' status be necessary for the table or would it be more relevant to just include the statuses that are currently in development/live? What do you think of a description column?

Swader:
A general one-line description would be fine as far as I'm concerned, but I'd lean towards more columns and maybe checkmarks for yes/no on feature, rather than descriptions - we want the page to be easier to translate.

docs/wallets.md
krichard410: Ah these two wallets were listed as "exploration" in their status, I now removed this since they weren't being developed :)

joepetrowski:
Thanks for making a PR! We already have a [list of wallets](https://wiki.polkadot.network/docs/en/build-tools-index#wallets), though, and resources about building them. This is certainly a more comprehensive list and probably does deserve its own page, but let's make sure to put it in the right place (`build/tools/wallets` IMO) and direct people to it from the current list so that we don't have conflicting lists.

laboon:
> Thanks for making a PR! We already have a [list of wallets](https://wiki.polkadot.network/docs/en/build-tools-index#wallets), though, and resources about building them. This is certainly a more comprehensive list and probably does deserve its own page, but let's make sure to put it in the right place (`build/tools/wallets` IMO) and direct people to it from the current list so that we don't have conflicting lists.

Joe - That list is out of date, and with the number of wallets appearing in the ecosystem, we wanted to add a new page for it.  

Docusaurus doesn't understand subdirectories very well, thus the "build-wallets" format I suggested for the filename.

We should definitely add a redirect from the tools that currently has wallets to this new page.  I want to have new pages for some of the other tools we have listed there, splitting them out from the one big tools page.

docs/build-tools-index.md
```diff
@@ -20,6 +20,8 @@ sections below will be filled out.
 
 ## Wallets
 
+See more on Wallets [here](build-wallets.md).
```
laboon:
```suggestion
See the [Wallets](build-wallets.md) page.
```

docs/build-tools-index.md
```diff
@@ -20,6 +20,8 @@ sections below will be filled out.
...
+
 - [Polkadot-JS Apps - Accounts](https://polkadot.js.org/apps/#/accounts)
```
laboon: Please remove all of these wallets so that there is only one place (the Wallets page) where there is a list of wallets (to prevent there being two sources of truth... just like normalizing a database)

docs/build-wallets.md
```diff
@@ -0,0 +1,30 @@
...
+title: Polkadot Wallets
+sidebar_label: Polkadot Wallets
+---
+
+## Polkadot Wallets
+
+The integration of a [wallet](https://wiki.polkadot.network/docs/en/glossary#wallet) and Polkadot
+allows for simple and easy access to private keys and signing transactions. Polkadot.js.org,
+Parity's Signer app, and the Polkadot{.js} plugin are just a few of Polkadots own wallets created.
+Below are wallets that support Polkadot along with their development statuses.
```
laboon:
```suggestion
The integration of a [wallet](https://wiki.polkadot.network/docs/en/glossary#wallet) with Polkadot
allows for simple and easy access to private keys and signing transactions. 
Below are some wallets that support Polkadot along with their development statuses.
Note that inclusion does not necessarily imply endorsement of that wallet.
```

website/sidebars.json
```diff
@@ -21,7 +21,8 @@
           "build-pdk",
           "build-cumulus",
           "build-smart-contracts",
-          "build-oracle"
+          "build-oracle",
+          "build-wallets"
         ]
       },
       {
```
laboon, line 28: Please also run "prettier" again after changes.

Mo has also reviewed and is OK with the basic structure.

laboon (changes requested):

laboon:
👍

PR #535, opened 2020-06-04
Update text code fence

### LINKED ISSUE #534: Plain text blocks are rendered as code blocks
Code blocks content are rendered as source code by default and receive syntax highlighting. When the content is plain-text it looks odd with many "keywords" rendered in bold:

image

I noticed this [here](https://wiki.polkadot.network/docs/en/learn-keys#appendix-a-on-the-security-of-curves) but there are likely other places.

Possibly the fix is as easy as using ` ```text` instead of ` ``` `?

Swader:
Closes #534


Swader:
Trivial. Merging.

PR #551, opened 2020-06-09
Add proxies

joepetrowski:
Add info about generalized proxies. Yes I made that diagram in Google Slides and took a screenshot. More artistic commits to this branch welcome.


docs/learn-proxies.md
```diff
@@ -0,0 +1,66 @@
...
+
+When a proxy account makes a `proxy` transaction, Polkadot filters the desired transaction to ensure
+that the proxy account has the appropriate permission to make that transaction on behalf of the cold
+account.
+
+As implied by the name, a proxy type of "Any" allows the proxy account to make any transaction,
+incuding balance transfers. In most cases, this should be avoided as the proxy account is used more
+frequently than the cold account and is therefore less secure. To allow any type of transaction
+except balance transfers (including vested transfers), the "Non-transfer" proxy type will be a more
+apposite fit.
```
Swader: Consider changing to "better fit" for simplicity.

docs/learn-proxies.md
```diff
@@ -0,0 +1,66 @@
...
+
+The "Governance" type will allow proxies to make transactions related to governance (i.e., from the
+Democracy, Council, Treasury, Technical Committee, and Elections pallets).
+
+The "Staking" type allows staking-related transactions, but do not confuse a staking proxy with the
```
Swader, line 35: I would clarify possible caveats here

docs/learn-proxies.md
```diff
@@ -0,0 +1,66 @@
...
+Controller account. Within the Staking pallet, some transactions must come from the Stash, while
+others must come from the Controller.
+
+### Anonymous Proxies
+
+Polkadot includes a function to create an anonymous proxy, an account that can only be accessed via
```
Swader, line 45: Consider explaining some possible use cases, it's not clear why one would need these other than something like multisig member replacements without rebuilding the multisig

Swader (approved):
Nits

docs/learn-proxies.md
joepetrowski: One of my favorite words, but you're right.

docs/learn-proxies.md
joepetrowski: Added a few lines

docs/learn-proxies.md
joepetrowski: TBH I'm a bit fuzzy on their applicability. They're not really anonymous since you still have to call from the account that created them. They are more like _exclusive_ proxies. Didn't come up with a use case but just tried to explain it better.

PR #555, opened 2020-06-10
Add kodadot and polkadot.js plugin to wallets

laboon:
Add kodadot and polkadot.js plugin to wallets


krichard410 (approved):
Looks good! Makes more sense to have [Polkadot-JS] as a Browser than a Browser Extension

PR #558, opened 2020-06-10
Num validators over 300

laboon:
Better than "rapidly approaching 200" and will last for a while longer, while still giving a number.


krichard410 (approved):
👍 👍

PR #584, opened 2020-06-17
Workflow to build and host Parachain Implementer's Guide

lamafab:
Hi

This PR contains:

* Github Actions workflow to automatically build the Parachain Implementer's Guide every hour (and only commit if something changed). The workflow runs for about 30 seconds. The output (live version) is saved in `website/static/implementers-guide/`.
* Link in the sidebar:
image
* A page in the `docs/` directory. I tried to link the live version directly, but Docusaurus [currently does not support it](https://github.com/facebook/docusaurus/issues/827).
image


docs/learn-implementers-guide.md
```diff
@@ -0,0 +1,12 @@
...
+id: learn-implementers-guide
+title: Parachain Implementer's Guide
+sidebar_label: Parachain Implementer's Guide
+---
+
+The Parachain Implementer's Guide is still heavily work in progress and maintained by Parity.
+
+## Live version
+
+[**The live version**](/implementers-guide/) is build from the source located in the official
```
Swader:
```suggestion
[**The live version**](/implementers-guide/) is built from the source located in the official
```

Swader (changes requested):
If a direct link in the sidebar would be better for UX, it's fairly easy to hack around with some JS which is already running custom stuff anyway.

lsaether (changes requested):
Would be better to place the link under the Build header in the sidebar (instead of the Learn section).

Is it best to place the built files in the repository? Maybe what we can do is just deploy every time a new version is built so we can avoid committing the built files to this repository?

lamafab:
@Swader 

> If a direct link in the sidebar would be better for UX, it's fairly easy to hack around with some JS which is already running custom stuff anyway.

If you can quickly do that, then why not. But maybe it's useful to keep it the way it is, since it also links to the source (there's not "Edit" button in the guide). Second opinions?

@lsaether 

> Would be better to place the link under the Build header in the sidebar (instead of the Learn section).

The Implementer's Guide is more for people who want to (re-)implement the Polkadot Node (Soramitsu, ChainSafe and probably for some Parity devs themselves), not necessarily for the devs who just want to **use** Polkadot. So "Learn" is the right place. Or maybe even link it in "Parachains" (in the "Resources" section, https://wiki.polkadot.network/docs/en/learn-parachains#resources)?


> Is it best to place the built files in the repository? Maybe what we can do is just deploy every time a new version is built so we can avoid committing the built files to this repository?

You're much more familiar with the design structure of the wiki, you tell me how this should be done :)

laboon:
> Would be better to place the link under the Build header in the sidebar (instead of the Learn section).

I think it might be better under Learn since the vast majority of developers on Polkadot won't need to ever deal with this directly, but it is technically about Building.

lsaether (approved):
Looks good. At some point in the future we should move the build action to correspond with deployment to avoid having to commit the static files directly to the source, but that's an enhancement that can come later.

lamafab:
> Looks good. At some point in the future we should move the build action to correspond with deployment to avoid having to commit the static files directly to the source, but that's an enhancement that can come later.

Once the Implementer's Guide is considered "stable" it will move in its own repository anyway (discussed with Rob). For now that's okay.

Swader (approved):

PR #585, opened 2020-06-18
Linking Teched Explainer videos

### LINKED ISSUE #570: Linking TechEd videos to Wiki
It might be helpful for new users (or visual learners) to watch a TechEd video after reading a page on the wiki.

krichard410:
Closes #570


lsaether (approved):
Nice! Eventually we can make it a goal to have a video accompanying every page.

Side note: We need to look into why our CI is not running here.

PR #605, opened 2020-06-24
Add explanation of secondary block difference

### LINKED ISSUE #597: Add note about difference in secondary block between Kusama and Polkadot
https://github.com/paritytech/substrate/blob/1efa919854649ba33f73059e62a72321306a17f2/client/consensus/babe/src/authorship.rs#L151

### LINKED ISSUE #598: Document difference in block signing between Kusama and Polkadot
Bruno | W3F
Basti said "Polkadot uses a different algorithm for secondary slots signing." - what does he mean?
Irt kusama vs polkadot
Bill | W3F
hmm, this is the first I've heard of it. where did he say it?
Bruno | W3F
Friday, 10:44am, discussion in Polkadot Validator Lounge on errors in 0.8.11. (btw catching up with stuff in Riot is SUCH a nightmare)
Bill | W3F
holy crap, yes, and trying to find something is like a double-nightmare
Logan
https://matrix.to/#/!NZrbtteFeqYKCUGQtr:matrix.parity.io/$159264274247753LUKac:matrix.parity.io?via=matrix.parity.io&via=matrix.org&via=web3.foundation
Bill | W3F
I'm not certain, but I think he might mean the algorithm for secondary slot selection? I remember seeing a PR for this
Logan
Wasn't aware of it either, but there is two secondary slots that can be claimed, a VRF slot and a Plain slot https://github.com/paritytech/substrate/blob/1efa919854649ba33f73059e62a72321306a17f2/client/consensus/babe/src/authorship.rs#L151
Bill | W3F
that's weird, I don't know why a VRF-selected block would be considered secondary.  I thought they were always primary.
Logan
I think it just adds the VRF output to Polkadot secondary blocks

lsaether:
closes #597 
closes #598

### DIFF
--- a/docs/learn-consensus.md
+++ b/docs/learn-consensus.md
@@ -105,6 +105,15 @@ either a _primary_ or a _secondary_ block, and no slots are ever skipped.
 For more details on BABE, please see the
 [working research draft](https://research.web3.foundation/en/latest/polkadot/BABE/Babe.html).
 
+#### Difference of BABE secondary blocks between Kusama and Polkadot
+
+Both Kusama and Polkadot uses the BABE block production mechanism outlined above. However, there is
+a slight difference in the secondary blocks that are produced between the two networks. Polkadot
+attaches the VRF output to secondary blocks (and therefore every block contributes to the Era
+randomness), while Kusama keeps the VRF output off the secondary blocks (meaning only the primary
+blocks contribute to Era randomness). The Polkadot method should give stronger randomness as more
+inputs are collected during every Era.
+
 ### GRANDPA: Finality gadget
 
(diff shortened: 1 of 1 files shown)


Swader (approved):
Sounds good, but can we explain *why* Kusama doesn't do this too?

wpank:
> Sounds good, but can we explain _why_ Kusama doesn't do this too?

I think the reason is that this wasn't really thought of for Kusama when when it came out. Polkadot just improves it.

wpank (approved):

Swader:
> I think the reason is that this wasn't really thought of for Kusama when when it came out. Polkadot just improves it.

What prevents us from putting this upgrade into Kusama too?

PR #611, opened 2020-06-26
Update README

lsaether:
Updates the README for the new commands and script locations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,19 @@ pull requests with these corrections. Larger contributions may start as issues t
 the subject with the maintainers. It is generally preferable to create a pull request over an issue
 to propose a change to the wiki content.
 
+### Running Locally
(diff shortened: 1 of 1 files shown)


lsaether:
Merging as insubstantial.

README.md
```diff
@@ -25,53 +38,56 @@ Use [Prettier](https://prettier.io/) to format markdown pages. To run Prettier o
 run:
 
 ```bash
-npx prettier --write ./docs/
+$ npx prettier --write ./docs/
```
Swader, line 41: Adding `$` makes these commands un-copy-pasteable. Usually, you would put `$` when the code fence has some output as well, so as to differentiate between commands and outputs, but when the command is meant to be run verbatim, the `$` is in the way.

PR #618, opened 2020-06-26
Fixed WSS setup

### LINKED ISSUE #482: Re-check WSS setup
There have been some complaints about [this tutorial](https://wiki.polkadot.network/docs/en/maintain-wss) no longer being up to date. Needs checking. This might affect people's setups with full nodes when PoA launches, and Kusama right now, so it's best handled asap.

Swader:
Closes #482


lsaether:
Can #559 be closed in favor of this?

Swader:
@lsaether yes

lsaether (approved):
Looks good. Would make sense to place it under the self-signed heading since it only applies to that option?

Swader:
The import only needs to be done *after* the nginx setup, which happens only after the cert generation, so it would be confusing before you have nginx imo.

PR #622, opened 2020-06-26
Fix #569

### LINKED ISSUE #569: Improve the Wiki Search Bar
When using the search bar on the wiki, there's always one search result that is concealed with pink. When hovered, the search result is visible, but it might be nice not having to hover to see a search result. 

Screen Shot 2020-06-15 at 8 13 18 AM

Screen Shot 2020-06-15 at 8 13 48 AM

(hovered over text):
Screen Shot 2020-06-15 at 8 19 39 AM

Swader:


krichard410 (commented):
Looks great! I think there is a `check links` error from CI though

Swader:
Sometimes a link burps temporarily. In such cases you can manually re-run the job in CI and it should pass. Checking...

PR #630, opened 2020-06-30
Link for checking inflation rate

### LINKED ISSUE #591: Make it easier to figure out inflation and interest rate
From the section like [this one](https://wiki.polkadot.network/docs/en/learn-staking#inflation) it is hard to figure out what the actual inflation and interest rate is. I twould be great to show the current staking rate that can be used for calculation or link people to some dashboard that shows it.

ansonla3:
#591

### DIFF
--- a/docs/learn-staking.md
+++ b/docs/learn-staking.md
@@ -342,7 +342,8 @@ For specific details about validator payouts, please see
 ## Inflation
 
 Inflation is designed to be 10% in the first year, with validator rewards being a function of amount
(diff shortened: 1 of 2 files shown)


lsaether (approved):

lsaether:
I think this only answers half of the question. You point the reader where to check the staking rate but you don't explain how they can determine the inflation rate from that. Maybe add a few more sentences about how to derive the inflation?

PR #635, opened 2020-06-30
Add section on multisig deposits

### LINKED ISSUE #568: Document deposit for multi-sig
There was a lot of discussion in Kusama Direction about deposits for using multi-sig accounts. This should be documented and explained in the proxy section.

lsaether:
closes #568

### DIFF
--- a/docs/learn-accounts.md
+++ b/docs/learn-accounts.md
@@ -230,8 +230,8 @@ should be used. This function takes only the other signatories and the raw call 
 However, in anything but the simple one approval case, you will likely need more than one of the
 signatories to approve the call before finally executing it. When you create a new call or approve a
 call as a multi-sig, you will need to place a small deposit. The deposit stays locked in the
-multi-sig until the call is executed. The reason for this deposit is to place an economic cost on
-the storage space that the multi-sig call takes up on the chain and discourage users from creating
+multi-sig until the call is executed. The reason for the deposit is to place an economic cost on the
+storage space that the multi-sig call takes up on the chain and discourage users from creating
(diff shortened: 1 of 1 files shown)


docs/learn-accounts.md
```diff
@@ -187,30 +187,93 @@ if the entities designated as the recipients come together in a new multi-sig un
...
+
+|               | Polkadot     | Kusama          |
+| ------------- | ------------ | --------------- |
+| DepositBase   | 200880000000 | 3347999999941.4 |
+| DepositFactor | 320000000    | 5333333312      |
+
+Let's consider an example of a multi-sig on Polkadot with a threshold of 2 and 3 signers: Alice,
+Bob, and Charlie. First Alice will create the call on chain by calling `as_multi` with the raw call.
+When doing this Alice will have to deposit 0.20152 DOTs while she waits for either Bob or Charlie to
+also approve the call. When Bob come to approve the call and execute the transaction, he will not
```
Swader:
```suggestion
also approve the call. When Bob comes to approve the call and execute the transaction, he will not
```

Swader (changes requested):
We should clarify where the deposit is coming from: should the multisig itself have the money to cover the deposit, or is it the initiating signatory?

Swader (commented):
If the script is the same as in your msig-utils script, something's off, it doesn't give the same address as the Polkadot JS Apps UI. Recommend fixing first.

lsaether:
msig-util has been fixed and the code example points to that repository now. The original example had a bug in how it sorted the addresses before hashing.

PR #640, opened 2020-07-01
Add guides page

### LINKED ISSUE #639: Index of How-to guide
We already have guides for most of the key functionality of Kusama and Polkadot, but they are scattered around different sections of the wiki. We should have an index of all of our how-to guides.

lsaether:
closes #639 

Some people were having trouble finding the different guides for interacting with the chain through Polkadot-JS Apps.


docs/getting-started.md
```diff
@@ -0,0 +1,30 @@
...
+- [Running a Validator][validator]
+- [Setting an Identity][identity]
+- [Creating a Proxy Account][proxy]
+- [Making Proposals and Voting for Referenda][democracy]
+- Social Recovery (Coming soon!)
+- [Running for the Council][council]
+- [Voting for Councillors][council voting]
+- [Using the Treasury][treasury]
+
+[account generation]: https://example.com
```
Swader, line 22: This should be `<li>`s as well, no?

Swader (changes requested):

docs/getting-started.md
joepetrowski: No, this is Logan's fancy markdown way of keeping the actual hyperlink out of the text :P

PR #643, opened 2020-07-02
Fix ribbon

ansonla3:


Swader (changes requested):
It's too flush with the edges, we need some padding, but the ribbon isn't as much of a problem as the menu is - if you open the menu on mobile and try scrolling, it's still broken. That should be fixed asap.

image

Swader:
Also keep in mind that these fixes will need to be applied to Kusama guide too

krichard410:
@ansonla3 a Wiki issue I created overlaps with this PR, did you want to add that into this PR? It's #674

Swader:
Sorry I'm going to close this in favor of a fix I'm preparing for #678 and #664 and #643

Issue #627, opened 2020-06-29
Moving Account generation and restoring accounts


krichard410:
I think it would make more sense if 'Account Generation' and 'Backing up/Restoring Accounts' were placed somewhere else than inside Miscellaneous, perhaps underneath Build? Somewhere easier for users to find in the side bar. 

<img width="286" alt="Screen Shot 2020-06-29 at 12 01 38 AM" src="https://user-images.githubusercontent.com/36748441/85968581-d7cfa300-b99b-11ea-9684-89a8afc1a0ec.png">

Swader:
I would go as far as advocating for another category besides learn, build and maintain. Something like "Getting Started" which could contain both extremely simple explanations and guides for basic things like these. @laboon @lsaether @ansonla3 thoughts?

lsaether:
"Getting Started" could also go under the Main heading (which currently doesn't have much visibility either). If we add another header button I would probably suggest we add "Main".

Swader:
Not sure how I feel about "Main". Indicates that something else is less important? Or is that just me? How about "General"?

PR #645, opened 2020-07-02
Switching order of navigation

### LINKED ISSUE #626: Switching Order of Build and Learn
I'm thinking switching the order of 'Build' and 'Learn' might make more sense to a new user, like first it's required you learn the material, then build with it, then maintain it? 

learn-polkadot

It might too be an interesting idea to have a 'Getting Started' tab underneath 'Learn' where it would suggest possible ways to get started on Polkadot (for example, is it best to start with making an account or watching a TechEd Explainer video on getting to know the UI or even going through the Web3 MOOC course?).

krichard410:
Closes #626 and #627

### DIFF
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -1,6 +1,6 @@
 {
   "docs": {
-    "Main": [
+    "General": [
       "index",
       "community",
       "contributing",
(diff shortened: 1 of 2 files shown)


Swader (approved):

Swader:
I like it. Way more intuitive imo.

Issue #66, opened 2019-10-07
Explain moving to/from parathread/parachain


laboon:
In learn-parathreads.md, we should go into more detail about moving back and forth between parachain and parathread.

https://github.com/w3f/polkadot-wiki/pull/52/files/7fb7bae49f4879f52fdb2d6b73c7dfc8b52e0e97#diff-1d51784cdb2416bf4d25a9ee46604911

laboon:
closed by #655

PR #655, opened 2020-07-03
Parachain->parathread and storage rent clarifications

### LINKED ISSUE #103: Storage rent and tombstones
There are two implementations of this.

1. In smart contracts, the system will function just like in Eth where you restore a contract for a price, from an archive node. This is documented in the substrate dev docs: https://substrate.dev/docs/en/ecosystem/contracts/ink#storage-rent\

2. There exists also the tombstoning concept of parathreads (and possibly chains?). A dead chain / thread will have its last block registered on the relay chain it's attached to. If all collators die, then that block becomes a tombstone. In such a case, all that's needed to restore this thread/chain to life if there ever appears a need to do this again is by paying the restore fee (explain? To whom? How?) and continuing to build on it as a revived collator. For this to happen, archive nodes are necessary (archive of what - the parachain / thread? Isn't that every collator basically?).

Further explanation necessary on what happens with hierarchical chains / threads. Consider a relay parachain with parathreads and child chains. So, say their blocks are being stored on the relay chain, and then one such sub-chain or sub-thread dies (no collators). It can be restored via the mechanics outlined above, that's ok. But what if our relay parachain dies - the one keeping the last block of this thread/chain? How do you restore that? You would first need to somehow restore the relay parachain to get the last block of its children, and then restore those? Again, to whom are the restore fees paid?

Further reading:

- https://github.com/paritytech/substrate/blob/master/srml/contracts/src/lib.rs
- https://github.com/paritytech/srml-contracts-waterfall/tree/master/contracts/rust/restore-contract

Swader:
This should close #103 and #66

### DIFF
--- a/docs/build-smart-contracts.md
+++ b/docs/build-smart-contracts.md
@@ -10,6 +10,14 @@ will support smart contracts. There are already announced projects such as
 [contract pallet](https://crates.parity.io/pallet_contracts/index.html), it is likely that more
 parachains will support this feature.
 
+Additionally, there is the EVM pallet which lets a parachain implement the Ethereum Virtual Machine,
+thereby supporting almost direct ports of Ethereum contracts. Some of the projects using this
+approach are [Moonbeam](https://moonbeam.network/) and
+[Frontier](https://github.com/paritytech/frontier).
+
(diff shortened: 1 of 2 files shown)


laboon:
looks good to me.

Issue #644, opened 2020-07-02
Edit button on mirror pages leads to 404


Swader:
Since mirrors are generated during the build process, the edit button does not work on those since there is no source file. We should customize the GH link for those to lead to the original pages.

PR #658, opened 2020-07-03
Fixed JS and CSS on Kusama Guide

### LINKED ISSUE #657: Custom CSS and JS broken on Kusama guide
The path `./js/...` and `./css/...` does not work when running the Kusama Guide. This makes Klaro (cookie consent) not work either, as well as custom analytics.

Swader:
Fixes #657 and #644


ansonla3:
Screenshot 2020-07-03 at 7 19 37 PM

Getting the above error @Swader

Issue #664, opened 2020-07-06
Fix the Kusama button


Swader:
The Kusama button looks "active" at all times. We should dampen that effect now that the wikis are separated again.

PR #671, opened 2020-07-07
Restored styles on Kusama button

laboon:
This is from Kirsten

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -423,15 +423,15 @@
       }
     },
     "links": {
-      "Build": "Build",
       "Learn": "Learn",
+      "Build": "Build",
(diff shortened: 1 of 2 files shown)


krichard410:
Does anyone know if 'fleek' needs to pass? I haven't seen that one before on a PR..

Swader:
I didn't know we were using fleek. The job creates and pins an IPFS hash for the content. I think @lsaether set this up, so he'll have to take a look when he's back because I don't have access to that fleek account. He should probably add us all so we can inspect it.

Swader:
Sorry I'm going to close this in favor of a fix I'm preparing for #678 and #664 and #643

PR #680, opened 2020-07-08
Fixes to a bunch of CSS and JS issues

Swader:
A lot of stuff, leftovers from the split plus some issues with rendering on mobile have been fixed. I also deleted some stray google tag manager files.


kusama-guide/static/css/custom.css
```diff
@@ -117,119 +117,29 @@ body {
...
-.github-fork-ribbon.right-bottom:after,
-.github-fork-ribbon.right-bottom:before {
-  -webkit-transform: rotate(-45deg);
-  -moz-transform: rotate(-45deg);
-  -ms-transform: rotate(-45deg);
-  -o-transform: rotate(-45deg);
-  transform: rotate(-45deg);
-}
-/*# sourceMappingURL=gh-fork-ribbon.min.css.map */
+  color: white !important;
```
krichard410, line 135: Should we also update the Kusama guide's Contributing ribbon too?

krichard410 (commented):
I just noticed that if the screen size is set to 320px wide, the nav links are a bit cut off (the english link and a bit of the search bar). Here's a photo:

Screen Shot 2020-07-08 at 5 07 16 PM

krichard410:
In the `index.html` file (inside static for the Polkadot wiki), I noticed that the loaded page directs to `http://localhost:3000/docs/en`, should it be this page or `http://localhost:3000/en` ?

krichard410 (approved):
This looks great! I'm a fan of the new Kusama button in the navigation :)

Swader:
Ah, you're testing on a much narrower screen than I. Let me fix those buttons.

kusama-guide/static/css/custom.css
Swader: It is updated

krichard410: Ah I see it now - i forgot to run `yarn kusama:start`. Looks all good!

Swader:
> In the `index.html` file (inside static for the Polkadot wiki), I noticed that the loaded page directs to `http://localhost:3000/docs/en`, should it be this page or `http://localhost:3000/en` ?

Good catch, fixed that too.

PR #683, opened 2020-07-09
Add section about unique transaction identifiers

joepetrowski:


Swader (approved):

Swader:
This is great, I think I'm going to make it into a tech ed explainer video.

joepetrowski:
> This is great, I think I'm going to make it into a tech ed explainer video.

The info in this guide is meant to be concise and stick to how someone should handle Polkadot, but there's a lot more that would be interesting in an explainer video, so that's a really good idea.

I was wondering if validators would have an incentive to stuff blocks with duplicate txs because, even if they were invalid and they didn't get the fee, the block's fullness would make the tx fees higher in future blocks. But I learned that if any of the pre-dispatch signed extensions fail, then not only is there no fee/dispatch, but the weight isn't even counted. So it's basically pointless for validators to put duplicate txs in a block unless they would both be valid.

PR #694, opened 2020-07-13
Add docusaurus patch for injection

### LINKED ISSUE #459: Looks like injector isn't running in CI
https://wiki.polkadot.network/docs/en/learn-faq#polkadot-launch

The placeholder `num_validators` is still there

lsaether:
closes #459 

The reason injection would not run when publishing using docusaurus is due to the script that was building and publishing in one go, with no intermediate step for us to run the injection. This adds a patch to the docusaurus script that will run the injection script during the publish process.


lsaether:
Tested on my [personal deployment](https://lsaether.github.io/test/docs/en/faq).

~~(Pretty sure this is right - but debugging why it doesn't replace the template on my personal deployment.)~~

PR #698, opened 2020-07-13
Adjusted the images on Account Generation

krichard410:
I noticed that the images on the Account generation page were very large and the mnemonic seeds were un-blurred, I've adjusted the images so users can read the content easier.


lsaether (approved):
Thanks!

PR #732, opened 2020-07-29
Remove duplicate iframe tag

spalladino:
The extra opening iframe tag was causing all content after the video not to be shown on the rendered page.


lsaether (approved):

lsaether:
Thank you!

PR #739, opened 2020-07-30
XCMP animation

krichard410:
UPDATE Sept. 29th: I can't upload the video to GitHub because it's far too large, but you can see my updated animated video here: https://drive.google.com/drive/folders/1CtAjE3sNELADCwHS8TRway8mhPMdB2cc?usp=sharing 

It's under the name '[FINAL] XCMP Explainer'

Feel free to leave comments/advice/feedback on this PR 

UPDATE PART TWO: Now you can run `yarn polkadot:start` on this pr and the video will show up on screen!


Swader:
Hmm, right now this animation doesn't tell me much. Are you sure you wouldn't rather use something like Adobe Animate? Super easy to use, lets you graphically depict exactly what you want to say. D3 is great but it's hard to use when you're doing something fully custom, and it is best suited for dynamic graphs that react to input or custom data rather than plain animation.

Swader:
I would say that the animation either needs to be sped up if the format it'll be presented in is GIF, because it's hard to pause and resume GIFs and rewinding would be a nightmare, or if it will be presented as a video, it would be good to have some narration reading the statements out loud.

krichard410:
> I would say that the animation either needs to be sped up if the format it'll be presented in is GIF, because it's hard to pause and resume GIFs and rewinding would be a nightmare, or if it will be presented as a video, it would be good to have some narration reading the statements out loud.

Hmmm I suppose a video could definitely work, but then it would be technically considered an animated video.

Swader:
Yes, correct. Is that a bad thing?

krichard410:
> Yes, correct. Is that a bad thing?

Oh no no not at all, I just need to then update the title of this PR so that it's accurate :)

ansonla3:
Maybe this would be a better way to illustrate the whole process, but it might involve too much work on the animation side?
Ref: https://animagraffs.com/how-cryptocurrency-works/

lsaether:
I think we need to be more specific on the wording in a few of the slides. Right now, I fear it might be possible to misinterpret the way that XCMP works. 

> Collators update all other collators on the Relay Chain with an cross-chain message.

When a new cross-chain message gets dispatched from a parachain, it's the collator's job to gossip this message among full nodes of the Relay Chain. Since collators are required to be nodes of the Relay Chain as well as their parachain it is expected that eventually the cross-chain message will be seen by the collator node on the receiving parachain. See the diagram below the green and oragne nodes act as network bridges across the two parachain networks since they are peered to nodes who are collators in both networks.

image

One important aspect that we haven't really mentioned yet on the wiki, is that XCMP will require a one-way channel to be open between parachain before any messages can be sent. So the first step is for Parachain A and parachain B to establish a channel between them, one is specified as sender and the other as a receiver.

> They are letting all collators know that they want to send a direct message.

Actually, they are not announcing the _intent_ to send a message but rather they are gossiping the message data itself. 

> Before the message is processed on the collator's next block it must be verified.

The collator will include the message with the proposed block so it's an added validation check that the validator will be doing. Right now the chronology of the animation makes me think that the validation check needs to happen before the collator produces the block, but it does not. If collators propose a block with an invalid XCMP message that block will not be validated by the validators, similarly if a collator does not include an XCMP message when there are messages in the queue will have their proposed block be invalidated. The collator should be doing the initial verification themselves before they propose their block to the validator, so that ideally the validator does not need to attempt to process an invalid message.

docs/learn-crosschain.md
```diff
@@ -56,6 +56,10 @@ The collator now hands this block to the validator, which itself will verify tha
 processed. If the message was processed and all other aspects of the block are valid, the validator
 will include this block for parachain B into the Relay Chain.
 
+Check out our interactive animation below that explores how XCMP works.
+
+<p><iframe src="../../website/static/published_Final_XCMP.html" width="100%" height="640"></iframe></p>
```
Swader:
```suggestion
<p><iframe src="/published_Final_XCMP.html" width="100%" height="640"></iframe></p>
```

Swader:
Do you need content feedback, i.e. has Logan's feedback been incorporated or not yet?

krichard410:
> Do you need content feedback, i.e. has Logan's feedback been incorporated or not yet?

Yes! This would be particularly helpful, and i've incorporated his feedback in. Any overall feedback would be nice too

Swader:
Looking through in order and just dropping comments as I see them:

1. I think the long paragraphs are on screen for too short a time. To optimize, consider reading them out loud when they're shown, and see if you can read them from beginning to end and add whatever time you think a viewer would need to both understand what they read and see it on the diagram. In fact, if we're already doing a movie, perhaps an accompanying voiceover would be better.

2. "Recall each parachain contains collators and validators". A parachain does not contain validators - a parachain is being validated by a subset of validators sampled from the active validator set, and the blocks being validated are being submitted by the collators.

3. Punctuation is missing in some statements and present in others. Please make it consistent.

4. "...with an cross-chain message" typo in "an"

5. "Collators update all other collators on the relay chain..." - see Logan's comment on how to clarify this better. It would also be good to have this visualized somehow.

6. On the "A cross-chain message is sent..." nothing has happened on the animation in a good while at that point and there's too much text (see 1). Move "containing a destination and a timestamp" to after "A cross-chain message". The second sentence doesn't make sense to me - why is it there? What's "this" in "this is used"? The message itself is used when you want to send a message? Redundant. Recommend removing.

7. "The cross-chain message sent from Parachain A has now been received from every collator on the relay chain". Relay Chain should be uppercase everywhere, it's how we capitalize it. **from** every collator is confusing - should it be from? Collators aren't **on** the RC so this sentence isn't very clear.

8. "Now the collators of Parachain B has received Parachain A's message. But before that message is processed on the collator's next block, it must be verified." - see Logan's comment. Also typo in first sentence: **has** -> **have**. Typo in second sentence **on** -> **in**.

9. "In order for a message to be verified..." paragraph is unclear to me. It does not have a logical flow - the first sentence seems redundant, stating the seemingly same thing through synonyms (to verify you, we'll validate you), given that the difference between validate and verify wasn't specified anywhere. "They will read the cross-chain message sent earlier to verify" - which message sent earlier? To verify what, that message from eariler, or the new one? How?

10. "Once the message..." paragraph is confusing. Consider rephrasing to "Once the validator-verified message has been received by all of Parachain B's collators, the message can be considered validated and verified." if that's what you meant.

11. By 00:47 the animation has stood still for a while, only changing text. Unless you have plans to animate more things, I do not believe this warrants a video.

12. Lastly, if we're going with a movie, consider making subtitles, uploading it to youtube, and including it with an embed. That way we can add subtitles from different langs, and the text does not get distorted / blurred when the video is resized onto a big screen. Subtitles are more accessible than in-video text.

Swader:
Hmm, I dunno, I'm not convinced.

Likes:

- playable video
- no text in video, way better with narration
- additional animation features

Dislikes:

- it's a bit hard to follow even for me. I'm not sure this would help a newbie understand things. Maybe a slower pace + more animation features like labels and arrows with names would help? Also, don't be afraid to pause the animation if you need to drive a point home, that's what I did in my [Transaction Uniqueness](https://www.youtube.com/watch?v=rn7GM8RZuew) animation.
- the audio has already begun when the video starts, so there's a cutoff at the beginning. I would also recommend investing in a pop filter.
- you say at one point that collators are full nodes of the relay chain. This is the first time I've heard of this, is this a recent change?

lsaether:
> you say at one point that collators are full nodes of the relay chain. This is the first time I've heard of this, is this a recent change?

I believe this is how they are currently implemented in Cumulus, when you run a parachain collator you have to run a full node of the Relay Chain. It's not a strict requirement for them to be full nodes though, they can also be light nodes. [This is the best documentation I can find on Cumulus](https://github.com/paritytech/cumulus/blob/master/docs/overview.md#node):

> A Parachain Collator will join the Parachain network and the relay-chain network... The Collator joins the relay-chain network for two reasons. First, the Collator uses it to send the Parachain blocks to the Parachain validators. Secondly, the Collator participates as light/full-node of the relay chain to be informed of new relay-chain blocks.

lsaether:
I agree with Bruno's points above regarding pacing and additional visual pointers. It would be good I think to pause the video now and then to unpack the statement that was just said/ visualized. I also wonder if there's a way to make the block proposal process a bit more explicit. This might mean switching to a second animation during the middle and middle-ending parts which shows the message as part of the PoV block being passed from collator to validator then "compressed" into a hash on the Relay Chain.

krichard410:
@Swader and @lsaether I've updated the video and is currently located inside https://drive.google.com/drive/folders/1CtAjE3sNELADCwHS8TRway8mhPMdB2cc?usp=sharing under the name 'Revamped_XCMP'. Could you guys give it a look?

Swader:
I'd say this is much better in all aspects. I would add visual cues to the video, specifically:

- whenever you say collator, highlight collators. Whenever you say parachain A's collators, highlight only parachain A's collators. That way the viewer always has context.
- in the beginning when you say "these are x, these are y" I would also add a label with an arrow pointing to the entity you're mentioning, so it's really clear, because nothing really explains that pink stuff will be the focus of your words. A label when you say "these" really drives it home
- when you include things onto the relay chain, one time you do it from the right side, one time from the left side (spiral arrow) but never in the place where it should go - it should hit the parachain's slot. As it is, it makes it look like the block is included somewhere far away from the chain that included it.
- when you say that collators send a block to validators, maybe it would be good to highlight collators when you say collators, illustrate a block that's sent to that chain's validators, and highlight validators, so the block proposal's path is clearly illustrated.

lsaether (approved):

Swader:
Newest video feedback:

- minor nitpick but collators send block candidates to the validators, not parachain blocks to the relay chain

I really like the new and improved labels and highlights, again an improvement in every way, well done! It's a bit unclear how the collators gossip messages amongst each other since in the diagram you're using (and I realize it's the only one we have), the collators are only a part of the parachain and are in no way connected, i.e. they do not overlap, so there is no reason to think that they'll see each other's messages (visually speaking), but I suppose that's a design problem.

And a note on sound - I think your computer's fan started spinning real hard in the tail end of the video, there's a noticeable loud hum in the background. If you get an external microphone, or record via some headphones that have a mic that's on the wire closer to you than to the computer, you should be fine. There is also a nice tool called Audacity which you can "train" for a specific hum, and then have it normalize the rest of the audio based on that training. So you would make sure your laptop is spinning hard, and then record "nothing" for about 20 seconds. Then record as usual. In the end during post-processing, in audacity you can just select the silence at the beginning, select to remove noise, and then copy that filter and apply it to the rest of the audio. But admittedly that's a final-stage upgrade you don't have to bother with now :)

ansonla3:
1. It would be better to highlight validators & collators in the video with text explicitly, but I am not quite sure whether it is difficult to do it using the video editing software like Adobe Premiere. 
2. Also, good to mention there is a "fallback delivery" https://medium.com/web3foundation/polkadots-messaging-scheme-b1ec560908b7. Did it change?

krichard410:
@Swader I've updated the video! It's inside the folder, called XCMP Project final.mov :) 
If you could take another look, that'd be awesome

Swader:
That part where you insert the soundbite is very jarring, but it's good otherwise I think. I would still clarify WHY the message reaches Parachain B's collators (i.e. all collators are full nodes of the relay chain and full nodes of the parachain, so they are in contact with each other) as that part isn't too clear, but other than that this has come a long way and I like how it improved, good job!

krichard410:
@Swader I ended up re-recording the audio (I do agree that tiny addition just jarred the entire thing) and I've included a note on why the message reaches Parachain B's collators. It's labelled "[FINAL] XCMP Explainer.mov" - could you give it another listen? 
I also used an embedded tool in my movie editor to remove hum noises, hopefully they can't be heard on your end :)

Swader:
Audio is great now, almost no hum and flows much better. I think this is publishable.

PR #743, opened 2020-07-31
Update learn-governance.md

rrtti:
from Kusama Council with 13 seats to 17 seats.


Swader:
I would sooner say 24 on both with a gradual increase over time, and specify the increase frequency.

PR #760, opened 2020-08-07
Updated various typos, wording, etc. 

### LINKED ISSUE #753: Update Sidebar and Links
I'm thinking of reducing the number of pages on the Polkadot wiki and have the sidebar a bit more organized. 

For example, in the Nominator section, there are three separate pages that all fall underneath the same category. We could blend this pages into one, call it 'Nominator Guide' and inside we have different subsections (that used to be the separate pages). 
Screen Shot 2020-08-04 at 4 40 00 PM

If we do something like this, we would have links to each of the subsections in the right sidebar so that users can click on a subsection without much scrolling. Example: 'What is a nominator?', 'How to stop being a Nominator", etc. 
Screen Shot 2020-08-04 at 4 41 06 PM

Of course though, depending on some of the material, having one page with multiple subsections could be a very long page, but I'm thinking we could do it for some of the smaller pages (like Nominator).

krichard410:
This PR contains:
- Rewording some content material 
- Renaming certain pages, IDs, and titles
- Renaming the Transactions page to Balance transfers

this PR isn't directly related to an issue, just things I stumbled upon in the wiki.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -4,9 +4,9 @@ title: Getting Started
 sidebar_label: Getting Started
 ---
 
-Welcome to the Polkadot Wiki! This central source of truth contains guides for interacting with the core functionality.
-Our wiki has three main sections: Learn (for learners), Build (for people that want to build on
-Polkadot), and Maintain (for people that want to maintain the network).
+Welcome to the Polkadot Wiki! This central source of truth contains guides for interacting with the
+core functionality. Our wiki has three main sections: Learn (for learners), Build (for people that
+want to build on Polkadot), and Maintain (for people that want to maintain the network).
 
 For brand-new learners of Blockchain technology:
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -36,7 +36,7 @@ For brand-new learners of Polkadot:
 - Reading [What is Polkadot? A Brief Introduction][article] on Medium. There are also other great
   articles to read on [Polkadot's Medium][p medium] or [Web3 Foundation's Medium][w medium].
 - [Creating a Polkadot Account][account generation]
-- [Sending a Balance transfer][transaction]
+- [Sending a Balance transfer][transfer]
 - [Staking as Nominator][nominator]
 - [Running a Validator][validator]
 - [Setting an Identity][identity]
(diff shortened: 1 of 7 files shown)


docs/getting-started.md
```diff
@@ -4,9 +4,9 @@ title: Getting Started
 sidebar_label: Getting Started
 ---
 
-Getting started on Polkadot is easy! The wiki already contains guides for doing most of the core
-functionality. Our wiki has three main sections: Learn (for learners), Build (for people that want
-to build on Polkadot), and Maintain (for people that want to maintain the network).
+Welcome to the Polkadot Wiki! This wiki contains guides for interacting with the core functionality.
+Our wiki has three main sections: Learn (for learners), Build (for people that want to build on
```
Swader: `wiki` said 3 times in 2 lines, maybe reduce.

Swader (approved):
Nitpick, otherwise looks okay

docs/learn-introduction.md
```diff
@@ -1,9 +1,31 @@
...
-sidebar_label: Polkadot
+title: What is Polkadot?
+sidebar_label: What is Polkadot?
 ---
 
+Polkadot enables scalability by allowing specialized blockchains to communicate with each other in a
+secure, trust-free environment.
+
+Polkadot is built to connect and secure unique blockchains, whether they be public, permission-less
+networks, private consortium chains, or oracles and other Web3 technologies. It enables an internet
```
lsaether, line 11:
```suggestion
networks, private consortium chains, oracle chains, or other Web3 technologies. It enables an internet
```

lsaether (approved):
The changes make sense to me.

docs/learn-introduction.md
krichard410: Just wondering but why should this be changed to 'or' instead of 'and'? =)

docs/getting-started.md
krichard410: Good point there, actually!

krichard410: Does "this central source of truth" sound better or a bit worse? :)

Swader: Haha, better I think!

PR #776, opened 2020-08-12
Removes roadmap

lsaether:
The PR removes the roadmap page from the wiki. It hasn't been maintained since launched.


krichard410 (approved):
I think this is a good idea :)

PR #793, opened 2020-08-17
Remove Social Recovery TODO

lsaether:
We have a [social recovery guide](https://guide.kusama.network/docs/en/kusama-social-recovery#docsNav) but it's a Kusama-only feature so we can remove the line from the Polkadot guides page.


Swader:
Any official word on why it's not on Polkadot? Seems kinda useful

Issue #63, opened 2019-10-04
PDF Downloads


lsaether:
We should enable PDF download so that people can have an offline way to read the content.

Swader:
I looked into this, and the only package currently capable of doing this is docusaurus-pdf but it [does not support custom styling](https://github.com/KohheePeace/docusaurus-pdf/issues/8).

Another possible approach is [pandoc](https://pandoc.org/demos.html). Here is an example implementation via Travis I found: [1](https://github.com/sourabh2311/MuftEducation.in/blob/master/.travis.yml), [2](https://github.com/sourabh2311/MuftEducation.in/tree/master/travis), [3](https://github.com/sourabh2311/MuftEducation.in/blob/master/requirements.txt) but I don't think this would respect styling.

I think we better just generate from HTML, experimenting with that next: https://github.com/alextiley/convert-html-to-pdf

lsaether:
Let me know how it goes. I don't think we really need too much styling though, as long as we can have proper headings and italics. Maybe even using a markdown -> pdf converter would work but would require us to integrate it into the build process.

PR #796, opened 2020-08-17
Pdf export

Swader:
Solves #63 

There are two new commands: `kusama:pdf` and `polkadot:pdf`. They generate `kusama-guide/static/kusama-wiki.pdf` and `website/static/polkadot-wiki.pdf` respectively. These commands take a while to execute and need to run against a locally running wiki, so they're not part of any other command just now. They feel a bit naked without a ToC front page, so that's up next along with the other stuff below.

Please test.

To do:

- [x] Add link to download on pages
- [ ] Build ToC front page
- [ ] Add export process to circle build


lsaether:
I would recommend adding a GitHub action ([example](https://github.com/w3f/polkadot-wiki/blob/master/.github/workflows/check-substrate-release.yml)) to export the PDF on push to master, then it can publish it to the releases assets.

lsaether:
Why not use a [markdown to pdf parser](https://md2pdf.netlify.app/)?

That would come up with cleaner end renders.

krichard410 (commented):
This is neat! I'm excited to see the ToC page. 
Just wondering though if the 'accept cookies' bar could be removed when the pdf has downloaded, as it covers up part of the content on the page. I'm also wondering if it's possible to remove the header and footer navigation if the page continues onto the next. The navigation on each page covers up some of the images/material that is there. 
It might also be good too if blank pages were removed, as they aren't really necessary.
For the download button that will be on each page, will there also be an option to download the whole site? or is it page by page?

Swader:
> @lsaether: I would recommend adding a GitHub action (example) to export the PDF on push to master, then it can publish it to the releases assets.

Sounds good, was planned, but need to iron out the output first.

> @lsaether: That would come up with cleaner end renders.

Cleaner in which way?

> @krichard410 'accept cookies' ... header and footer navigation

None of those should be showing - the print CSS has been modified to remove them. Let me check what's up...

Download would be full book only, did not consider page by page (seems pointless).

Swader:
Looks like I forgot to `git add` the main CSS file. Done now, please re-test for cleaner output.

krichard410:
Just re-tested it, it looks great. Could we also get the page numbers in there too?

krichard410:
One thing I noticed were these unlinked images on the Kusama Governance page. Could these be fixed?

Screen Shot 2020-08-18 at 12 04 51 PM

krichard410:
Is there a way to remove embedded videos from the pdf format (and replace them with the link)? I saw them throughout both pdfs, but it seems a bit pointless to have them in there since they don't actually play.

Swader:
Yes, that'll need tweaks. It's a matter of modifying the CSS for print style so that it hides an embed and shows a video, same for the formula images above I assume. I'll try and get that polished. Thanks for testing!

Swader:
Ok, so the entire flow works now, but the problem with Github Actions is that one workflow cannot `needs` another workflow. So I can't build the PDF in one workflow and the use it the deploy workflow - I'll need to modify the two deploys. Working on that now. Hopefully they add dependency actions soon.

Swader:
I have no way to check if this passes now. @lsaether any ideas? This triggers on commits to master only and *should* work, but not sure how to test. Tried [act](https://github.com/nektos/act) but it doesn't work without netcat, and I'm not sure how to make sure the container that runs the workflow gets it. Experimenting...

Edit: [ah, drat](https://stackoverflow.com/questions/57982945/how-to-apt-get-install-in-a-github-action).

Swader:
Ok, I recommend trying to merge this and seeing what happens, then reverting what doesn't work. All in favor @lsaether @ansonla3?

Swader:
Unless there are any objections I'm merging this tonight and seeing what blows up.

PR #804, opened 2020-08-18
Fix #docsnav container height for smaller width displays

igorline:
In browser windows with width between 700px and 1024px there has been always an annoying container of doc navigation covering the main content. This pr fixes it.

pic-selected-200818-2137-41


Swader:
Thanks!

PR #810, opened 2020-08-21
Add research links

laboon:
Seems like we should add some links to our papers / articles on the research.md page.

My worry is that the links might go stale, but a) I feel like they let people know what to expect before heading off the wiki to the research page and b) they've stayed pretty constant for a few months already and c) CI will yell at us if they go stale anyways.


krichard410 (approved):
this looks great to me

PR #817, opened 2020-08-24
Redenomination now in past

laboon:
Made some redenomination updates


Swader:
Should we say that the DOT is referred to as New DOT, or just DOT? I think the latter would be better. We want people to default to _just_ DOT, no?

Swader (approved):

laboon:
No, we are supposed to say _New DOT_ (in italics) vs. "DOT (Old)" (no italics)

At some point we will start referring to _New DOT_ as just DOT, once enough time has passed post-redenomination

PR #848, opened 2020-08-30
Marking Cobo Wallet as Live + Adding Safepal

SuragSheth:
Cobo Wallet is Live for both Polkadot & Kusama. 
Safepal Hardware Wallet is Live for both Polkadot & Kusama.


ansonla3:
same as the previous one ;)

PR #858, opened 2020-08-31
Removes the Kusama sidebar

### LINKED ISSUE #852: Kusama guide still seems inside the Polkadot wiki?
Opening an issue for discussion, but I noticed that if you use the search bar on the Polkadot Wiki to search up something from Kusama and click on it, the Kusama Guide is inside the website. 

For example, if I search up "address book" in the search bar and click on the first result:
Screen Shot 2020-08-31 at 2 35 32 PM

Then the Polkadot Wiki now looks like this:
Screen Shot 2020-08-31 at 2 36 12 PM

I don't think the Polkadot Wiki sidebar should be replaced with the Kusama Guide sidebar. This seems like a possible bug - I'd love to hear some feedback on this.

lsaether:
closes #852 

This removes the Kusama sidebar when a Kusama page is visited - but I think we might still run into the issue of Kusama pages getting indexed by the Algolia search engine. A possible extension to this would be to add a blanket redirect for any URL on the wiki that contains "kusama" to be redirected to the Kusama guide. That might be overkill though since we have the Kusama button clearly visible, and users can understand that's where the Kusama information lives (right?).

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -433,8 +433,7 @@
       "General": "General",
       "Learn": "Learn",
       "Build": "Build",
-      "Maintain": "Maintain",
-      "Kusama": "Kusama"
+      "Maintain": "Maintain"
     }
   },
   "pages-strings": {
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -154,37 +154,5 @@
         ]
       }
     ]
-  },
-  "kusama": {
-    "Kusama": [
-      "kusama-index",
-      {
-        "type": "subcategory",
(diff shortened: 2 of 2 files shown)


krichard410:
I think the redirect would be overkill and could make users annoyed if we had that implemented. For now, I think it's best to just delete the Kusama sidebar so that users can still navigate thru the wiki. Perhaps we could even add a "notice" at the top of every Kusama page on the wiki that says it's best to head over to the guide.

krichard410 (approved):
This works well!

PR #865, opened 2020-09-01
Update DOT decimal

krichard410:
will
Dot has 10 decimals now right?
joe
indeed
https://wiki.polkadot.network/docs/en/maintain-guides-how-to-nominate-polkadot

**from Polkadot Wiki channel on Element**


laboon:
👍

PR #900, opened 2020-09-04
Update staking info

### LINKED ISSUE #899: Add information to staking on oversubscribed / active / inactive / waiting nominations
Q: What does active / inactive / waiting mean?

A: When you nominate a set of validators (up to 16), the validator election algorithm will rotate your stake among validators on a per-era basis. In other words, you will have a different _active validator_ amongst your selected validators each era.  You cannot choose which specific validator you will be actively nominating in a given era, unless you are nominating a single validator.

Active: This is the validator you are actively nominating this era.  This validator will pay rewards to you for this era once the payout has been activated.
Inactive: These validators are in the active validator set, and you have selected to nominate them.  However, you are not actively nominating this era and they will not give rewards to you for this era.
Waiting: These validators did not get into the active validator set this era.  They may get into the active validator set in the future.

Note that it does not hurt your rewards to nominate waiting validators, as long as you have nominated at least one validator in the active set.  Your stake will be distributed only among those in the active validator set. 

Q: What is an oversubscribed validator?

A: Validators can only pay out to 64 nominators per era.  If more than 64 nominators nominate the same validator, it is "oversubscribed", and only the top 64 staked nominators (ranked by amount of stake) are paid rewards.  Other nominators will receive no rewards for that era.

If you are subscribed to an oversubscribed validator and are not staking a very large amount (in the top 64 for that validator), you should stop nominating the oversubscribed validator, and you should look for other validators which are not oversubscribed and nominate them.

In polkadot.js UI, you will be able to see the number of subscribed nominators in the current era.  Although it is difficult to determine exactly how many nominators will nominate a given validator in the next era, one can estimate based on the current number of nominators.  A nominator with only 5 nominators in this era, for instance, is unlikely to have more than 64 in the next era.  An already-oversubscribed validator with 100 nominators this era, however, is much more likely to be oversubscribed in the next era as well.

laboon:
1. Added terms to glossary
2. Updated learn-staking to mention oversubscribing
3. Added info to nominator guide about oversubscribing
4. Fixed descriptions of active vs inactive and waiting in nominator guide

Fixes https://github.com/w3f/polkadot-wiki/issues/899

### DIFF
--- a/docs/learn-staking.md
+++ b/docs/learn-staking.md
@@ -97,9 +97,9 @@ will get paid more).
 The following example should clarify the above. For simplicity, we have the following assumptions:
 
 - These validators do not have a stake of their own.
-- They do NOT charge any commission fees
-- Reward amount is 100 DOT tokens
-- The least amount of DOT to be a validator is 350
+- They do NOT charge any commission fees.
+- Reward amount is 100 DOT tokens.
+- The current minimum amount of DOT to be a validator is 350.
 
 |               | **A - Validator Pool** |                             |         |
 | :-----------: | :--------------------: | :-------------------------: | :-----: |
@@ -122,6 +122,15 @@ _Both validator pools A & B have 4 nominators with the total stake 600 and 400 r
 Based on the above rewards distribution, nominators in validator pool B get more rewards per DOT
 than those in pool A because pool A has more overall stake. Sam has staked 50 DOT in pool A, but he
 only gets 8.3 in return, whereas Kitty gets 12.5 with the same amount of stake.
+
+There is an additional factor to consider in terms of rewards. While there is no limit to the number
+of nominators a validator may have, a validator does have a limit to how many nominators to which it
+can pay rewards. In Polkadot and Kusama, this limit is currently 64, although this can be modified
+via runtime upgrade. A validator with more than 64 nominators is _oversubscribed_. When payouts
+occur, only the top 64 nominators as measured by amount of stake allocated to that validator will
+receive rewards. All other nominators are essentially "wasting" their stake - they used their
+nomination to elect that validator to the active stake, but receive no rewards in exchange for doing
+so.
 
 We also remark that when the network slashes a validator slot for a misbehavior (e.g. validator
 offline, equivocation, etc.) the slashed amount is a fixed percentage (and NOT a fixed amount of
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -4,6 +4,12 @@ title: Glossary
 sidebar_label: Glossary
 ---
 
+## Active Nomination
+
+A validator (or validators) that a nominator is actively validating this era. The nominator is
+placing their stake behind this validator for this era and will potentially receive staking rewards
+in return for doing so.
+
 ## Alexander
 
 The fourth (now defunct) proof of concept (PoC-4) testnet for Polkadot.
@@ -178,6 +184,11 @@ for how it stores all messages in the Relay Chain storage, therefore making it m
 demanding more resources than XCMP. The plan is to retire HRMP once the implementation of XCMP is
 complete.
 
+## Inactive Nomination
+
+A validator (or validators) that a nominator has selected to nominate, but is not actively
+validating this era. This type of nomination may become active in a future era.
+
 ## Inherent
 
 Extrinsics that are "inherently true." Inherents are not gossiped on the network and are put into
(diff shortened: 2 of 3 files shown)


docs/learn-staking.md
```diff
@@ -99,7 +99,7 @@ The following example should clarify the above. For simplicity, we have the foll
 - These validators do not have a stake of their own.
 - They do NOT charge any commission fees
```
krichard410:
```suggestion
- They do NOT charge any commission fees.
```

docs/learn-staking.md
```diff
@@ -99,7 +99,7 @@ The following example should clarify the above. For simplicity, we have the foll
...
 - Reward amount is 100 DOT tokens
```
krichard410:
```suggestion
- Reward amount is 100 DOT tokens.
```

docs/learn-staking.md
```diff
@@ -99,7 +99,7 @@ The following example should clarify the above. For simplicity, we have the foll
...
-- The least amount of DOT to be a validator is 350
+- The current minimum amount of DOT to be a validator is 350
```
krichard410:
```suggestion
- The current minimum amount of DOT to be a validator is 350.
```
Just adding in periods since the first point has one above =)

docs/learn-staking.md
```diff
@@ -123,6 +123,14 @@ Based on the above rewards distribution, nominators in validator pool B get more
 than those in pool A because pool A has more overall stake. Sam has staked 50 DOT in pool A, but he
 only gets 8.3 in return, whereas Kitty gets 12.5 with the same amount of stake.
 
+There is an additional factor to consider in terms of rewards. While there is no limit to the number
+of nominators a validator may have, a validator does have a limit to how many nominators to which it
+can pay rewards. In Polkadot and Kusama, this limit is currently 64. A validator with more than 64
+nominators is _oversubscribed_. When payouts occur, only the top 64 nominators as measured by amount
+of stake allocated to that validator will receive rewards. All other nominators are essentially
+"wasting" their stake - they used their nomination to elect that validator to the active stake, but
+receive no rewards in exchange for doing so.
```
krichard410: In this case, what would be the best option for nominators to do? And how what does it look like on the UI if a validator doesn't select the nominator for payout?

laboon: It mentions what to do to avoid this in the nomination guide, this is meant as an overview.  I'm hesitant to describe _too_ specifically what it looks like in the UI here since not everyone will be using the Polkadot UI.

docs/learn-staking.md
```diff
@@ -123,6 +123,14 @@ Based on the above rewards distribution, nominators in validator pool B get more
 than those in pool A because pool A has more overall stake. Sam has staked 50 DOT in pool A, but he
 only gets 8.3 in return, whereas Kitty gets 12.5 with the same amount of stake.
 
+There is an additional factor to consider in terms of rewards. While there is no limit to the number
+of nominators a validator may have, a validator does have a limit to how many nominators to which it
+can pay rewards. In Polkadot and Kusama, this limit is currently 64. A validator with more than 64
```
wpank: I would additionally add that this can be changed via governance and may be a higher value in the future.

docs/learn-staking.md
krichard410: Gotcha! That does make sense

docs/learn-staking.md
laboon: good point

laboon: It's specified in the runtime, so while technically the upgrade happens via on-chain governance, I think specifying that it happens via runtime upgrade makes more sense.

laboon:
@krichard410 can you re-review please?

docs/learn-staking.md
```diff
@@ -123,6 +123,15 @@ Based on the above rewards distribution, nominators in validator pool B get more
 than those in pool A because pool A has more overall stake. Sam has staked 50 DOT in pool A, but he
 only gets 8.3 in return, whereas Kitty gets 12.5 with the same amount of stake.
 
+There is an additional factor to consider in terms of rewards. While there is no limit to the number
+of nominators a validator may have, a validator does have a limit to how many nominators to which it
+can pay rewards. In Polkadot and Kusama, this limit is currently 64, although this can be modified
```
krichard410, line 128: Is the limit going to change anytime soon?

krichard410 (approved):
This looks great to me, I could find anymore issues 👍

docs/learn-staking.md
laboon: Probably.  There's a PR to update it to 128 on Kusama and 256 on Polkadot.  But I don't want to mention this yet.

PR #930, opened 2020-09-09
Updated ledger guides

Swader:


docs/kusama-ledger.md
```diff
@@ -80,27 +78,31 @@ At the end of the process you should have the newly installed Kusama application
...
-show "ledger".
+Depending on your browser and its security settings, you might need to confirm the USB connection
+through a popup like the one below:
 
-![](assets/ledger-3.png)
+![Display of the Ledger account in the Accounts list of PolkadotJS](/img/ledger/query-device.png)
 
-You can now use this account to interact with Kusama on Polkadot-JS Apps and it will prompt your
+You should be able to scroll down and find a new account on the page named "Ledger". Under the type
+column, the account will also show "ledger".
```
krichard410, line 101: Here is it "Ledger" or "ledger" ?

docs/kusama-ledger.md
```diff
@@ -145,17 +149,32 @@ method is to use [Polkadot-JS Apps][].
 - Confirm the transaction on your device.
 - A green success notification will be displayed when the transaction is included in a block.
 
+> Note the "Transfer with Keep-Alive Checks" toggle. While this toggle is in the _On_ state, your
+> account will be unable to make transactions which would get its balance below the existential
+> deposit. This prevents reaping of accounts with low balances. If you toggle this to _Off_, you
+> will be able to go below ED balance, causing your account to be deleted and any dust amount of KSM
```
krichard410: Here would it be worth it to say Existential Deposit instead of ED?

lsaether (changes requested):
We should put images that go into docs in the `docs/assets` folder so that we don't duplicate them in both `static` folders. This is part of what my image PR does to optimize the amount of files hosted in the wiki repository. The `static` files can be used for anything that is not documentation images or specific to one of the individual websites.

Otherwise the text looks good.

Swader:
On it.

lsaether (approved):

PR #949, opened 2020-09-14
Updated Auctions page

### LINKED ISSUE #918: Update parachain auction info
https://wiki.polkadot.network/docs/en/learn-auction
it talks about how candle auctions work but never actually says that the tokens are reserved!
the only talk of "bond" is in the code sample:

Bids[
  ({
    range: [1, 2, 3, 4],
    bond_amount: 300, //DOT
  },
  {
    range: [1, 2],
    bond_amount: 777, //DOT
  },
  {
    range: [2, 3, 4],
    bond_amount: 450, // DOT
  })
];

Logan
It does say that the tokens are "locked up" in the preceding paragraph:

Bidders will submit a configuration of bids specifying the DOT amount they are willing to lock up and for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.

joe
eh ok, but we should have some parachain basics

krichard410:
Closes #918

### DIFF
--- a/docs/learn-auction.md
+++ b/docs/learn-auction.md
@@ -105,8 +105,11 @@ into its own `lease period`. More than one continuous `period` is a `range`.
 Several auctions will take place in the preceding six months before a set of parachain slot leases
 begin.
 
-Bidders will submit a configuration of bids specifying the DOT amount they are willing to lock up
-and for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
+Bidders will submit a configuration of bids specifying the DOT amount they are willing to bond and
+for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
+
+> Please note: If you bond tokens with a parachain slot, you cannot stake with those tokens. In this
+> case, you don't pay DOT for a parachain slot.
 
 A bidder configuration for a single bidder may look like this:
 
--- a/docs/learn-auction.md
+++ b/docs/learn-auction.md
@@ -109,8 +109,7 @@ Bidders will submit a configuration of bids specifying the DOT amount they are w
 for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
 
 > Please note: If you bond tokens with a parachain slot, you cannot stake with those tokens. In this
-> way, you pay for the parachain slot by forfeiting the profit opportunity from staking the bonded
-> tokens.
+> way, you pay for the parachain slot by forfeiting the opportunity to earn staking rewards.
 
 A bidder configuration for a single bidder may look like this:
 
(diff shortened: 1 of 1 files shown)


docs/learn-auction.md
```diff
@@ -105,8 +105,11 @@ into its own `lease period`. More than one continuous `period` is a `range`.
 Several auctions will take place in the preceding six months before a set of parachain slot leases
 begin.
 
-Bidders will submit a configuration of bids specifying the DOT amount they are willing to lock up
-and for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
+Bidders will submit a configuration of bids specifying the DOT amount they are willing to bond and
+for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
+
+> Please note: If you bond tokens with a parachain slot, you cannot stake with those tokens. In this
+> case, you don't pay DOT for a parachain slot.
```
Swader: I don't understand this - "In this case you don't pay DOT for a parachain slot"? Do you mean "In this way, you pay for the parachain slot by forfeiting the profit opportunity from staking the bonded tokens."

Swader (changes requested):

docs/learn-auction.md
```diff
@@ -105,8 +105,12 @@ into its own `lease period`. More than one continuous `period` is a `range`.
...
 begin.
 
-Bidders will submit a configuration of bids specifying the DOT amount they are willing to lock up
-and for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
+Bidders will submit a configuration of bids specifying the DOT amount they are willing to bond and
+for which ranges. The slot ranges may be any continuous range of the periods 1 - 4.
+
+> Please note: If you bond tokens with a parachain slot, you cannot stake with those tokens. In this
+> way, you pay for the parachain slot by forfeiting the profit opportunity from staking the bonded
+> tokens.
```
joepetrowski:
```suggestion
> way, you pay for the parachain slot by forfeiting the opportunity to earn staking rewards.
```

joepetrowski (approved):
I think this change is good as in the page is definitely better with it. But to close #918 I'd like to see even more introductory material. As in, before even the "Mechanics of a Candle Auction" section, something should explain what a parachain slot is (what is being auctioned) and this information regarding tokens being reserved and returned later (what do parachains bid _with_).

As the page is now, diving right into candle auction mechanics without any context makes the page less useful.

laboon:
Yeah, we need to add even more information on parachains, but let's just get this info for now.  I've added an issue here - https://github.com/w3f/polkadot-wiki/issues/970 - for more general information.

PR #958, opened 2020-09-15
Adding IOS and Android for Safepal

SuragSheth:
Adding IOS and Android for Safepal


krichard410 (approved):
👍  i've also just tested this out today

PR #963, opened 2020-09-15
Added ledger info due to recent changes

### LINKED ISSUE #959: Polkadot.js ledger changes?
ronnie_rodriguez
Yesterday
luka
quick question about ledger and polkadot.js. following this guide but I can't find the setting option for attaching Ledger devices. did I miss something? or it's not support yet?
It appears there may have been a change to the JS interface. "Query Ledger" is now replaced with "Add Ledger", which defaults to create new accounts. Is there a different reason "Query Ledger" wouldn't show?
ronnie_rodriguez
In reply to luka
ronnie_rodriguez
It appears there may have been a change to the JS interface. "Query Ledger" is now replaced with "Add Ledger", which defaults to create new accounts. Is there a different reason "Query Ledger" wouldn't show?
Correction...Add Ledger appears to now query the ledger to show existing accounts, and not create a new account. Maybe just a change in the menu text.

Swader:
Fixed #959


laboon (commented):
I can't run through the step myself (Ledger coming soon) but text looks OK.

Do you think it makes sense to add a temporary message at the top letting people know about the version mismatch?

Swader:
It might. I also need to add a runtime version clarification so I'll push that in a few mins.

PR #974, opened 2020-09-16
Adds an introduction to auctions page

### LINKED ISSUE #968: Add information on selling parachain slots midway through
From Gav today in Polkadot Watercooler:

DOT/KSM is bonded for the amount of time agreed when the slot was leased. For DOT, this can be 6, 12, 18 or 24 months. This cannot be freed (except by governance intervention).

However, parachain slot holders can "sell" their slot (and its associated deposit) to some other would-be parachain. So if they find a team willing to pay them at least the deposit which they put down for the slot, then they can "unlock".

### LINKED ISSUE #970: Parachain auction info
We need to add more information on the basics of the parachain auction (what is a parachain slot?  why do you want to bid on it? etc.) before diving into the mechanics of the candle auction.

https://wiki.polkadot.network/docs/en/learn-auction

lsaether:
closes #970 
closes #968

### DIFF
--- a/docs/learn-auction.md
+++ b/docs/learn-auction.md
@@ -4,6 +4,13 @@ title: Parachain Slots Auction
 sidebar_label: Parachain Slots Auction
 ---
 
+For a [parachain](learn-parachains) to be added to Polkadot it must inhabit one of the available
+_parachain slots_. A parachain slot is a scarce resource on Polkadot and only a limited amount will
+be available. As parachains ramp up there may only be a few slots that are unlocked every few
+months. The goal is to eventually have 100 parachain slots available on Polkadot (these will be
+split between parachains and the [parathread pool](learn-parathreads)). If a parachain wants to have
+guarenteed block inclusion at every Relay Chain block, it must acquire a parachain slot.
+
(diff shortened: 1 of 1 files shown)


docs/learn-auction.md
```diff
@@ -81,6 +88,10 @@ slot for any contiguous range of the slot duration. Parachains may lease more th
 time, meaning that they could extend their lease to Polkadot past the 2 year slot duration simply by
 leasing a contiguous slot.
 
+> Note: Individual parachain slots are not unique with each other, this means that parachains do not
+> need to always inhabit the same slot, but as long as it inhabits any slot it can continue as a
+> parachain.
```
Swader:
```suggestion
> Note: Individual parachain slots are fungible. This means that parachains do not
> need to always inhabit the same slot, but as long as a parachain inhabits any slot 
> it can continue as a parachain.
```

Swader (changes requested):
Nitpick

docs/learn-auction.md
lsaether: Thanks, that's a better way to phrase it.

Swader (approved):

PR #975, opened 2020-09-16
Fix NL link

Swader:


Swader:
Why didn't the link checker detect this before?

lsaether:
Not sure - site returns a 404.

Swader:
Does the link checker run on kusama or just on polkadot wiki?

PR #978, opened 2020-09-17
Add second bonding release info to maintain-democracy page

laboon:
This seems directly relevant to "seconding a proposal", so I am adding some info to make it more explicit here.


krichard410 (approved):
Great addition 👍

PR #981, opened 2020-09-18
Ledger update

Swader:
The firmware is different for the two devices.


krichard410 (approved):
good thing to add in!

PR #984, opened 2020-09-19
Correct the Existential Deposit amount

alias0x:
Following DOT redenomination, the Existential Deposit amount is now 1 DOT instead of the previous 0.01 DOT.

### DIFF
--- a/docs/maintain-guides-how-to-nominate-polkadot.md
+++ b/docs/maintain-guides-how-to-nominate-polkadot.md
@@ -77,7 +77,7 @@ left over to pay transaction fees. Transaction fees are currently around 0.01 DO
 dynamic based on a variety of factors including the load of recent blocks.
 
 Also be mindful of the reaping threshold - the amount that must remain in an account lest it be
-burned. That amount is 0.01 DOT on Polkadot, so it's recommended to keep around 0.1 DOT in your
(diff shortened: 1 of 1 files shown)


docs/maintain-guides-how-to-nominate-polkadot.md
```diff
@@ -77,7 +77,7 @@ left over to pay transaction fees. Transaction fees are currently around 0.01 DO
 dynamic based on a variety of factors including the load of recent blocks.
 
 Also be mindful of the reaping threshold - the amount that must remain in an account lest it be
-burned. That amount is 0.01 DOT on Polkadot, so it's recommended to keep around 0.1 DOT in your
+burned. That amount is 1 DOT on Polkadot, so it's recommended to keep around 10 DOT in your
```
Swader:
```suggestion
burned. That amount is 1 DOT on Polkadot, so it's recommended to keep at least 1.5 DOT in your
```

Swader (changes requested):
Reduced recommended amount due to only existential deposit being high - this does not mean that other actions are more expensive too. 0.5 DOT should be plenty for many transactions of different types before it gets you below 1 DOT.

Swader:
Thanks!

PR #995, opened 2020-09-21
Removed uneeded files from build

### LINKED ISSUE #936: Old Kusama Claims page in Polkadot Wiki
Accessing Kusama claims via the website goes to the updated page with screenshots https://guide.kusama.network/docs/en/kusama-claims
When googling "Kusama claims" this page also comes up: same text but no screenshots (and not accessible via the website as far as I can see.) https://wiki.polkadot.network/docs/en/kusama-claims 
Merge/delete?

### LINKED ISSUE #938: Algolia un-index Kusama
@lsaether since you're the one in charge of the Algolia account for the wiki, can you tell them to unindex the *kusama* URL pattern? Otherwise we get stuff like this in search, which leads to a sidebar-less page.

image

lsaether:
closes #938 
closes #936 

This will remove built files that are not located in the `sidebar.json` from the build so that we no longer have hidden pages. Also updates node version in CircleCI and updates the docusaurus patch.


Swader:
We've got quite an application going by now...

PR #1027, opened 2020-09-27
Clarified image

### LINKED ISSUE #1014: Ledger Set up: Replace screenshot of Account Type & Index
Can the screenshot that currently shows "Account type 1 and Index 2" please be replaced with the default settings of 0 and 0 in order not to mislead users. 
It would also be good to point out that regular users should just leave the default settings. They should only be changed if people know what they are doing. 

https://wiki.polkadot.network/docs/en/learn-ledger#loading-your-account

Swader:
Fixes #1014


Swader:
Merging, easy and necessary hotfix.

PR #1031, opened 2020-09-27
Closes #999, closes #1017

### LINKED ISSUE #999: Mention Ledger in Polkadot Account and Account Generation
We how have a new option to recommend, given that it's officially added to Polkadot UI. We should make a note, however, that Ledger is not without its risks. Notably, [restoring from a Ledger seed phrase won't work when importing into any Substrate seed phrase tool](https://github.com/polkadot-js/apps/issues/3703) because the derivation is so different. This means an out of date Ledger app prevents people from interacting with the chain - it's a direct dependency on Ledger, the company, unless people trust the dev version of the app and install it manually.

The following pages need a Ledger note:

- Polkadot Accounts
- Account Generation

### LINKED ISSUE #1017: Account page is stale
The learn-accounts page needs a refresh. The indices part stands out as outdated but all of it could use a review.

Swader:
- Closes #999
- Closes #1017


krichard410 (approved):
👍 this looks great!

PR #1050, opened 2020-09-29
Adding repo secret token to Action

Swader:
The stale check action failed because I forgot to add the token into the action's YML. This should fix it. The token is automatically taken from the repo's environment, and generated by Github.


Swader:
Merging as hotfix.

PR #1053, opened 2020-09-30
Update KSM amount for deposit on the explanation about required deposit

rrtti:
Update KSM amount for deposit on the explanation about required deposit, from 100ksm to 3.3.ksm.


TomaszWaszczyk:
Is it the price of sending proposal?

PR #1115, opened 2020-10-10
Adding 5 wallets, updating 1

SuragSheth:
Adding Dether, Spatium, Swipe, Blockchain.com, Crypto.com, Flipping Cobo to Live.


Swader:
Have you tried using all of these? I would also like to add a new column indicating "custodial", so people can tell that risk level at a glance.

lsaether (approved):
I'm going to merge under the assumption that @SuragSheth has relationships with these teams and/or has tried these wallets for himself.

PR #1159, opened 2020-10-15
Fixes the checking that stale issues already exist

### LINKED ISSUE #1138: Do not post duplicate stale page issues if one is already open
Currently the stale page issue bot will post a duplicate issue for a page if it has not been updated yet and there's already an issue open. Instead, it should check if there's an open issue and ping the assignee to update it.

See for example #1071 and #1127 or #1069 and #1123 for indication of duplicated issues.

lsaether:
closes #1138 

The `pull_request` field doesn't exist on the issue object that is returned. Even on issues that do have pull requests open, I checked for the one that I have open.

### DIFF
--- a/scripts/stalecheck.js
+++ b/scripts/stalecheck.js
@@ -1,12 +1,22 @@
 const fs = require("fs");
 const cp = require("child_process");
-const core = require("@actions/core");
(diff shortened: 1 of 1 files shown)


ansonla3:
looks good

PR #1208, opened 2020-10-24
Broken hashquark link - replacment?

nuke-web3:
Removed, as it goes to a 404.

Possibly replaced by this tool by them instead? https://polkacube.hashquark.io/#/polkadot

### DIFF
--- a/docs/learn-validator.md
+++ b/docs/learn-validator.md
@@ -68,8 +68,6 @@ exchange for their activities.
 
 ## Validator Stats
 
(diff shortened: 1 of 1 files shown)


lsaether:
Yep it looks like this is the same tool that was referenced before https://polkacube.hashquark.io/#/polkadot/strategy

@NukeManDan can you swap out the link instead of removing it?

nuke-web3:
Done. 😁 
I added a note that this is for more than just nominator/validator estimated rewards. LMK if you this it should be seporately referenced / changed.

lsaether (approved):
Looks good

PR #1228, opened 2020-10-29
Mention the rejection of web verification

lamafab:
The registrar does not support "legal_name" or "web".


Swader:
Would it not make more sense to mark as OK instead? Seems like an easy trap to fall into

PR #1308, opened 2020-11-19
Change word "candidate" to "voter" in Phragmen algorithm

jaybutera:
It says "load of the candidate", but candidates don't have loads. It's also clear from the example below that this should be "load of the voter".


Swader:
Thanks!

PR #1330, opened 2020-11-25
Fixes stalecheck

lsaether:
Otctokit returns paginated responses with a max of 100 responses per page. At the moment we are getting a total of 207 responses. So if a page was already marked sufficiently long in the past it would get duplicated. Now it won't.


Swader:
Sweet! Let's see in the next round (in an hour or so).

PR #1341, opened 2020-11-27
Syntax / clarification fix :pencil2:

sacha-l:
Clarifying piece on using Phragmén in electoral system.


lsaether (approved):
Thanks! This version is more clear.

PR #1369, opened 2020-12-02
define shard (wikipedia)

nuke-web3:
Not sure if this is the _best_ link, but this is not an intuitive concept to new folks. Most have not heard of this outside web3 and database nerds 😉


lsaether (approved):
Thanks!

PR #1407, opened 2020-12-08
Fix broken link to Kusama parameters

CurlyBracketEffect:


laboon:
Thanks, looks like this used to work but not once we separated the wikis!

PR #1410, opened 2020-12-09
Update oversubscribed numbers

laboon:
Updated the nominator oversubscription limits and changed it to use an injected variable.

This value can't (AFAIK) be directly queried on-chain, so I have just set it to default.  We'll still have to update it but at least it's all in one place.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -264,9 +264,9 @@ biasing, or identity-based quadratic voting.
 
 ## Oversubscribed
 
-If more than 256 nominators nominate the same validator, it is "oversubscribed", and only the top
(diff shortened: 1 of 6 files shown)


laboon:
I'm not sure how to test the injection locally, does anyone else know?

ansonla3:
> I'm not sure how to test the injection locally, does anyone else know?

steps:
1. yarn polkadot:build
2. yarn polkadot:inject 

after executing the above commands, open the page (path: website/build/polkadot-wiki/docs/) that contains the injection vars and you should see the var get replaced by the on-chain value.

laboon:
Looks good if people are ready to review

lsaether (approved):

PR #1413, opened 2020-12-09
Add instructions for removing old democracy locks

### LINKED ISSUE #1349: Add command for unlocking old democracy votes
A few people must manually clear democracy votes (especially if they are very old).  We should add a step for this in `maintain-guides-democracy#unlocking-locked-tokens`.

How to unlock lockup
1) Enter https://polkadot.js.org/apps/#/extrinsics.
   -How to set
     * submit the following extrinsic: democracy / removeVote(index)
     * index: ReferendumIndex: (Enter the number you voted for) 10

   -Check if the voting record is cleared

2) Enter https://polkadot.js.org/apps/#/extrinsics.
   -How to set
     * submit the following extrinsic: democracy / unlock(target)
     * target: AccountID (Select account to unlock)

3) Enter https://polkadot.js.org/apps/#/accounts.
   -As a result of the verification, you can see that the democracy lock and locked are unlocked.

// Why not before

It is confirmed that the vote for democracy was not deleted from the beginning, and was not released because the unlock command was issued.

laboon:
fixes #1349


docs/maintain-guides-democracy.md
```diff
@@ -158,6 +158,31 @@ when successful.
 
 ![democracy clear locks](assets/democracy_clear_locks.png)
 
+If you don not see an option to clear expired democracy votes, it may be that the lock is very old.
```
lsaether:
```suggestion
If you do not see an option to clear expired democracy votes, it may be that the lock is very old.
```

docs/maintain-guides-democracy.md
```diff
@@ -158,6 +158,31 @@ when successful.
...
+You then must clear it by directly issuing the correct extrinsics.
```
lsaether:
```suggestion
You then must clear the lock by directly issuing the correct extrinsics.
```

docs/maintain-guides-democracy.md
```diff
@@ -158,6 +158,31 @@ when successful.
...
+Now submit the following extrinsic: `democracy.unlock(target)`, where target is your your account
+address.
+
+![democracy clear_lock_extrinsic_2](assets/democracy_clear_lock_extrinsic_2.png)
+
+If you return to the [Accounts page](https://polkadot.js.org/apps/#/accounts), you should see that
+the democracy lock has been released.
+
+Note that this applies only to locked DOT that were used for voting on referenda. In order to unlock
+DOT locked by voting for Polkadot Council members, you need to go to the
```
lsaether: What's the justification for capitalizing "Council"?

laboon: I have always capitalized it.  Since there is only one, it seems like a proper noun.  Is this against our style guide?

laboon: E.g., even though "court" is not a proper noun, I would write "the Supreme Court".

lsaether: Ok, that makes sense. I don't think it is in the style guide that why I'm asking. Maybe we can rephrase to "members of the Polkadot Council" since I think that sits better?

docs/maintain-guides-democracy.md
```diff
@@ -158,6 +158,31 @@ when successful.
...
```
laboon:
```suggestion
DOT locked by voting for members of the Polkadot Council, you need to go to the
```

laboon:
Let me know if this looks good @lsaether

lsaether (approved):

PR #1426, opened 2020-12-10
Add better information on high availability

### LINKED ISSUE #1411: Emphasize to validators that they should never run keys in multiple places
Already we advise validators that high availability set-ups that depend on session key duplication is dangerous. Now with the parachain validity slashing being 100%, once parachains are live this would mean that session key duplication leads to a full slash. We should __strongly emphasize__ that validators never run duplicated keys since being offline will be much tamer than the parachain invalidity slash.

lsaether:
closes #1411


laboon (approved):
LGTM

PR #1460, opened 2020-12-16
Update some locking wording and time periods

laboon:
Also the locking periods were incorrect; they are 28 on Polkadot and 7 on Kusama

### DIFF
--- a/docs/learn-governance.md
+++ b/docs/learn-governance.md
@@ -211,14 +211,15 @@ period.
 #### Voluntary Locking
 
 Polkadot utilizes an idea called `Voluntary Locking` that allows token holders to increase their
-voting power by declaring how long they are willing to lock-up their tokens, hence, the
-number of votes for each token holder will be calculated by the following formula:
+voting power by declaring how long they are willing to lock-up their tokens, hence, the number of
+votes for each token holder will be calculated by the following formula:
 
 ```
 votes = tokens * conviction_multiplier
 ```
 
-The conviction multiplier increases the vote multiplier by one every time the number of lock periods double.
+The conviction multiplier increases the vote multiplier by one every time the number of lock periods
+double.
 
 | Lock Periods | Vote Multiplier |
 | :----------: | :-------------: |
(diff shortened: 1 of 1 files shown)


docs/learn-governance.md
```diff
@@ -230,8 +231,9 @@ The conviction multiplier increases the vote multiplier by one every time the lo
 |      16      |        5        |
 |      32      |        6        |
 
-The maximum number of lock periods is set to 6 and the lock period is 30 days on Polkadot and eight
-days on Kusama.
+The maximum number of "doublings" of the lock period is set to 6 (and thus 32 lock periods in
+total), and the lock period is 28 days on Polkadot and 8 days on Kusama. Only doublings are allowed;
```
jonasW3F: Do I understand correctly that one period is always 28 days in Polkadot? That means when I lock for 4 Periods, I have my tokens locked for 112 days? If so, I'd propose to change
"and the lock period is 28 days" to "and one lock period is 28 days..."

jonasW3F: And I still don't fully get the idea of that doubling. If I want to vote for something is it not that I say:

 "I have 100 DOTs and I want to vote for proposal XY and lock it for 8 periods." 

Then I sign the transaction and basically vote with 400 tokens, but they are locked for 8 * 28 days?

Or is it that I lock my tokens, vote and with increasing time I don't unlock them, they gain more weight? But then at what point are the votes counted? And what happens for things to vote on, which are in the immediate future. I just can't utilize on certain multipliers? And does that mean I can vote with my "heavy" tokens for several different proposals?

jonasW3F (commented):
I still have some questions, which are included as comments.

docs/learn-governance.md
laboon: 1. Right, one period is always 28 days. I'll update to say "one"

2. Your vote is "worth" 400 DOT and they count immediately.  Your 100 DOT will be locked for 8 * 28 days after the vote is over.  You can't unlock these beforehand; the democracy lock will exist until after 8 * 28 days is over.

3. Yes, you can still vote with your "heavy" tokens.

laboon: I'll write some stuff in tomorrow to clarify these.

jonasW3F: ah okay great, then it's actually easier than I thought and could be clarified :) thanks

laboon:
Can you take another look @jonasW3F ?

docs/learn-governance.md
```diff
@@ -232,8 +232,14 @@ double.
 |      32      |        6        |
 
 The maximum number of "doublings" of the lock period is set to 6 (and thus 32 lock periods in
-total), and the lock period is 28 days on Polkadot and 8 days on Kusama. Only doublings are allowed;
-you cannot lock for, say, 24 periods and increase your conviction by 5.5, for instance.
+total), and one lock period equals 28 days on Polkadot and 8 days on Kusama. Only doublings are
+allowed; you cannot lock for, say, 24 periods and increase your conviction by 5.5, for instance.
+
+While a token is locked, you can still use it for voting and staking; you are only prohibited from
+transferring these DOT to another account.
```
jonasW3F: tokens to keep it general.

jonasW3F (changes requested):
I'd change the one reference of DOT to "tokens" to keep it general. Apart from this, it's good and now better understandable.

docs/learn-governance.md
laboon:
```suggestion
transferring these tokens to another account.
```

PR #1508, opened 2020-12-30
Add some information on parachain hubs

lsaether:


docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
 Parachains are not required to have their own token. If they do, is up to the parachain to make the
 economic case for their token, not Polkadot.
 
+## Parachain Hubs
+
+While Polkadot enables crosschain functionality amongst the parachains, it necessitates that there
+is some latency between the dispatch of a message from one parachain until the desitnation parachain
```
laboon:
```suggestion
is some latency between the dispatch of a message from one parachain until the destination parachain
```

docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
...
+recieves the message. In the optimistic scenario, the latency for this message should be at least 2
```
laboon:
```suggestion
receives the message. In the optimistic scenario, the latency for this message should be at least 2
```

docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
...
+blocks, one block for the message to be dispatched and one block for the receiving parachain to
```
laboon:
```suggestion
blocks - one block for the message to be dispatched and one block for the receiving parachain to
```

docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
...
+process and produce a block that acts upon the message. However, in some cases we may see that the
+latency for messages are higher if there is a lot of messages that are in queue to be processed, or
```
laboon:
```suggestion
latency for messages is higher if there are many messages that are in queue to be processed, or
```

laboon (changes requested):
This looks good to me other than some typos.

docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
...
+if there exists no nodes that are running both of the parachain networks that can quickly gossip the
```
Swader:
```suggestion
if there exist no nodes that are running both of the parachain networks that can quickly gossip the
```

docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
...
+message across the networks.
```
Swader, line 63: As far as I understand it, all collators will have to be full nodes of the relay chain, so they will by default all be in contact with each other, no?

docs/learn-parachains.md
```diff
@@ -53,6 +53,30 @@ could not include that in their implementation and Polkadot would still enforce
...
+
+Due to the neccesary latency involved in sending crosschain messages, some parachains are planning
+to become _hubs_ for an entire industry. For example, a parachain project
+[Acala](https://acala.network) is planning to become a hub for decentralized finance (DeFi)
```
Swader, line 67: How does this work, concretely? It's assumed there would still be DeFi products on other chains, like Moonbeam. How does Acala help with this? Or is the idea behind a hub simply that they plan to entice all DeFi projects to just build on them?

Swader (commented):
Just a couple of questions

docs/learn-parachains.md
lsaether: That's a good point - they all should be gossiping relay chain messages to each other but I don't know if this includes the parachain messages or if there is a way for nodes to identify which parachain collator nodes other full nodes of the relay chain are also running.

docs/learn-parachains.md
lsaether: The idea of the hub is primarily that it makes composability between applications easier. That is, both Acala and Moonbeam will have smooth interoperability between applications on their own parachain. While applications on Moonbeam would experience latency when interacting with anything that's deployed to Acala and vice versa.

laboon:
Is this ready for re-review @lsaether ?  If so can you resolve conflicts?

docs/learn-parachains.md
Swader: Did you find out?

docs/learn-parachains.md
Swader: So literally "they plan to do what eth is doing now and concentrate all apps and protocols on them", so it's a battle for dev attention between the chains?

Swader (changes requested):

laboon:
I think that even though there's not a lot of concrete detail and more can be added, we can merge this as-is.

What do you think @Swader ?

PR #1523, opened 2021-01-04
Update TokenPocket wallet status building -> live

chendatony31:


Swader:
There's no need to update translated docs, those are auto generated

Issue #1379, opened 2020-12-03
learn-treasury page: add Kusama spend_period variable


laboon:
`spend_period` was used for both Kusama and Polkadot.  We should have separate variables.  Current kusama spend period is 6.

I updated it to 6 manually as a stopgap measure.

laboon:
Actually, we should probably just not use a Kusama-specific variable here.  We probably shouldn't even discuss Kusama spend periods on the _Polkadot_ wiki.

PR #1533, opened 2021-01-05
Remove kusama reference from polkadot wiki

laboon:
Instead of making an ad hoc `kusama_spend_period` variable, just removing Kusama ref here.  See  https://github.com/w3f/polkadot-wiki/issues/1379#issuecomment-754734871

Let me know if you disagree, otherwise this is a simple PR.


Swader:
Good temp fix, yeah

PR #1534, opened 2021-01-05
Clarified councillor bonds 

### LINKED ISSUE #1295: Clarify councilor bonds and slashes
> In the elections-phragmen pallet, each candidate has a life cycle with 3 possible stages, namely Candidate, Member or Runner-up. First, accounts can submit_candidacy(), and the pallet will reserve CandidacyBond. If the account makes it into the Members or Runners-up list, then this bond is kept. Else, a loser candidate will always lose their bond (slash_reserved()). Moreover, members and runners-up must also renounce_candidacy() in time. If they lose their spot in an election round (i.e. no longer hold any role), they also lose their bond as the consequence.

midegdugarova:
fixes #1295


Swader:
Thanks!

PR #1546, opened 2021-01-06
Updated the alertmanager.service  example

ilhanu:
The example was the full flag, localhost and port, this set-up gives me a working alertmanager service


ilhanu:
Closed it to check if everything was working with the alert-manager.

lsaether (approved):

PR #1592, opened 2021-01-15
corrected internal link

nuke-web3:
made an error in syntax. Double checked to work now!


Swader:
That link was actually leading to the link from the bottom of the text, but it was included in the text wrong (should have been `[thousand validators]` (no hyphen). Can you please also remove that link from the bottom of the page, so it's not there orphaned?

nuke-web3:
good catch. Updated all external links using "`[text](link)`" syntax to avoid this thing in the future.

nuke-web3:
@Swader sorry to bother you on it, I am not sure why the CI check here fails, the error is not helpful lol

AFAIK these is no reference to substrate releases in this PR

Swader:
It's fine, plug-in error :) no worries

Swader:
Generally the convention is up to the author (I know @lsaether prefers to keep links at the bottom, I'm more of an inline guy), but it is indeed important to make sure the references and links match when doing so, thanks for the fix.

PR #1597, opened 2021-01-16
Update maintain-guides-how-to-validate-kusama.md

TheoMuffin:
Docker command featured an old version of the image and did not specify the flag needed to run on the kusama chain as opposed to the default polkadot chain.


Swader:
Thanks!

PR #1604, opened 2021-01-19
add datahighway proposal

sheenhx:
thanks!


Swader:
What is this? There are still only 5 parachains on Rococo afaict.

laboon:
Actually there are only four at the moment - Mandala has a slot but is not producing blocks. https://polkadot.js.org/apps/#/parachains

PR #1618, opened 2021-01-23
Removed unsupported `--ws-cors` parameter.

legacycode:
--ws-cors parameter does not exists and leads to an error when starting polkadot.


Swader:
Thanks!

PR #1625, opened 2021-01-25
Emphasize tokens locked for proposals/seconding

laboon:
Emphasize that locking tokens for proposals -> indeterminate amount of time


ghost:
I think there needs to be clear language that "proposals cannot be revoked, even if they never turn into a referendum".

lsaether (approved):

PR #1630, opened 2021-01-26
Fixes to Kusama treasury page (fixes #1603)

### LINKED ISSUE #1603: Dynamic value placeholders have not been replaced on Kusama treasury page
See [here](https://guide.kusama.network/docs/en/mirror-learn-treasury).
For example, on the first paragraph:

> ...with current defaults set to {{ spend_period }} days for Polkadot

I've had a quick look and it seems to be the only page where this is the case.

alxs:
(should fix) #1603

See other commits as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,12 @@ kusama-wiki.zip
 
 # autogenerates at each run / build
 docs/build-ss58-registry.md
+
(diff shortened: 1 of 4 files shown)


Swader:
Nope, didn't work :(

Issue #1629, opened 2021-01-25
Change CI to Github Actions


Swader:
Consolidating into Github Actions might be a good idea - one less service to keep around / maintain / dabble with permissions on.

salmad3:
Has been addressed.

PR #1635, opened 2021-01-26
Attempt to fix #1603

### LINKED ISSUE #1603: Dynamic value placeholders have not been replaced on Kusama treasury page

alxs:
#1603

Works locally for me. Not sure if this should be run before `node ./postprocess;`.


Swader:
@lsaether do we need this postprocess thing to be separate? Can it be a more integrated part of the build process? Or maybe we can merge/remove some other things, or make them scheduled github actions instead? It seems like we're invoking an awful lot of commands during start / build, and the CI is starting to look very messing to the point of being hard to follow. Let's think about normalizing? (e.g. #1629)

lsaether:
So the reason this is working locally but failing when deploying is because there are two different pipelines for building a static site and deploying to GitHub pages. The `yarn polkadot:build` and `yarn kusama:build` variants are used to build the static site but we use the `yarn polkadot:publish-gh-pages` and `yarn kusama:publish-gh-pages` which itself calls the Docusaurus script `docusaurus-publish`. 

The core of this issue now comes in that `docusaurus-publish` itself is calling multiple commands sequentially and in order to make injections we need to do it after the pages are built but before they are deployed (which happens inside the `docusuaurus-publish` script). So [the solution I used was to patch the docusaurus script](https://github.com/w3f/polkadot-wiki/blob/master/patches/docusaurus%2B1.14.6.patch) and put the inject script inside of it.

But I don't know why this isn't working anymore...

lsaether:
> @lsaether do we need this postprocess thing to be separate? Can it be a more integrated part of the build process? Or maybe we can merge/remove some other things, or make them scheduled github actions instead? It seems like we're invoking an awful lot of commands during start / build, and the CI is starting to look very messing to the point of being hard to follow. Let's think about normalizing? (e.g. #1629)

Yeah I think auditing the build process and seeing if we can streamline anything would be good to do. I'm not sure if using scheduled actions would benefit us greatly since the processes we're running usually are to change, add, or remove things during the build pipeline.

Swader:
That's true, yeah. Maybe inject could be scheduled though, to keep itself up to date without us needing to push 🤔 Right now we're updating several times per day so that's not an issue, but maybe there will one day be a time when multiple days (or even weeks, whaaaaat?!) will go by without an update. Anyway, streamlining is next on the agenda then definitely. Can you by any change look into the publish patch you wrote please so the placeholders in mirrors get replaced, and then we can look at the CI and streamline things collectively?

lsaether:
The issue was that the scripts were using the wrong directory to find the files. [This is the commit that fixed it](https://github.com/w3f/polkadot-wiki/commit/da71fae03a1f06a6463612bdc2a68456a8979ad0).

Swader:
Nice! 🎉

PR #1649, opened 2021-01-29
1064

whitesunset:
Hello there, 

Two separate actions are published here (check README for available options): 
* https://github.com/w3f/action-create-issues-for-old-files
* https://github.com/w3f/action-find-old-files

Updated workflow with those two actions are tested on my fork, they're working as described in this issue. Please check and share a feedback.

And thanks Bruno for help, for sure!

### DIFF
--- a/.github/workflows/stalecheck.yml
+++ b/.github/workflows/stalecheck.yml
@@ -25,5 +25,6 @@ jobs:
       - name: Create Issues
         uses: w3f/action-create-issues-for-old-files@master
         with:
+          randAssignees: ["swader", "lsaether", "ansonla3", "laboon"]
           files: ${{ steps.find.outputs.files }}
 
--- a/.github/workflows/stalecheck.yml
+++ b/.github/workflows/stalecheck.yml
@@ -8,11 +8,22 @@ jobs:
   check-stale:
     name: Checks for stale files
     runs-on: ubuntu-latest
+    env: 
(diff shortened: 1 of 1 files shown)


.github/workflows/stalecheck.yml
```diff
@@ -8,11 +8,22 @@ jobs:
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Find Docs
+        id:   find
+        uses: w3f/action-find-old-files@master
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Stale docs
+          path: ${{ steps.find.outputs.files }}
+      - name: Create Issues
+        uses: w3f/action-create-issues-for-old-files@master
```
Swader, line 26: This should define the tech ed team as the current action does for random assignee selection

Swader (changes requested):

.github/workflows/stalecheck.yml
whitesunset: Hello! Do you mean randAssignees list should be passed to second action?

Like this?
```
     - name: Create Issues
       uses: w3f/action-create-issues-for-old-files@master
       with:
         randAssignees: ["swader", "lsaether", "ansonla3", "laboon"]
```

Swader: Yep!

PR #1662, opened 2021-02-03
Update build-wallets.md

laboon:


Swader:
But Parity Signer has no staking, it's just a signer, why would we say it has staking?

laboon:
Well, if that's the case, the extension also doesn't allow staking. But people were confused because they thought you couldn't stake if you used PS.  What about changing it to "Allows indirect staking" or "Staking available via Polkadot-JS" or something?

Swader:
I'd add columns:

- Native staking: Y/N
- Custom endpoints: Y/N

Then later add more like:

- Governance support: Y/N
- Treasury support: Y/N

etc...

laboon:
Canceling this for a while, OBE anyways

PR #1669, opened 2021-02-03
Fix a logical error in Learn Phragmén 

wischli:
This PR replaces a single word in the the "Learn Phragmén" wiki entry: `validator` to `nominator`. I believe this an overseen logical error. 

As the text states that
* `validator` == `candidate`
* `nominator` == `voter`, 

it does not make sense to me that 
> "as candidates are elected, a weighted mapping is build, defining the weights of each selection of a `[candidate]` by each candidate". 

I suppose you meant to say `nominator` or `voter`.


laboon:
@wischli You are correct that the sentence is incorrectly worded, but it's actually the opposite change that should be made - "the selection of a candidate by each nominator", as nominators select validators and not the other way around.

Fixed in https://github.com/w3f/polkadot-wiki/pull/1674

Thank you for the report!

PR #1681, opened 2021-02-07
Added timeline diagram

saki-osive:
Added a diagram that explains the launch phases diagrammatically.


Swader:
Thanks but this doesn't seem necessary and it takes up a lot of screen real estate that we'd rather conserve. It also has some extra files in the commit from your IDE. Closing, we'll have our designer make a timeline if we decide to include one.

PR #1684, opened 2021-02-08
Fix staking mention and naming

Tbaut:
- Polkadot-js apps and extension are both part of the Polkadot-js repo. To ease search engine indexing, I suggest using Polkadot-js (the name of the repo).
- Adding the full name of Parity Signer, as it appears on mobile stores
- Adding the Staking ability for Signer.

### DIFF
--- a/docs/build-wallets.md
+++ b/docs/build-wallets.md
@@ -25,11 +25,11 @@ that the wallet software is using the latest version of the
 
 ### Supported Wallets
 
-| Wallet Name                                               | Development State | Team Name | Description       | Custody       | Supports |
-| --------------------------------------------------------- | ----------------- | --------- | ----------------- | ------------- | -------- |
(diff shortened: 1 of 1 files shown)


Swader:
Signer has no staking ability, though. It only allows you to interact with staking-enabled interfaces.

Tbaut:
Well if you see it like this, then neither can the extension. 
The way I see it, is that you can stake the dots hold by the extension or Signer, you can't if you're using say Trust wallet.

I actually made the PR as a result of a discussion that happened on Element a while ago (couldn't find it again, @laboon was part of it and might remember).

Swader:
I think it's somewhat misleading but won't fight it, fine by me.

Tbaut:
I don't have a strong opinion tbh, as long as both Signer and the extension are in the same category.

PR #1694, opened 2021-02-11
docs: fix argument of --pruning flag

asymmetric:
@Swader I was also wondering if we should mention a reasonable value, e.g. 1000.


Swader:
1000 sounds fine, yes, let's mention it.

PR #1705, opened 2021-02-13
Clarifying which sale the claims process applies to

Swader:


laboon:
There was some yelling from the CI but it doesn't seem related to this

PR #1720, opened 2021-02-16
Fixed a typo.

evgeny-s:


Swader:
Thank you - there's no need to fix translation files, those update automatically

PR #1726, opened 2021-02-17
Add Polkadot JS Desktop App to Wallets page

krzysztof-jelski:
As part of the [Open Grant](https://github.com/EthWorks/Open-Grants-Program/blob/master/applications/polkadot-desktop-app.md) we've packaged the Polkadot JS Apps UI as an Electron desktop app. This version stores account on user's hard drive, so it can be considered much safer than the web UI (which uses browser's local storage).

The Electron App is maintained and supported by @jacogr and myself.

### DIFF
--- a/docs/build-wallets.md
+++ b/docs/build-wallets.md
@@ -25,11 +25,12 @@ that the wallet software is using the latest version of the
 
 ### Supported Wallets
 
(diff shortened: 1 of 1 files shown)


Swader:
Good idea

Issue #1219, opened 2020-10-27
Add practical detail on multisig transactions


laboon:
Right now our section on multisig transactions ( https://wiki.polkadot.network/docs/en/learn-accounts#making-transactions-with-a-multi-signature-account ) is rather theoretical.  We should include some detail including screenshots of how to actually do a multisig tx.

Swader:
Closed in #1799

PR #1799, opened 2021-03-06
Add multisig deposit info, practical details on multisig transactions

### LINKED ISSUE #1677: Multisig deposits - add to multisig info
Each account should be funded with enough tokens in order to pay the multisig reverve (3.34 KSM or 20.152 DOT).
For any multisig transaction the initiator for that transaction needs to have about 3.4 KSM in their account.  
To create an anonymous proxy the account creating the proxy needs to have 3.4 KSM
To create an anonymous proxy controlled by a multisig, both the initiator and the multisig account need to have 3.4 KSM.

jakehemmerle:
- Clarified multisig transaction initiation deposit
- added practical example of how to do multisignature transactions with screenshots, etc.

Should close #1677 and #1219


Swader:
This is great. Note for future: when needing to test and demo something for Polkadot / Kusama, it's generally cheaper and faster to just spin up a local version of either with the `--dev` flag - that way you keep the constants and chain params as they are on the live version, and have plenty of money to play with due to the demo unlocked accounts.

PR #1819, opened 2021-03-08
docs: added info on how to get runtime diffs

### LINKED ISSUE #1802: Add diff between Polkadot and Westend for devs
Known notable differences are:
- Existential deposit: 1 DOT vs 0.01 WND
- multisig deposit: ~20 DOT vs 1 WND

0.01 would make it seem like Westend did not go through the 100x upgrade, but that doesn't explain why multisig deposit isn't off by the same factor.

As @laboon suggested, we can do a diff on the runtimes:

DOT:
https://github.com/paritytech/polkadot/tree/master/runtime
and here in Polkadot : https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L231

Westend: https://github.com/paritytech/polkadot/blob/58ff7f59e9a9e43f5b1e30a126ee023a41af9d39/runtime/westend/src/lib.rs#L207

jakehemmerle:
Should close #1802 

Added to this page after "Westend Test Network": https://wiki.polkadot.network/docs/en/maintain-networks#network-differences

### DIFF
--- a/docs/maintain-networks.md
+++ b/docs/maintain-networks.md
@@ -60,7 +60,7 @@ Check your node is connected by viewing it on
 
 ### Differences
 
-Runtime differences (ie. existential and multisignature deposit sizes) between the different
+Runtime differences (e.g. existential and multisignature deposit sizes) between the different
 networks can be found by doing a `diff` between the `src/lib.rs` of the respositories. For example,
 to compare the Polkadot and Westend runtimes:
 
--- a/docs/maintain-networks.md
+++ b/docs/maintain-networks.md
@@ -68,6 +68,12 @@ to compare the Polkadot and Westend runtimes:
 - `ls` - show the available runtimes
 - `diff polkadot/src/lib.rs westend/src/lib.rs`
 
+You can also paste the runtimes
+([Polkadot](https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs),
+[Westend](https://github.com/paritytech/polkadot/blob/master/runtime/westend/src/lib.rs)) into a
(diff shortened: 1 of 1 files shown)


docs/maintain-networks.md
```diff
@@ -58,6 +58,16 @@ Follow the instruction [here](learn-DOT#getting-westies) for instructions on acq
 Check your node is connected by viewing it on
 [Telemetry](https://telemetry.polkadot.io/#list/Westend).
 
+### Differences
+
+Runtime differences (ie. existential and multisignature deposit sizes) between the different
```
Swader:
```suggestion
Runtime differences (e.g. existential and multisignature deposit sizes) between the different
```

docs/maintain-networks.md
```diff
@@ -58,6 +58,16 @@ Follow the instruction [here](learn-DOT#getting-westies) for instructions on acq
...
+networks can be found by doing a `diff` between the `src/lib.rs` of the respositories. For example,
+to compare the Polkadot and Westend runtimes:
+
+- `git clone https://github.com/paritytech/polkadot && cd polkadot/runtime`
+- `ls` - show the available runtimes
+- `diff polkadot/src/lib.rs westend/src/lib.rs`
+
```
Swader, line 70: Maybe add here a few links to some web-based diff tools that might do the job too, not everyone will be comfortable on the command line.

Swader (changes requested):

jakehemmerle:
@Swader added a web-based diff tool. Not sure why changes aren't being showed in this view though

Issue #1731, opened 2021-02-18
Get rid of CircleCI


whitesunset:
Self assigned

salmad3:
PR for fix merged.

PR #1831, opened 2021-03-09
fixes #1813, Use esm module imports synthax instead of commonJS and update @polkadotjs/api to version 4.0.3

### LINKED ISSUE #1813: Make Github Actions use NodeJS 14+ due to new ESM requirement
Polkadot API now uses ESM which breaks on anything below NodeJS 14.

Yuripetusko:
use esm synthax for node scripts


Swader:
Thanks but I don't think this is needed any more with API version 4.0.3?

Swader:
Ah, no, I didn't test our custom scripts on below v14 of node yet, could very well be it doesn't work. But rather than make the custom commands more complex, I would rather use this opportunity to clean house a little bit.

Yuripetusko:
Ok fair enough @Swader. I have now updated this PR to use ESM modules purely without using babel transpiler. But for some reason CircleCI fails (I cannot access it to see the details)

Swader:
Yeah we're using some kind of docker image with an outdated node it seems, need to figure out what's up or how to update it.

image

I would *really* love to move away from Circle and onto Github Actions in full (#1731), as Circle is just a nuisance at this point.

Swader:
Due to the error message shown, I don't think these v13 nodes are the problem (though they should absolutely also be updated), instead, I think this is the offending line: https://github.com/w3f/polkadot-wiki/blob/master/.circleci/config.yml#L20 - which means we should somehow update node inside of that image. And I have no idea what that image is, where to find it, or how to update it.

Yuripetusko:
Looks like it is this particular image

https://github.com/w3f/polkadot-wiki/blob/master/.circleci/config.yml#L20

Swader:
Ah, looks like it's this: https://github.com/w3f/link-checker

Swader:
I do not know or like Docker, are you able to figure out how to update this from node10 to node14?

Yuripetusko:
I hate docker myself, but I can try :D might have to fork docker config

Swader:
Actually, this image itself depends on Circle, so even if we fix it it'll be troublesome to move from Circle to Github actions. I would actually propose we remove the link checker altogether and build an alternative because this one never really worked well anyway. Yes, let's do that.

Yuripetusko:
Should I remove linkchecked from circleci config in this PR?

Swader:
Yes please, let's kill it with fire

Yuripetusko:
Remove linkchecker CircleCI section

Yuripetusko:
Whoop! it's all green now

Swader:
Hah, nice, thanks!

PR #1871, opened 2021-03-19
Clarified and updated bridging methods

### LINKED ISSUE #1798: Feedback from Tomasz on bridges
Related to this - https://wiki.polkadot.network/docs/en/learn-bridges#types-of-bridges

"Honestly I see this repeated many times, but I never understand the split between Bridge modules and Bridge contracts.

For instance Ethereum bridge is a smart contract on Ethereum (obviously) but a pallet (assuming "a bridge module") on the Polkadot side.
Kusama <> Polkadot bridge is a bunch of pallets that need to be included on (relay/para) chains on both networks.
The bridge should always imply bi-directional communication, while it's possible to have one-way bridge which is what I kind of read from the "Bridge module" paragraph, it's not super useful."

jakehemmerle:
Should close #1798

### DIFF
--- a/docs/learn-bridges.md
+++ b/docs/learn-bridges.md
@@ -19,35 +19,36 @@ page a work in progress. It will be updated as more information is determined an
 > interoperability technology that allows parachains to trustlessly communicate, please see the
 > dedicated [crosschain][] page on the Wiki.
 
-# Types of bridges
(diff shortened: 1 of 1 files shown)


Swader:
Nice refresh

Issue #1816, opened 2021-03-08
Add instructions for converting vote bitfield to actual vote


laboon:
If you look at your votes on a block explorer, trying to understand what you voted on is difficult, since it's a bitfield with both conviction and the aye/nay boolean stored in one byte. This is usually displayed by block explorers as a decimal integer, so it's very confusing.

```
Nay 0.1 => 0
Nay 1x => 1
Nay 2x => 2
Nay 3x => 3
Nay 4x => 4
Nay 5x => 5
Aye 0.1x => 128
Aye 1x => 129
Aye 2x => 130
Aye 3x => 131
Aye 4x => 132
Aye 5x => 133
```

laboon:
Also add Aye 6x and Nay 6x (hopefully it should be easy to figure out what they should be =) )

laboon:
https://wiki.polkadot.network/docs/en/maintain-guides-democracy

Add a section called "### Interpreting On-Chain Voting Data" at the bottom

https://github.com/paritytech/substrate/blob/9b08105b8c7106d723c4f470304ad9e2868569d9/frame/democracy/src/vote.rs#L40

salmad3:
PR reviewed by Bruno & site updated.

PR #1882, opened 2021-03-22
Add Section for Interpreting On-Chain Voting Data

salmad3:
Addresses: https://github.com/w3f/polkadot-wiki/issues/1816


Swader:
Interesting

PR #1890, opened 2021-03-23
Add Jake as an active wiki contributor

jakehemmerle:


salmad3:
Noticed we both opened a PR against contributor.md.

PR #1894, opened 2021-03-23
Updated links for Validator Setup document.

quopius:
First time using github, sorry if I didn't do it correctly, just trying to help!


Swader:
Thanks, there's no need for `docs` or `.md` in the files, you can just reference their filenames, like `learn-phragmen` and that'll work. For best results on checking if these links work when you change them, you should download and test locally by running the wiki on your machine. We'll check this out and fix if needed.

salmad3:
@quopius thanks for the effort. There was a link that had been updated recently on the validator setup doc, and as was mentioned adding the `docs` and `.md` isn't necessary. Will close.

PR #1895, opened 2021-03-23
fix collator running instruction

stechu:
When running collator that is intended to connect to rococo testnet, the chain spec need to be separated by `--`.


Swader:
Sorry, missed this. Can you bring it up to date with master?

stechu:
@Swader the instruction this PR intended to fix no longer exists, LOL.

PR #1950, opened 2021-04-07
Fixes #1226

### LINKED ISSUE #1226: [STALE] learn-comparison-dfinity.md
Last edit happened 150 days ago 😬

Swader:
Only a mild update to the Dfinity post, but I am tempted to remove Dfinity from the wiki altogether as I do not have the time to dive into what's been released since Logan's writeup and do an objective comparison.


laboon:
Agreed. Nobody's asking about DFINITY anymore and we don't want to waste time maintaining it if so.  Feel free to just remove it.

PR #1982, opened 2021-04-16
Update thousand-validators.md

paradox-tt:
Changed line 94 from 5'000 DOTs to 5,000 DOTs

### DIFF
--- a/docs/thousand-validators.md
+++ b/docs/thousand-validators.md
@@ -91,7 +91,7 @@ and fill in the [application form][polkadot 1kv form]. You will hear back from t
 - Connect to dedicated Telemetry (use
   `--telemetry-url 'wss://telemetry-backend.w3f.community/submit 1'` when starting the node)
 - Rank 25 or higher on Kusama Thousand Validators Programme
-- Mininmum of 5,000 DOTs self stake (exceptions by approval for good intentions)
+- Mininmum of 5_000 DOTs self stake (exceptions by approval for good intentions)
 - Reward destination 'Staked'
 - No more than 3% commission
 - Separate stash and controller (or have a Staking proxy set up)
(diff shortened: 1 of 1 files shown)


wpank (approved):

laboon:
I'd prefer not to use commas or periods where they could be confused (e.g. 5,000 could be seen as five point zero zero zero by Europeans or five thousand by Americans). I personally prefer the Swiss way of using `'` for the thousands separator but I'm open to other suggestions.

Would 5_000 be easier to understand?

Swader:
I think `_` makes sense because it matches the Rust syntax.

docs/thousand-validators.md
```diff
@@ -91,7 +91,7 @@ and fill in the [application form][polkadot 1kv form]. You will hear back from t
 - Connect to dedicated Telemetry (use
   `--telemetry-url 'wss://telemetry-backend.w3f.community/submit 1'` when starting the node)
 - Rank 25 or higher on Kusama Thousand Validators Programme
-- Mininmum of 5'000 DOTs self stake (exceptions by approval for good intentions)
+- Mininmum of 5,000 DOTs self stake (exceptions by approval for good intentions)
```
laboon:
```suggestion
- Mininmum of 5_000 DOTs self stake (exceptions by approval for good intentions)
```

laboon:
Switched to _ and merged

paradox-tt:
I might be a little late to the party, I hope there's still some non-alcoholic beverages around :)

Mr. Laboon enlightened me as the the different representations of this pesky "thousand separator" as used in different regions of the world.  

My humble take on it now that I'm enlightened is that whatever is used be used consistently across the site.  I do note that when it comes to blocks and other numeric references the ',' is used in other areas of the site.

laboon:
Thanks @paradoxxx-tt . You are correct that no matter what separator we use, we should be consistent across the site.

I just ran some greps with regexes, and it turns out we used `'` in some places and `,` in others. Another PR was just merged ( https://github.com/w3f/polkadot-wiki/pull/1989 ) which changes all thousands separators to `_`.

PR #1992, opened 2021-04-21
Update Identity Page

### LINKED ISSUE #582: Update Identity Page for Polkadot Amounts
Right now the identity page lists some ksm amounts for setting an identity. 

> Users can register some default fields like legal name, display name, website, Twitter handle, Riot handle, etc. along with extra, custom fields for which they would like attestations (see Judgements). Users must reserve funds in a bond to store their information on chain - 10 KSM per identity, and 2.5 KSM per each field beyond the legal name. These funds are locked, not spent - they are returned when the identity is cleared. Each field can store up to 32 bytes of information, so the data must be less than that. When inputting the data manually through the Extrinsics UI, a UTF8 to bytes converter can help.

These amounts have changed since then, but it should also be added what the costs in DOTs are.

### LINKED ISSUE #1926: Update Identity Page
https://wiki.polkadot.network/docs/en/learn-identity

salmad3:
fixes #582,  fixes #1926; add DOT amounts + edits

### DIFF
--- a/docs/learn-identity.md
+++ b/docs/learn-identity.md
@@ -12,11 +12,14 @@ on-chain account and subsequently ask for verification of this information by
 
 Users can set an identity by registering through default fields such as legal name, display 
 name, website, Twitter handle, Riot handle, etc. along with some extra, custom fields for which 
-they would like attestations (see [Judgements](#judgements)). Users must reserve funds in a bond 
-to store their information on chain (for both Polkadot and Kusama) - 20 DOT or 1.666 KSM per identity, 
-and 1 DOT or 0.416666 KSM per each field beyond the legal name. These funds are _locked_, not spent - 
-they are returned when the identity is cleared. Each field can store up to 32 bytes of information, 
-so the data must be less than that. When inputting the data manually through the 
+they would like attestations (see [Judgements](#judgements)). 
+
+Users must reserve funds in a bond to store their information on chain (for both Polkadot and Kusama):
+{{ polkadot_identity_reserve_funds }} DOT or {{ kusama_identity_reserve_funds }} KSM per identity, and 
+{{ polkadot_identity_field_funds }} DOT or {{ kusama_identity_field_funds }} KSM per each field beyond 
+the legal name. These funds are _locked_, not spent - they are returned when the identity is cleared. 
+Each field can store up to 32 bytes of information, so the data must be less than that. When inputting 
+the data manually through the 
 [Extrinsics UI](https://polkadot.js.org/apps/#/extrinsics), a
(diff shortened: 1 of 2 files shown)


Swader:
I propose you try implementing this with conditional rendering.

The steps are:

- define the value in `scripts/inject_dict`
- inject the value by using `{{ my_key }}` into the relevant file
- run the injection locally to see if it works

Let me know if the README is not adequately detailed about this and I will be happy to clarify steps.

Swader:
These changes only reference hardcoded defaults from the dict, which is no different from just hardcoding them into the text. I would like you to:

- make the dict actually fetch the live data from chain (the defaults are there only as a fallback)
- split up the conditional so that Kusama's version only refers to Kusama, and Polkadot's version only refers to Polkadot. Right now the page is mirrored with both, so both Kusama and Polkadot talk about both Kusama and Polkadot. Kusama's page should have only values pertaining to Kusama, and Polkadot to Polkadot.

salmad3:
> These changes only reference hardcoded defaults from the dict, which is no different from just hardcoding them into the text. I would like you to:
> 
> * make the dict actually fetch the live data from chain (the defaults are there only as a fallback)
> * split up the conditional so that Kusama's version only refers to Kusama, and Polkadot's version only refers to Polkadot. Right now the page is mirrored with both, so both Kusama and Polkadot talk about both Kusama and Polkadot. Kusama's page should have only values pertaining to Kusama, and Polkadot to Polkadot.

Right that's a good point. I was just a little confused on how to get the pages to recognize the different values since the doc is the same for both: https://wiki.polkadot.network/docs/en/learn-identity#docsNav, https://guide.kusama.network/docs/en/mirror-learn-identity#docsNav since atm there is only mention of KSM reserve funds.

I originally added 

```
  {
    "tpl": "identity_reserve_funds",
    "default": { "kusama": 1.666, "polkadot": 20 },
    "path": "consts.identity.reserveFunds",
    "filters": ["humanReadableToken"]
  },
  {
    "tpl": "identity_field_funds",
    "default": { "kusama": 0.41666, "polkadot": 1 },
    "path": "consts.identity.reserveFieldFunds",
    "filters": ["humanReadableToken"]
  }
```

but wasn't sure how they would recognize the different chains if being referenced in the same doc, if that makes sense. Also, is the path self-defined? If you don't mind clarifying steps that would be great!

Swader:
The [mirror](https://github.com/w3f/polkadot-wiki/blob/master/scripts/inject.mjs) script should reveal all mysteries. Not sure what you mean by the path being self-defined, but depending on which chain it's injecting to, that's the API it's going to use. So if you put "reserveFunds" into a page, it's going to read from the polkadot chain in the polkadot wiki, and the kusama chain in the kusama wiki

PR #2016, opened 2021-04-28
fix grammatical error in section title

jkhaui:
`How to connect to a Parachains` should be `How to connect to a Parachain`


Swader:
Thanks!

PR #2019, opened 2021-05-01
Update learn-accounts.md

reenigneEsrever92:
It says here there are four types of balances, which are: free, reserved, misc_frozen and fee_frozen. So I guess this was a mix up!?


Swader:
No, [fee_frozen](https://github.com/paritytech/substrate/blob/c93ef27486e5f14696e5b6d36edafea7936edbc8/frame/balances/src/lib.rs#L585) is correct.

Swader:
Oh, sorry, I thought you put in free instead of fee but I see now that you did the other way around, thanks!

PR #2022, opened 2021-05-03
replace incorrect word in democracy wiki

jkhaui:
`will it can be voted upon` should be `where it can be voted upon`


Swader:
Thanks!

PR #2030, opened 2021-05-04
minimum nomination now the ED

nuke-web3:


laboon:
This is incorrect, though - there is still a minimum, it's just that it's generally 0 (you need to stake the ED but split nominations can occur down to very small values, I guess theoretically to 1 Planck). But if we get a bunch more nominators come in, the "stripping away" of low DOT nominations will still occur.

nuke-web3:
What would you suggest specifically @laboon ? 

I also see this information conflicting on other pages... should it just be removed completely from all but one page?

Open to direct suggestions in the PR files here, or in the comments so I can amend 🙏🏼

nuke-web3:
should be this _or_ #2026

Swader:
@NukeManDan can you bring this up to date with master so I can merge it?

nuke-web3:
Sure, but as per the conversation here, @laboon should chime in on the correct content, as this may be misleading, and we should at least leave an indication that the minimum may be an issue again, and if so, what would be the cause.

docs/learn-nominator.md
```diff
@@ -99,6 +87,20 @@ validator's nominators. Any account can do this, although in practice validator
...
+Due to the way the [Phragmen algorithm](learn-phragmen) generates the solution set, there was
+**previously** a minimum number of DOT needed to nominate with in order to receive staking
+rewards due to the fact that the solution set need to be both proccessed and stored within
+a single block, 6 seconds on Polkadot.
+
+A [runtime upgrade](build-protocol-info#runtime-upgrades) moved the logic to process the rewards
+for nominators to an [off-chain worker](learn-phragmen#off-chain-phragmén), and now the minumum
+DOT required is only limited by the [existential deposit](build-protocol-info#existential-deposit)
+See the [Election Solution Set](learn-nominator#the-election-solution-set) section on the
+Nominator page for more details.
```
Swader:
```suggestion
DOT required is only limited by the [existential deposit](build-protocol-info#existential-deposit).
Note that this might change if enough new nominators nominate. In that case, the minimum can 
rise above the ED again. This is unlikely to happen in the next 6 to 12 months.

See the [Election Solution Set](learn-nominator#the-election-solution-set) section on the
Nominator page for more details.
```

Swader: Something like this should do?

nuke-web3: I am generally against any reference to the present in specific terms like " next 6 to 12 months" - from when? Readers don't know when and this doc (likely) will go stale before it's corrected if this happens.

Swader: Agreed, but "in 2021" felt strange and leaving it out makes things sound less safe than they are, leaving readers with uncertainty. Ideas?

laboon: How about "Although at the time of writing, with the number of nominators at ~ 25_000, there is no minimum, if the number of nominators increases dramatically in the future, the solution will be pruned and nominators with the least amount of stake will once again be pruned and receive no staking rewards."

nuke-web3: I have updated the text, take a look at the changes that sum up what you suggest without referencing a specific time @laboon 🙏🏼

laboon: lgtm

PR #2038, opened 2021-05-07
Update build-build-with-polkadot.md

nuke-web3:
- remove "pending mainnet launch" 
- refactor text
- remove explicit versions of code
- error corrections
There is more work on content that relates to "costs" of parachains vs. parathreads  out of scope of this PR


Swader:
Thank you!

PR #2039, opened 2021-05-08
Added  --chain kusama to command line

playing-with-dust:
I found it confusing that these commands create a polkadot node so I couldn't find it in the kusama telemetry, and then had to look elsewhere to find the correct switch.


Swader:
Thanks though this is a document that applies equally to Kusama and Polkadot so needs a bit of [conditional rendering](https://github.com/w3f/polkadot-wiki#conditional-rendering). We'll handle it.

playing-with-dust:
Nice one - thank you, and also thanks for the incredibly comprehensive documentation!

PR #2065, opened 2021-05-15
spelling: adversarial not adverserial

nuke-web3:
spelling


Swader:
Thanks, note that you don't need to modify the translation files, those are done automatically

PR #2080, opened 2021-05-22
git clone for users without a w3f github account

Amallyn:
For users who don't access github via ssh


Swader:
Thanks but SSH is a sane default, there's no need to change it. If someone wants to clone the wiki via HTTPS, they can do so by looking up the URL.

PR #2086, opened 2021-05-24
Update Amounts & Add Content for Chain state Constants

salmad3:


Swader:
Needs prettier.

`npx run prettier --save` IIRC, or just use an editor that will auto-read prettier configuration in the wiki repo, like VSCode.

Swader:
Fix conflicts please and we can merge

salmad3:
> Fix conflicts please and we can merge

conflicts were fixed - merging

PR #2077, opened 2021-05-20
Migrate to Docusaurus v2

nam-hle:
The PR is to migrate two websites to the Docusaurus v2. Here is noticeable changes:

# Images

- Moved all documentation-related images from the sites' `static` folder to `/docs/assets`. It reduces the number of duplicated images between two sites and be less confused when referencing them in docs (all image paths will be started with `assets/...`, no more mixed with `img/...`)

**Note**: There are 02 referenced images I can't find elsewhere in `docs/maintain-guides-how-to-vote-councillor`.

# Documentation

- Some broken links: I changed to `#` for simplicity. Please fix them with proper links for me (by review maybe :D)
- `` element: it needs to change `src` attribute from absolute one to `require`, otherwise the Docusaurus will throw not found errors.

# Injection dictionary values

We can't replace `{{ ... }}` syntax in HTML files with true values after the building step due to the new Docusaurus server-side mechanism. So my solution is:
- Fetch the necessary values then saved them in a JSON file temporarily.
- After the markdown abstract syntax trees (AST) are created, a new `remarkable` [plugin](https://github.com/nam-hle/polkadot-wiki/blob/migrate/scripts/injectPlugin.js) will transform placeholder value to real one from candidate tree nodes (`text`, `paragraph`,...) in build time **before** generating HTML files.

# i18n

- Docusaurus v2 requires a specific directory structure for i18n (see [more](https://docusaurus.io/docs/i18n/introduction) ):
```
website/i18n
└── zh-CN
    ├── code.json: contains keys for general UI labels as well as our custom page keys
    │
    ├── docusaurus-plugin-content-docs
    │   ├── current: contains translated docs
    │   │   ├── doc1.md
    │   │   └── doc2.md
    │   └── current.json: keys for category labels in sidebar
    │
    └── docusaurus-theme-classic: keys for specific components
        ├── footer.json
        └── navbar.json
```

- With new [`Translate API`](https://docusaurus.io/docs/docusaurus-core#translate), we can translate [homepage](https://github.com/nam-hle/polkadot-wiki/blob/migrate/website-v2/src/pages/index.js), footer as well as category label in the sidebar completely. 
- We also only need to save the source string in `i18n/en`, other languages can be downloaded in build time. Currently, I only added the scripts to download two current supported language documentations: Russian and Chinese Simplified. (Crowdin CLI does not allow to download multiple languages once so I use two separate scripts for it. Please let me know if there are possible solutions)
[...]

PR #2091, opened 2021-05-25
Run Prettier

joepetrowski:
Why was this removed???


Swader:
Prettier? Because we're migrating to Docusaurus V2 and there we'll re-integrate all these helper CI things, it's been broken for a while esp. on forks, hacky initial setup.

joepetrowski:
But you haven't migrated... I don't see Docusaurus as a reason to get sloppy about maintaining some quality standards.

Swader:
The migration is under way in #2077

PR #2143, opened 2021-06-10
Update glossary

### LINKED ISSUE #2031: Terms to add to Glossary
Had a question about some stuff in the Polkadot.js >Network >Staking. There are few things here that are not too clear.
 
Next Session - What does it mean?
Online message  - What does it mean? 
Oversubscribed -Is that based on a number of nominators or the amount staked? 
Slashed - How would one go about finding out why someone was slashed?
Blocks nominations  - What does it mean? 

Might be worth adding a link to the glossary sections on these elements in the UI

laboon:
Added a bunch of definitions to the glossary

Most of these (especially teleport) also need their own or other pages updates

 fixes #2031


salmad3 (approved):
Oh was going to open a PR; thanks Bill. LGTM.

PR #2146, opened 2021-06-10
Subtract should be plus in total balance

laboon:


laboon:
See here if you don't believe me =)

https://github.com/paritytech/substrate/blob/eb9033b826b9e8115c20707fe66af0ceb177e99c/frame/balances/src/lib.rs#L42

wischli:
Thanks! We were really confused by the Wiki entry.

salmad3:
Just need to resolve conflicts caused by: https://github.com/w3f/polkadot-wiki/pull/2145

tjwelde:
The Pr seems to want the same as we wanted to achieve with #2145. So if @laboon doesn't want to change wording, I think we can close this.

salmad3:
> The Pr seems to want the same as we wanted to achieve with #2145. So if @laboon doesn't want to change wording, I think we can close this.

Can confirm that PR changes satisfy the same change in this PR - will close as duplicate.

PR #2171, opened 2021-06-16
Common good parachains info

laboon:
Change "public utility" term to "common good"
Add IDs and FAQs


laboon:
DO NOT MERGE - need to update based on conversation with Joe

emresurmeli:
Please make sure to rebase master and check to see your changes are reflected locally before continuing work. We merged a new version of Docusaurus and have ran basic checks to see if it works, but could run into minor issues.

PR #2195, opened 2021-06-21
Update Crowdloan Content

salmad3:
…#2141

Listed crowdloans are based on ones listed on polkadotjs apps


joepetrowski:
This is a really bad idea. There's no way it will be kept up to date, missing teams will rightfully complain about favoritism, it's not scalable/sustainable as it requires manually checking the Crowdloan status all the time and issuing a new PR for every change, and an error here pointing to a scam would be highly embarrassing for us.

This should be left to Polkadot JS Apps the same way referenda/motions/tips/etc. have a link to the Polkassembly/Subscan/etc. discussions.

salmad3:
Right that's true. It could serve as a point of reference if kept up to date the right way. A portion of the teams also reference the page.

On the page right now we mention:

```
Several teams in the ecosystem are preparing dashboards to track auctions and crowdloan campaigns. Check back with this page for updates as these are released.
```

So this is just trying to satisfy that.

joepetrowski:
Well we should re-evaluate that statement, not blindly follow along.

salmad3:
It makes sense to have a point of reference for users (especially for something new) to go off of, although yes, risky & unsustainable. It's now updated without the links to the dashboards.

joepetrowski:
Why don't you link to the [Apps page](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fkusama-rpc.polkadot.io#/parachains/crowdloan), which lists each project's homepage and is maintained and sustainable?

jakehemmerle (approved):
lgtm!

salmad3:
Joe's feedback addressed and reviewed by Jake - merging

PR #2196, opened 2021-06-21
Check for Stale before Creating Issue

salmad3:


salmad3:
will revert if unsuccessful (just can't trigger a manual run)

PR #2217, opened 2021-06-23
Fix redirect

nam-hle:


Swader:
Might need revert, testing in prod

PR #2223, opened 2021-06-24
Add job to check for stale issues + Attempt fix

salmad3:
- adding initial job implementation to check for stale issues. Motive: many older issues with merit have been opened and forgotten - others outdated or N/A
- attempt fix for duplicate check


salmad3:
will revert if unsuccessful

PR #2239, opened 2021-06-27
Update contributors.md

salmad3:
@DrW3RK can add content to this PR :) and anyone else for that matter


laboon:
I think we should merge this and have @DrW3RK create his own PR to add himself to the contributors page.

PR #2255, opened 2021-06-30
Fix wrong order in landing page

nam-hle:
The order of navigating sections is not right after #2251. Also, reduce the delay time.


salmad3:
Oh I just noticed the PR, thanks @nam-hle

PR #2289, opened 2021-07-05
Update maintain-polkadot-parameters.md

### LINKED ISSUE #2225: Update new staking parameters
Max. 20,000 nominators; min. stake 20 DOT

https://polkadot.network/polkadot-and-kusama-staking-changes/?utm_source=twittee&utm_medium=social&utm_campaign=NewStakingLimits&utm_content=blogPolka

### LINKED ISSUE #2282: Minimum nomination outdated
https://wiki.polkadot.network/docs/learn-nominator/#required-minimum-stake

This points to scripts that I believe are no longer useful with recent changes in the network (I have not tested 😅 )

#2030 was reverted that included (aslo now deprecated) info, I think the reason being wording. Possibly best practice not to set a specific number (as this may change before someone updates this wiki again)   but a reference to somewhere that one can actually find this, say on the apps UI would be good IMHO.

DrW3RK:
As this is my first update to the Wiki, I am making sure I get the convention right

Do we link to referendums and include dates when the changes happen? 

Also, is this the correct location where we update the details or should we update this in FAQ page? Thanks!

### DIFF
--- a/docs/learn-nominator.md
+++ b/docs/learn-nominator.md
@@ -52,7 +52,7 @@ Due to the way the [Phragmen algorithm](learn-phragmen.md) generates the solutio
 fact that the solution set must fit in a single block, in some eras, a minimum number of DOT will be
 required to nominate with in order to receive staking rewards.
 
-This number fluctuates every era, and the most recent and up to date version can be found on [chain state](https://polkadot.js.org/apps/#/chainstate) (selected state query - staking - minimumNominatorBond)
+This parameter can be updated via on-chain governance and the most recent and up to date version can be found on [chain state](https://polkadot.js.org/apps/#/chainstate) (selected state query - staking - minimumNominatorBond)
 
 ### Oversubscribed Validators
 
--- a/docs/learn-nominator.md
+++ b/docs/learn-nominator.md
@@ -52,11 +52,7 @@ Due to the way the [Phragmen algorithm](learn-phragmen.md) generates the solutio
 fact that the solution set must fit in a single block, in some eras, a minimum number of DOT will be
 required to nominate with in order to receive staking rewards.
 
-This number fluctuates every era, and the most recent and up to date version can be found out using
-[these scripts](https://github.com/w3f/validator-stats) which you can run on your machine and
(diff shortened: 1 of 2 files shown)


DannyS03:
Thanks @DrW3RK, would also add it on the [Nominator](https://wiki.polkadot.network/docs/learn-nominator/#required-minimum-stake) page. A similar issue was opened over the weekend: https://github.com/w3f/polkadot-wiki/issues/2282

I'd say it's a good idea to add some info on the FAQ page - especially since this update aims to be temporary. There have also been many questions/concerns about it. In this case, it could be something along of the lines of "Why did the minimum stake change?" "Until When?", "What's Next?" etc... 

I don't believe we link or include dates when changes happen. 
 
It is also good practice to reference the issue if the PR (or commit) aims to address it. Here's some [GitHub Docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for reference.

salmad3 (commented):
Thanks @DrW3RK, would also add it on the [Nominator](https://wiki.polkadot.network/docs/learn-nominator/#required-minimum-stake) page. A similar issue was opened over the weekend: https://github.com/w3f/polkadot-wiki/issues/2282

I'd say it's a good idea to add some info on the FAQ page - especially since this update aims to be temporary. There have also been many questions/concerns about it. In this case, it could be something along of the lines of "Why did the minimum stake change?" "Until When?", "What's Next?" etc... 

I don't believe we link or include dates when changes happen. 
 
It is also good practice to reference the issue if the PR (or commit) aims to address it (generally speaking). Here's some [GitHub Docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for reference.

DrW3RK:
@DannyS03 Thank you very much for all the suggestions above. I believe this takes care of the issues linked to this PR for the time being. I will work on implementing dynamic inserts through a new PR. If everything looks good, this can be merged with Master.

salmad3 (approved):

PR #2297, opened 2021-07-06
New Page: Runtime Upgrades 

### LINKED ISSUE #1039: Document Runtime Upgrades
I was rather surprised that we don't have a page or even section on runtime upgrades. Should include:

- Runtime upgrade overview
  - Similarities and differences to a hard fork
  - "There's a new client release, when do I need to update by?"
- What does it mean if I am a ______?
  - Validator -> probably nothing, check client release notes for urgency
  - Nominator -> definitely nothing, just relax, man
  - Infrastructure provider (block explorer, exchange) -> update parsing logic, sidecar, may be new types
  - Wallet -> watch out for tx broadcasting (i.e. don't construct with spec `n` and broadcast with spec `>n`)
- How to monitor for changes (see below)

1. Monitor chain for `democracy.Started` events and log `index` and `blockNumber`.
2. Get `pallets/democracy/storage/ReferendumInfoOf?key1=index&at=blockNumber` from Sidecar to get the referendum info. It should have a status of `Ongoing`. Find ending block number (`end`) and the enactment delay (`delay`).
3. Execution block number will be `end + delay`.
4. Monitor chain for `democracy.Passed`, `democracy.NotPassed`, or `democracy.Cancelled` events citing the index. If `Passed`, need to look at `scheduler.Scheduled` event in same block for enactment block.
5. Monitor chain for `democracy.PreimageNoted` events with the same hash as the `ReferendumInfoOf(index)` item, may be up to the last block before execution, but will not work if this is missing.
6. Monitor chain for `democracy.Executed` events for actual execution. In the case of a runtime upgrade, there will also be a `system.CodeUpdated` event.

salmad3:
Resolves: #1039

### DIFF
--- a/docs/build-protocol-info.md
+++ b/docs/build-protocol-info.md
@@ -223,25 +223,8 @@ transaction data using the
 
 ## Runtime Upgrades
 
-Runtime upgrades allow Polkadot to change the logic of the chain without the need for a hard fork. A
-hard fork would require node operators to manually upgrade their nodes to the latest runtime
-version. In a distributed system, this is a complex process to coordinate and communicate. Polkadot
-can upgrade without a hard fork. The existing runtime logic is followed to update the Wasm runtime
-stored on the blockchain to a new version. The upgrade is then included in the blockchain itself,
-meaning that all the nodes on the network execute it.
-
-Generally there is no need to upgrade your nodes manually before the runtime upgrade as they will
-automatically start to follow the new logic of the chain. Nodes only need to be updated when the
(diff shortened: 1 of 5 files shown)


jakehemmerle:
Make sure to merge latest master into here as this was started on a broken version of master: error is

```
Error: Invalid sidebar file at "website-v2/sidebars.js".
These sidebar document ids do not exist:
- maintain-guides-how-to-use-polkadot-secure-validator
```

jakehemmerle (approved):

salmad3:
Merging after Jake's review

PR #2307, opened 2021-07-08
Update learn-account-generation.md

### LINKED ISSUE #2267: Add information about deriving new accounts on PS on Account Generation page
https://wiki.polkadot.network/docs/learn-account-generation/

If you "derive new account", you will also need to add a different derivation path. This can be as simple as //1, //2, etc. but you do need different ones for each derived account. Using the seed plus the derivation path, you now have multiple totally separate account that share the same seed.

DrW3RK:
Added placeholder for a tutorial video. Also, will work on augmenting the tutorial with more instructions


DrW3RK:
Added minor updates. Can benefit from a detailed video and instructions on how to operate the account using Parity Signer app and Polkadot js apps. Will reopen the issue once the new version of PS is released

Issue #2294, opened 2021-07-06
Wiki-Ops


salmad3:
## Workflows
 - [x] Enhance jobs and configuration files
    - build and deploy.
    - ~Substrate version check: https://github.com/w3f/polkadot-wiki/issues/1815~ Resolved in: https://github.com/w3f/polkadot-wiki/issues/2330
    - jobs should be swift and concise where they can be.
  - [x] Fix CI/CD substrate rustdoc link updater
    - master CICD should only fail if there is a build error
  - [x] Clean-up previous CircleCI implementation
 - [x] Stage non-prod versions off of main branch. 
 - ~[x] Fix Prettier~
 - ~[ ] Run Prettier on external PRs: https://github.com/w3f/polkadot-wiki/issues/944~ Using `pretty-quick` instead 
 - ~[ ] Abstract stalecheck: https://github.com/w3f/polkadot-wiki/issues/1064~
   - ~should also have a stale checker for older issues along with current stale checker for docs, and check for duplicate issues before creating an issue.~ -> instead of this updating in this manner, issues can be managed using a kanban board (`ZenHub` has been setup, can add with https://www.zenhub.com/extension). 
 - [x] updating the wiki results in all wiki pages mentioning the individual that committed recently - should only be those pages that were updated: https://github.com/w3f/polkadot-wiki/issues/2452

## Version 2
- [x] Fix link issue: https://github.com/w3f/polkadot-wiki/issues/2250
   team attempts thus far:
  - play around with config.
  - disable crowdin: https://github.com/w3f/polkadot-wiki/pull/2221
  - disable trailing slash: https://github.com/w3f/polkadot-wiki/pull/2281
  -  referencing update for possible docusaurus compatibility: https://github.com/w3f/polkadot-wiki/pull/2278
- [x] Naming convention
- [x] Injections: https://github.com/w3f/polkadot-wiki/issues/2443
- [x] Conversion tool: https://github.com/w3f/polkadot-wiki/issues/2415
  - configuration & reformatting? 

## Suggestions
  - [x] Create sub-directories for "docs" instead of having all wiki pages under one directory. (i.e.  `learn` has learn guides, `build` has build guides, etc.)
  - [x] Need to clean up repo in general - docker related files, configuration files, duplicates, etc.
  - ~[ ] Doc images could use a name convention and also organized into better sub-directories.~ I would say this can be revisited, but perhaps not a priority atm.
  - [x] Update README
[...]

PR #2308, opened 2021-07-08
[CI] fix Substrate rustdoc link checker

jakehemmerle:
Don't fail building if the latest Substrate version doesn't have rustdoc links for it. Could be more concise (code was pretty wet to start) but master should only fail now when there are real issues.


jakehemmerle:
should fix substrate rustdoc link updater in #2294

scripts/update-substrate-links.mjs
```diff
@@ -10,30 +15,28 @@ import fs from "fs";
...
+    
+    while (!pass) {
+      let testUrl = `http://substrate.dev/rustdocs/${tagTemp}/sc_service/index.html`;
+      try {
+        const res = await axios.get(testUrl);
+        if (res.status === 200) pass = true;
+      } catch (err) {
+        tagTemp = tagTemp.slice(0, -1);
+        if (tagTemp.length === 0) {
+          console.log(`NOTE: '${tag}' and variants of its prefixes are not valid Substrate rustdoc version numbers. No links to update!`);
```
emresurmeli, line 31: We can log the err as well

salmad3 (commented):
tested and LGTM

scripts/update-substrate-links.mjs
jakehemmerle: I don't think it's noteworthy, every error is always a 404 and will pollute the logs with every prefix try. If we hit that console.log, it means we've tried every prefix

PR #2315, opened 2021-07-09
Added Westend WND units

trevor-crypto:


salmad3:
Thank you @trevor-crypto,

Since WND tokens are part of the testnet and hold no economic value,  do you think its better to just add content about WNDs and compare it to the native token units? Could even mention something along the lines of "Westend having the same parameters as Polkadot like session length and governance enactment period."

trevor-crypto:
@DannyS03 Sorry, I'm a bit new to Substrate/Polkadot. Doesn't Kusama also have no economic value? I'm not really sure if it's appropriate to put this Westend info here, but I was looking for it and it wasn't in an accessible location (within source code network configs)

salmad3:
Oh no worries at all. So Kusama does have real economic value. Perhaps this page can offer some meaningful insight: https://wiki.polkadot.network/docs/learn-kusama-vs-polkadot/#:~:text=Although%20they%20share%20many%20parts,conservative%2C%20prioritizing%20stability%20and%20dependability.

How about we add the WND units under the `Obtaining Testnet DOT` section?

trevor-crypto:
> How about we add the WND units under the Obtaining Testnet DOT section?
@DannyS03 That's a good idea. I will push a change.

Thanks for the info

salmad3 (approved):

PR #2337, opened 2021-07-14
Enable Crowdin

nam-hle:
Continue #2213.

Resolve #2333, a part of #2294.

For maintainers, please help me config Github env variables for:

- CROWDIN_PROJECT_ID: it should be "371273"
- CROWDIN_API_TOKEN: generate a new one from your Crowdin profile https://support.crowdin.com/api/api-integration-setup/

PR #2346, opened 2021-07-16
Enable crowdin for polkadot wiki only

nam-hle:
Another attempt after the failure of #2337.


salmad3 (approved):
LGTM

salmad3:
Environment variables are updated now right?

nam-hle:
No. I just changed the configuration of Crowdin Yaml file. I assume that @jakehemmerle set the right value that I gave.

Or can you please check that current CROWDIN_API_TOKEN starts with `b902dd`?

If this PR can not solved the issue, I will have no clues :( It works perfectly in my local environment

salmad3:
@nam-hle reverted as CI didn't pass on master - seems like something with the token

PR #2348, opened 2021-07-16
replace self-referencing link

grenade:
this document currently references itself in a link embedded in some text referencing kusama. possibly because it was at some point copied from another document describing kusama (see: https://guide.kusama.network/docs/kusama-parameters). this change alters the text to make the inverse comparison from a polkadot, rather than kusama perspective.

### DIFF
--- a/docs/maintain-polkadot-parameters.md
+++ b/docs/maintain-polkadot-parameters.md
@@ -12,7 +12,7 @@ at the [chain state](https://polkadot.js.org/apps/#/chainstate/constants) and/or
 ### Periods of common actions and attributes
 
 _NOTE: Polkadot generally runs at &frac14;th the speed of Kusama, except in the time slot duration itself. See
-[Kusama Parameters](https://guide.kusama.network/docs/kusama-parameters) for more details on how Polkadot's parameters
+[Kusama Parameters](kusama-parameters) for more details on how Polkadot's parameters
 differ from Kusama's._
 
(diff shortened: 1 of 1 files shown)


docs/maintain-polkadot-parameters.md
```diff
@@ -11,9 +11,9 @@ at the [chain state](https://polkadot.js.org/apps/#/chainstate/constants) and/or
 
 ### Periods of common actions and attributes
 
-_NOTE: Kusama generally runs 4x as fast as Polkadot, except in the time slot duration itself. See
-[Polkadot Parameters](maintain-polkadot-parameters.md) for more details on how Kusama's parameters
-differ from Polkadot's._
+_NOTE: Polkadot generally runs at &frac14;th the speed of Kusama, except in the time slot duration itself. See
+[Kusama Parameters](https://guide.kusama.network/docs/kusama-parameters) for more details on how Polkadot's parameters
```
salmad3: For the Kusama Parameters page, we can just change it to "kusama-parameters.md" and it will reference the page the same way it referenced "maintain-polkadot-parameters.md"

salmad3 (changes requested):
Thanks @grenade

docs/maintain-polkadot-parameters.md
grenade: ah! thanks.

salmad3 (approved):

PR #2349, opened 2021-07-16
[3] Issue Knock-off 

salmad3:
General review done for pages committed too. For the most part - content on pages is up-to-date, unless there is something specific to add or that was missed. 

However, this intends to close some outstanding stales, where some pages will be updated soon, having addressed other issues currently open and in progress.


jakehemmerle (approved):
lgtm!

PR #2358, opened 2021-07-19
Update learn-accounts.md

### LINKED ISSUE #2270: Add more information about soft vs hard derivation paths
https://wiki.polkadot.network/docs/learn-accounts/#derivation-paths

This should be a layman level understanding. Here was my explanation, but I'm not a cryptographer - I've asked Jeff to comment.

A soft derivation path has two weaknesses - it allows you to "go backwards" to figure out previous private keys IF you know the second private key, and it's possible to determine that different accounts generated from the same seed are linked to that seed. A hard derivation path does not allow either of these - if you know a derived private key, it's not feasible to figure out the initial private key of the root address, and it's impossible to prove that the first account is linked with the second.

So let's say you have Seed A on Parity Signer. This generates a root account (let's call it R). You then do a soft derivation (/1), which generates a new keypair for a new account (let's call it S for Soft) and a hard derivation (//1) which generates a different keypair for another, totally separate account (let's call it H for Hard). You then give the private keys of these accounts away along with their derivation paths. The person who received the private key for S could work backwards to figure out the private key (seed) for R, but not someone who knew H.

If you only show public information, someone who knows the public keys (i.e. anyone who knows the addresses), can determine that R and S are made from the same seed (although they won't know the seed itself), but they will not know that R and H are linked.

DrW3RK:
Added placeholder for content on Soft vs Hard Derivation paths


DrW3RK:
@laboon I made some preliminary updates to the page using your explanation in the issue. Let me know if I need to augment this with screenshots or a video. Also, let me know if you have any feedback on my writing style. Thanks!

docs/learn-accounts.md
```diff
@@ -44,16 +44,41 @@ addresses from the same mnemonic.
...
-keys. See [Substrate BIP39](https://github.com/paritytech/substrate-bip39) for more information.
+rounds of PBKDF2 on the mnemonic. As such, the same mnemonic may generate different private keys on different wallets, due to the difference in the cryptographic algorithms used. See [Substrate BIP39](https://github.com/paritytech/substrate-bip39) for more information.
 
 ### Derivation Paths
 
-Many Polkadot key generation tools support hard and soft derivation. See the
+If you would like to create and manage several accounts on the network using the same mnemonic seed phrase, you can use derivation paths. We can think of the derived accounts as the child accounts of the root account created using the original mnemonic seed phrase. Many Polkadot key generation tools support hard and soft derivation. For instance, if you intend to create an account to be used on the Polkadot chain, you can derive a **hard key** child account using **//** after the mnemonic phrase.
+
+```js
+"caution juice atom organ advance problem want pledge someone senior holiday very//polkadot//0"
```
laboon:
```suggestion
"caution juice atom organ advance problem want pledge someone senior holiday very//0"
```
The "polkadot" isn't necessary, but is added by default by Parity Signer and probably some other wallets.

docs/learn-accounts.md
```diff
@@ -44,16 +44,41 @@ addresses from the same mnemonic.
...
+```
+
+and a **soft key** child account using **/** after the mnemonic phrase
+
+```js
+"caution juice atom organ advance problem want pledge someone senior holiday very/polkadot/0"
```
laboon:
```suggestion
"caution juice atom organ advance problem want pledge someone senior holiday very/0"
```

docs/learn-accounts.md
```diff
@@ -44,16 +44,41 @@ addresses from the same mnemonic.
...
+```
+If you would like to create another account for usage on the Polkadot chain, you can change the number at the end of the string above from 0 to 1. You can use any words or numbers in the derivation path as long as they make sense to you. A soft derivation allows you to "go backwards" to figure out previous private keys IF you know the second private key, and it's possible to determine that different accounts generated from the same seed are linked to that seed. A hard derivation path does not allow either of these - if you know a derived private key, it's not feasible to figure out the initial private key of the root address, and it's impossible to prove that the first account is linked with the second. Both of these derivation methods have their use cases, given the private keys for all the derived accounts are fully secure. If it is likely that you may use a derived account in a less secure setting, it is recommended to generate the account using a hard derivation path.
```
laboon:
```suggestion
If you would like to create another account for usage on the Polkadot chain using the same seed, you can change the number at the end of the string above. For example, `/1`, `/2`, and `/3` will all create different derived accounts.

You can use any letters or numbers in the derivation path as long as they make sense to you; they do not have to follow any specific pattern. You may also combine multiple derivations in your path, as well. For instance, `//bill//account//1` and `//john/polkadot/initial` are both valid. You will need to know both the seed and the derivation path in order to recreate a derived account, so you should either use a well-defined sequence (e.g. //0, //1, //2...) or be sure to write down any derivation paths you use.

> It is not possible to generate a derived account without also knowing the derivation path.

### Soft vs. Hard Derivation

A soft derivation allows someone to potentially "go backwards" to figure out the private key of the initial account IF they know the private key of the derived account. It is also possible to determine that different accounts generated from the same seed are linked to that seed. A hard derivation path does not allow either of these - even if you know a derived private key, it's not feasible to figure out the private key of the root address, and it's impossible to prove that the first account is linked with the second. Both of these derivation methods have their use cases, given the private keys for all the derived accounts are fully secure. Unless you have a specific need for a soft derivation, it is recommended to generate the account using a hard derivation path.
```

laboon: We should also put the Soft vs. Hard description further up above (before we mention using / and //)

docs/learn-accounts.md
```diff
@@ -44,16 +44,41 @@ addresses from the same mnemonic.
...
 entropy byte array to generate the private key, while full BIP39 wallets (like Ledger) use 2048
-rounds of PBKDF2 on the mnemonic. As such, the same mnemonic will not generate the same private
-keys. See [Substrate BIP39](https://github.com/paritytech/substrate-bip39) for more information.
+rounds of PBKDF2 on the mnemonic. As such, the same mnemonic may generate different private keys on different wallets, due to the difference in the cryptographic algorithms used. See [Substrate BIP39](https://github.com/paritytech/substrate-bip39) for more information.
 
 ### Derivation Paths
 
-Many Polkadot key generation tools support hard and soft derivation. See the
+If you would like to create and manage several accounts on the network using the same mnemonic seed phrase, you can use derivation paths. We can think of the derived accounts as the child accounts of the root account created using the original mnemonic seed phrase. Many Polkadot key generation tools support hard and soft derivation. For instance, if you intend to create an account to be used on the Polkadot chain, you can derive a **hard key** child account using **//** after the mnemonic phrase.
```
laboon:
```suggestion
If you would like to create and manage several accounts on the network using the same seed, you can use derivation paths. We can think of the derived accounts as child accounts of the root account created using the original mnemonic seed phrase. Many Polkadot key generation tools support hard and soft derivation. For instance, if you intend to create an account to be used on the Polkadot chain, you can derive a **hard key** child account using **//** after the mnemonic phrase.
```

laboon (changes requested):

docs/learn-accounts.md
```diff
@@ -44,16 +44,41 @@ addresses from the same mnemonic.
 
 Most wallets generate a mnemonic phrase for users to back up their wallets and generate a private
 key from the mnemonic. Not all wallets use the same algorithm to convert from mnemonic phrase to
-private key.
+private key. A typical mnemonic seed phrase generated by [the Subkey tool](https://substrate.dev/docs/en/knowledgebase/integrate/subkey) is shown below
```
laboon:
```suggestion
private key. A typical mnemonic phrase generated by [the Subkey tool](https://substrate.dev/docs/en/knowledgebase/integrate/subkey) is shown below
```

PR #2363, opened 2021-07-20
Update simple payouts and prettier

laboon:
The only real changes here are in learn-simple-payouts, the rest are just prettier fluff, but I figure let's just put it in now.

Right now the wiki notes that you have 84 eras to claim. This is true UNLESS a validator unbonds and then kills a stash without issuing a payout which gives users 28 eras to claim (during the unbonding period). This has caused problems for people so I recommend we update the wiki with this ASAP.


jakehemmerle (approved):
lgtm!

if anyone is curious as to why validators aren't forced to check and payout before they unbond, shawn explains why here (bill found this link): https://github.com/paritytech/substrate/pull/9388

Issue #2173, opened 2021-06-16
Add staking section on new `chill_other` extrinsic and governance set validator & nominator bounds


emostov:
https://github.com/paritytech/substrate/pull/8920 adds

- `Min{Validator, Nominator}Bond`, the minimum amount to bond in order to stake, as set by governance
- `chill_other`, a permissionless extrinsic allowing anyone chill a {validator, nominator} below `Min{Validator, Nominator}Bond`
- `Max{Validator, Nominator}Count`, a hard cap on the number of stakers in the system

The rational behind these changes is to ensure we don't have unbounded growth to the input of the phragmen elections, which are very compute (NP Hard) and memory intensive for a redundant, distributed system. Worth noting that these changes are planned to only be used in the short term; longer term strategies are being investigated to allow minimally bounded growth to staking participation.

cc @wpank

PR #2375, opened 2021-07-22
Update maintain-guides-how-to-chill.md

DrW3RK:
Summarize the contents in https://github.com/paritytech/substrate/pull/8920

https://github.com/w3f/polkadot-wiki/issues/2173


docs/maintain-guides-how-to-chill.md
```diff
@@ -59,7 +59,7 @@ order to start nominating you again.
 
 ## Chill Other
 
-(placeholder for content)
+Due to constraints in the runtime, an unbounded and unlimited number of nominators and validators is just not possible. Multiple checks are incorporated to keep the size of staking system manageable, like mandating minimum active bond requirements for both nominators and validators. When these requirements are modified through on-chain governance, they can be enforced only on the accounts that newly call `nominate` or `validate` after the update. The changes to the parameters would not automatically chill the active accounts on chain that do not meet the bonding requirements. `chill_other` extrinsic was incorporated to keep things backwards compatible and safe. `chill_other` extrinsic is permissionless and any third party user can target it on an account where the minimum active bond is not satisfied, and chill that account.
```
DannyS03, line 62: - Perhaps we can make it flow even better - something along the lines of: 
`An unlimited number of unbounded actors in Polkadot's NPoS is just not possible due to constraints in the runtime`....`As a result, multiple checks are ...` ... and so on 

- `The changes to the parameters would not automatically chill the active accounts on chain that do not meet the bonding requirements. ` 
  - "on-chain"
  - user might wonder, "which parameters exactly?" 
- `'chill_other' extrinsic was incorporated to keep things backwards compatible and safe.` & `'chill_other' extrinsic is permissionless and any third party user can target it on an account where the minimum active bond is not satisfied, and chill that account. `
  - Just adding "The" before 'chill_other'

DrW3RK:
Thank you @DannyS03 for the preliminary review and feedback. I updated the content with your suggestions. If it is good to go, please merge it to Master. Thanks!

salmad3 (commented):
- Perhaps we can make it flow even better - something along the lines of: 
`An unlimited number of unbounded actors in Polkadot's NPoS is just not possible due to constraints in the runtime`....`As a result, multiple checks are ...` ... and so on 

- `The changes to the parameters would not automatically chill the active accounts on chain that do not meet the bonding requirements. ` 
  - "on-chain"
  - user might wonder, "which parameters exactly?" 
- `'chill_other' extrinsic was incorporated to keep things backwards compatible and safe.` & `'chill_other' extrinsic is permissionless and any third party user can target it on an account where the minimum active bond is not satisfied, and chill that account. `
  - Just adding "The" before 'chill_other'

salmad3 (approved):

PR #2403, opened 2021-07-29
Deprecate Fishermen

### LINKED ISSUE #2396: Deprecate concept of fishermen
@burdges suggested that there is currently no sensible way (and also no necessity) to implement fishermen.

As far as I'm aware (though I definitely could be missing something) this concept is outdated and can be deprecated in the documentation, possibly in more places than just the wiki. Since the concept is well-known in the ecosystem and has been thematised since the whitepaper, we probably shouldn't just "forget about it" but explain why polkadot is already secure without fishermen etc.

It would be good to get confirmation from someone "inside" like @bkchr @shawntabrizi.

salmad3:
Resolves #2396

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -176,11 +176,6 @@ Chain uses a deterministic finality gadget known as [GRANDPA](#grandpa-finality-
 
 A mechanism that determines finality.
 
-## Fisherman
-
-Nodes that monitor the network for validators or collators who are behaving badly. Fishermen must
-stake a small amount of DOT but can be rewarded greatly if they find bad behavior.
-
 ## Frame
 
 The collection of Substrate-provided pallets (Substrate Runtime Modules).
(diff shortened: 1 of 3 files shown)


DrW3RK (dismissed):
There are plans to add fisherman into the ecosystem in the future - ref: Polkadot overview paper - https://arxiv.org/pdf/2005.13456.pdf

Gav's interaction last week - 
"The simple answer is that the chance of all validators which are randomly assigned to a parachain (and rotated often) all being malicious in a coordinated way is so low as to be within acceptable security bounds. Assuming 10 primary validators and 10 backing validators (hopefully I'm getting the latest vocabulary right), then you have a 1/3^(10+10) chance of them all being malicious given a maximum of 1/3 malicious validators. This is 0.0000000002867972, which assuming one sample each 6 second block, is something we would expect to happen approximately once every 663 years. (This assumes that there are no external fishermen, which is something that could be added later, and also that we resample each block - we actually resample on longer periods, multiplying up the expectation even further)"

DrW3RK:
After interacting with research team, I confirm that your edit accurately reflects the current stance of Web3 Foundation regarding the implementation of Fishermen. Please go ahead and merge your edit with Master

PR #2411, opened 2021-07-30
Add os tabs, Update maintain-sync

### LINKED ISSUE #1866: Consider os-specific tabs in wiki (and tabs in general)
Instead of os sections being separated one after another (such as https://wiki.polkadot.network/docs/en/maintain-sync#fast-install-instructions-mac), tabs could be used. The site could also recognize the user's os and pre-set the default.  

e.g.
### Fast Install Instructions
Mac || Windows || Linux

salmad3:
Resolves #1866

Seems like this is the only page with os-specific instructions, unless others would benefit from it. Also not sure if injection will work inside the the tab element: hopefully staging will show this. When ran locally, the injections don't seem to work. 

Tabs should look like:
image


jakehemmerle (changes requested):
remove package-lock.json ; you must have accidentally run `npm install` or related

PR #2422, opened 2021-08-03
Revert "Update learn-accounts.md"

DrW3RK:
Reverts w3f/polkadot-wiki#2416


DrW3RK:
As redirects only work after deploying them, I had to merge these commits to master. Reverting because the updates could not work

PR #2442, opened 2021-08-09
Clean up and Fixes for Restructure

salmad3:
Feel free to commit to this branch 
- ~link redirects (backwards compatibility) -> endpoints should redirect to `/docs/`~
- ~noticed injections aren't working (even on prod)~

Going to merge this as updates to the original restructure to not stop other updates on master. Still figuring out redirects and the injection issue (redirects from before don't seem to be recognized properly either).


nam-hle:
HI @DannyS03, sorry I can't follow the project for a while. 
Please help me remove the condition in the path: `pokadot-wiki/docusaurus.config.js -> presets -> remarkPlugins`. It should become: `remarkPlugins: injectPlugin({ isPolkadot: true })`, so the injection always works regardless other factors.

salmad3:
> HI @DannyS03, sorry I can't follow the project for a while.
> Please help me remove the condition in the path: `pokadot-wiki/docusaurus.config.js -> presets -> remarkPlugins`. It should become: `remarkPlugins: injectPlugin({ isPolkadot: true })`, so the injection always works regardless other factors.

Hey:) no worries! Thanks for checking in - I updated the presets. It seems it is not recognizing some of the values in inject-dict

`isPolkadot`: 
image


We also wanted to redirect all the urls as they were before, to `/docs/` - we tried to organize the docs in subfolders and their endpoints are `/docs/general/<doc-name>/`, `/docs/learn/<doc-name>/`, etc... but we would like to redirect all these to `/docs/<doc-name>`.

I have been trying with the `"@docusaurus/plugin-client-redirects"` but no luck yet

nam-hle:
> We also wanted to redirect all the urls as they were before, to `/docs/` - we tried to organize the docs in subfolders and their endpoints are `/docs/general/<doc-name>/`, `/docs/learn/<doc-name>/`, etc... but we would like to redirect all these to `/docs/<doc-name>`.
> 
> I have been trying with the `"@docusaurus/plugin-client-redirects"` but no luck yet

Sorry, I don't get it. I saw that the doc structure is split into subfolders (learn, build...), and current URLs have the format: `/docs/learn-architecture`. So what URLs do you want to redirect from and to?

PR #2463, opened 2021-08-12
Update CI for Staging to Deploy on PRs

salmad3:
PR being used as test

It's helpful to have the staging site updated when a change is requested so one is able to see their changes and make others as needed before it gets merged to master - master can serve as QA check before prod


DrW3RK:
Good idea! So, this workflow makes testing locally obsolete right? You make commits to your PR and you see them reflected in staging?

PR #2486, opened 2021-08-17
multiple redirects warning

DrW3RK:
Although the redirects work, the warning message shows up when building and deploying


salmad3:
@DrW3RK is this just a placeholder?

DrW3RK:
Yeah. Trying to understand why it throws the warnings even though the redirects are successfully incorporated

DrW3RK:
Could not identify the issue. Redirects work as expected

PR #2508, opened 2021-08-24
Update Wiki Pages to Address Issues

salmad3:
- For the thousand validators page, strangely enough adding "docs" to the doc `id` fixes the sidebar issue.


DrW3RK:
The edits look good. Might need a deeper look at why thousand validators page behaved that way. This interim fix can be rolled out

PR #2522, opened 2021-08-26
Deprecate old guide and assets

lamafab:
Deprecate the old guide, since it seems to confuse (new) people. Apparently they miss the warning:

image

One could consider adding a new guide for the new service, but I'd argue that's not really necessary since the website itself tells you what to do and new comers do not have any problems with the process itself. Only people who are familiar with the old service keep sending signatures of the challenges rather than just the challenge itself, or expect the service to initialize first contact.

Any thoughts?


laboon:
Let's get rid of it, yeah. @DannyS03 why is the Staging failing?

PR #2564, opened 2021-09-08
Correct `alertmanager` binary extract path

azerella:
We're trying to move a binary here from the compressed archive rather than the extracted folder output


salmad3 (approved):
Thanks @azerella

PR #2571, opened 2021-09-10
Fix wrong DOT value in tx construction page

GCrispino:
Hello! I couldn't find any more info about this apart from the ["What is DOT?"](https://wiki.polkadot.network/docs/learn-DOT#polkadot) page, so please let me know if I'm mistaken.

This changes a command line example available on the page, that shows how to create a raw transfer transaction. More specifically, this changes the amount to be transferred. The page mentions that 0.5 DOT will be transferred, but the value passed to the `submit` command is of `500000000000` (plancks?), which according to the "What is DOT?" page, corresponds to 50 DOT.
Because of this, this PR replaces the value `500000000000` by `5000000000` in the text.


salmad3 (approved):
Thanks @GCrispino

GCrispino:
> Thanks @GCrispino 

No problem @DannyS03 :)

PR #2572, opened 2021-09-11
Fixed typos and grammatical errors in the polkadot consensus wiki page

zeim839:
**Typos:**
utiliies => utilizes
progres => progress

**Progress progresses?**
Its self-explanatory and redundant to say `... as progress progresses`. 
`As progress ensues` would be a better-sounding alternative.


salmad3 (approved):
Thanks @zeim839

PR #2587, opened 2021-09-14
fixing image file names

DrW3RK:
The fix for image rendering issue may be much simpler than what we thought earlier


DrW3RK:
Nope. That did not work!

PR #2590, opened 2021-09-14
Add Re-Genesis to Glossary

### LINKED ISSUE #1949: Add re-genesis information
See:
https://github.com/polkadot-js/extension/issues/687#issuecomment-811981857
https://github.com/polkadot-js/extension/issues/687

We should include in the wiki:

1. An explanation of what re-genesis is in the glossary
2. Perhaps some steps on how it is done

This will most likely involve having one of the core devs review our work very carefully, so we should probably wait until post-parachains to do this (or maybe split this issue up and just add the term to the glossary for now)

DrW3RK:
As the Re-Genesis issue is still open - https://github.com/paritytech/substrate/issues/7458 I am adding the term to the Glossary on Wiki. Will expand this to a dedicated Wiki page once the process gets formalized on Substrate.

### DIFF
--- a/docs/general/glossary.md
+++ b/docs/general/glossary.md
@@ -457,6 +457,10 @@ the Governance Council, by a member of the public, or as the result of a previou
 Stakeholders vote on referenda, weighted by both the size of their stake (i.e. number of DOT held)
 and the amount of time they are willing to lock their tokens.
 
+## Re-Genesis
+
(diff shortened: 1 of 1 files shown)


salmad3 (commented):
@DrW3RK We can add a subsection that briefly introduces Re-Genesis (similar to https://wiki.polkadot.network/docs/learn-consensus#badass-babe-sassafras), by describing some of the rationale and design, and have more of a swift description on the glossary page, such as: 
```
Re-Genesis is a process of exporting the current chain state and creating a new chain that builds on it. Re-Genesis is still being developed.
``` 

Then, we can update the subsection when developments are made.

PR #2609, opened 2021-09-17
Update Anonymous Proxy with Note about type `Any`

### LINKED ISSUE #2594: Update Anonymous Proxies article
There should be a warning in the anonymous proxy section that tells users the first proxy they add should always be of type "Any". Also, if there are multiple proxies for the proxied account, they shouldn't remove all `Any` types unless they're ready to kill the relationship.

Without an `Any` type they won't be able to send funds, add new proxies or kill the anonymous proxy or take any action not specifically allowed by the types of the proxies the account has.

salmad3:


DrW3RK:
@michalisFr let us know if this addresses the issue

docs/learn/learn-proxies.md
```diff
@@ -105,6 +105,13 @@ multisig would be created.
 
 ## How to set up a Proxy
 
+> NOTE: The first proxy you add should always be of _type_ `Any`.
```
michalisFr: This warning applies only to Anonymous proxies, so it should go under `Creating Anonymous Proxies`.

salmad3: Updated

salmad3:
Merging after addressing feedback.

PR #2610, opened 2021-09-17
Update `set-value` Version 

salmad3:
To address Dependbot Alert


DrW3RK:
Not sure about this one! I guess you know what you are doing. Looks like this will reduce the number of alerts? Can revert if any issues arise.

PR #2622, opened 2021-09-21
Update learn-proxies.md

DrW3RK:
Clarifying the permissions of anonymous proxies

### DIFF
--- a/docs/learn/learn-proxies.md
+++ b/docs/learn/learn-proxies.md
@@ -123,13 +123,13 @@ selected account at the top is the account that will be the primary account.
 > empty `enum`, and if you try to add this in as a proxy, nothing will happen. No new proxy will be
 > created.
 
-**Please be aware of how critical it is to set up an anonymous proxy and the overall dangers.**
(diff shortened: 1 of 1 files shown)


docs/learn/learn-proxies.md
```diff
@@ -123,13 +123,13 @@ selected account at the top is the account that will be the primary account.
 > empty `enum`, and if you try to add this in as a proxy, nothing will happen. No new proxy will be
 > created.
 
-**Please be aware of how critical it is to set up an anonymous proxy and the overall dangers.**
+**It is critical to setup Anonymous Proxies with appropriate permissions**
```
salmad3: If rewording, I would keep the "warning" attribute of the statement, as there is still a risk involved in setting up and using anonymous proxies even with the right setup.

salmad3 (changes requested):

docs/learn/learn-proxies.md
profradha: Added. It also makes sense to expand on the dangers involved by giving few scenarios - like lost deposit

salmad3 (approved):

Issue #2597, opened 2021-09-16
Kusama Guide Revamp


salmad3:
https://github.com/w3f/polkadot-wiki/issues/1919#issuecomment-905867464

Issue will now focus on revamping the Kusama guide in general.

DannyS03:
@jakehemmerle 
```
2021-09-17 03:42:20 (28.4 MB/s) - ‘ipfs-cluster-ctl_v0.14.0_linux-amd64.tar.gz’ saved [11802521/11802521]

/home/runner/work/polkadot-wiki/polkadot-wiki/ipfs-cluster-ctl/ipfs-cluster-ctl
2021-09-17T03:42:21.798Z	ERROR	apiclient	client/lbclient.go:105	reached maximum number of retries without success, retries: 1
(node:1946) UnhandledPromiseRejectionWarning: Error: Command failed: ipfs-cluster-ctl     --enc json     --basic-auth ***:***     --host /dns4/hemmerle.dev/tcp/9097/p2p/12D3KooWBxH7hKkvvp2oCDay9jSh8gw5FkdBsdgPytXFjRgqDvf8     status
2021-09-17T03:42:21.798Z	ERROR	apiclient	client/lbclient.go:105	reached maximum number of retries without success, retries: 1

    at checkExecSyncError (child_process.js:760:11)
    at execSync (child_process.js:833:15)
    at runCommandOnCluster (/home/runner/work/polkadot-wiki/polkadot-wiki/scripts/ipfs-cluster-pin.js:30:50)
    at getCidByPinName (/home/runner/work/polkadot-wiki/polkadot-wiki/scripts/ipfs-cluster-pin.js:40:35)
    at main (/home/runner/work/polkadot-wiki/polkadot-wiki/scripts/ipfs-cluster-pin.js:46:25)
    at Object.<anonymous> (/home/runner/work/polkadot-wiki/polkadot-wiki/scripts/ipfs-cluster-pin.js:62:3)
    at Module._compile (internal/modules/cjs/loader.js:1072:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
    at Module.load (internal/modules/cjs/loader.js:937:32)
    at Function.Module._load (internal/modules/cjs/loader.js:778:12)
(Use `node --trace-warnings ...` to show where the warning was created)
[...]

PR #2625, opened 2021-09-22
[1] Kusama Guide Revamp: Landing Page

salmad3:
### Main Highlights
- Landing page shows a Kusama-based wallpaper. While it is nice, the **UX should be improved** and match the utility of the Polkadot Wiki landing page.
- A **reference to the Polkadot Wiki** (e.g. *Kusama's cousin, Polkadot*), and a **Contribution Section** (Help Improve this Wiki) instead of having a contribution bar always living at the top of the site.  
- General improvements

The focus here is to create more utility and flow on the page (1), where additional UI/UX improvements (2) can be added after.
  - This would also welcome the opportunity to create brand assets that the Wiki could benefit from. At the moment, I have used some existing assets like the gif that exists on the https://kusama.network/ website. 

Addresses #2597


salmad3:
image

Will include more sections once the sections on the actual guide are updated.

salmad3:
image

This approach still accounts for the uniqueness and theme of the Kusama guide, instead of mimicking the cards on the Polkadot Wiki main page: 

image

as the main page also reference the Kusama guide since it stands as the "Main Wiki", where the Kusama guide is seen as an extension.

image

PR #2643, opened 2021-09-25
Fix XCVM. Cross-Chain to Cross-Consensus

nkmr-jp:
Hello!
The XCVM description was wrong and has been corrected.

[XCM: The Cross-Consensus Message Format | by Gavin Wood | Polkadot Network | Sep, 2021 | Medium](https://medium.com/polkadot-network/xcm-the-cross-consensus-message-format-3b77b1373392)
> At the core of the XCM format lies the XCVM. Contrary to how it might look to some, this is not a (valid) roman numeral (though if it were, it’d probably mean 905). In fact this stands for Cross-Consensus Virtual Machine.

✗ Cross-Chain Virtual Machine
○ Cross-Consensus Virtual Machine


salmad3 (approved):
Thanks @nkmr-jp

PR #2646, opened 2021-09-27
Address Dependabot Alerts

salmad3:
Addresses the latest dependency alerts by updating versions for `ansi-regex` & `nth-check`.


DannyS03:
Merge is pending a build fix, but review can happen.

PR #2658, opened 2021-09-27
Nginx Formatting and Chain Name fixes

Todo56:
- Fixed nginx formatting so it no longer throws an error.
- Fixed chain name


salmad3 (approved):
Thanks @Todo56, LGTM

PR #2676, opened 2021-09-30
fix `kusama:build` and injection issues

jakehemmerle:
This should fix `kusama:build`, and the injection issues in both kusama-guide and polkadot-wiki

- adds injection back into `polkadot-wiki/docusaurus.config.js`. Relevant line is on line 60, ignore all the linting.
- downgrades `unist-util-visit` to older version. We can upgrade if/when we decide to migrate all the JS files from CommonJS to ESM.


salmad3 (approved):
Thanks! Tested it locally, LGTM

PR #2696, opened 2021-10-06
Phragmms algorithm fix

a-jwc:
Includes small fixes that weren't included in the original PR.


salmad3 (changes requested):
- We should keep the `*` instead of `_` for italicizing
- Needs pretty (mainly the `Phragmms (fka Balphragmms)` section)

salmad3 (approved):
LGTM

PR #2699, opened 2021-10-07
Add AirGap Wallet

AndreasGassmann:
AirGap is a secure mobile wallet and has been supported by the Web3Foundation to integrate DOT. (Not through the treasury). We are actively working on both the apps in general and also the DOT integration.

The AirGap Vault application is an offline application similar to the Parity Signer.
The AirGap Wallet application is a companion app to the Vault, which is used to prepare and broadcast transactions, but does not hold any private keys, similar to the polkadot-js apps.

If you need any other information to merge this PR, please let me know.

### DIFF
--- a/docs/build/build-wallets.md
+++ b/docs/build/build-wallets.md
@@ -16,10 +16,12 @@ development statuses.
 
 | Wallet Name                                                         | Development State | Team Name | Description       | Custody       | Supports |
 | ------------------------------------------------------------------- | ----------------- | --------- | ----------------- | ------------- | -------- |
-| [Parity Signer](https://www.parity.io/signer/) (see warning below)  | Live              | Parity    | iOS and Android   | Non-custodial | Staking  |
-| [Polkadot-js Desktop](https://github.com/polkadot-js/apps/releases) | Live              | Parity    | Win, Mac, Linux   | Non-custodial | Staking  |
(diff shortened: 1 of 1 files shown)


salmad3:
Thanks for the PR @AndreasGassmann. The supported wallets are wallets that we can offer official support for (e.g. through Polkadot Support) as they aren't directly considered to be "third-party" and are *native* to the ecosystem.

With other wallets, we have limited these to ones that are Treasury funded, which may be revisited in the future, however, there is an ongoing responsibility to keep the content neutral as there are many third-party applications.

AndreasGassmann:
The Polkadot integration in our wallet was funded by the Web3Foundation, but it did not go through the treasury (I was not involved in this process, so I don't know the specifics).

Is there a place where wallets like ours are listed in the polkadot wiki?

---

After hearing your explanations, I would suggest to change the title from "Supported Wallets" to "Official Wallets". Because the way the title is now, I would assume that "Supported Wallets" refers to "wallets that are financially supported by the Web3Foundation", which is where our wallet would fit in as well.

Furthermore, if the "Supported Wallets" are only Parity Signer and polkadot.js, then this text is redundant and can be removed, as it suggests there are other "Supported Wallets" where Polkadot support does not provide assistance:
> Polkadot support can provide assistance with issues related to Polkadot-JS, the Polkadot{.js} extension, or Parity Signer. For other wallet software, you should contact the developers of that wallet.

Instead you could write something like:

> Polkadot support can provide assistance with issues related to the wallets listed above. For other wallet software, you should contact the developers of that wallet.

PR #2710, opened 2021-10-11
Update Auction Content to Include Breakdown of Randomness

### LINKED ISSUE #2674: Add more details to candle auction (learn-auction)
https://wiki.polkadot.network/docs/learn-auction#mechanics-of-a-candle-auction

Discuss how randomness is retrieved, sampling of blocks, etc. From Rob H's description in the Kusama Direction channel -

 Here's the event where the auction determined that the winning sample during the ending period had index '1078' . https://kusama.subscan.io/extrinsic/9434278-0?event=9434278-0 . The parent block was a new BABE session in the 'Logs', which updated the randomness which was used to select that sample index: https://kusama.subscan.io/block/9434277
With an archive node, you'd be able to inspect the state at the end of block 9434278 and see the sample indices
rphmeier
ramimali more info.

https://kusama.subscan.io/extrinsic/0x7b67d653c9522b623a97e20a967b83a8517fe3821370475ddb6611cd37c29a03?event=9335014-26 is the event where Auction 9 started. It was set to end at the latest at block 9362014 + 72000. 72000 is the "ending period" which is divided into 3600 samples of 20 blocks. Sample 1078 refers to the winner as of block 9362014 + 21560, so block 9383574: https://kusama.subscan.io/block/9383574

parachains.info/auctions has a nice display of this information.
image.png
I've zoomed in the chart from Sep 24 07:00 to Sep 29 08:00 which is approximately the entire ending period.
As for actually drawing out the BABE randomness that was used to draw the sample number, that is more involved and requires extracting the BABE randomness from the blockchain.
But in principle, let's say that the auction ended one block after some BABE epoch 'E' began. The digest in the 'Logs' of 9434277 is decodable and contains the random value as well as the BABE authorities.

salmad3:
resolves #2674

### DIFF
--- a/docs/learn/learn-auction.md
+++ b/docs/learn/learn-auction.md
@@ -33,12 +33,34 @@ that point would win.
 When candle auctions are used online, they require a random number to decide the moment of
 termination.
 
-Parachain slot auctions will differ slightly from a normal candle auction in that it does not use
-the random number to decide the duration of its opening phase. Instead, it has a known open phase
+Parachain slot auctions differ slightly from a normal candle auction in that it does not use
+the random number to decide the duration of its opening phase. Instead, it has a *known open phase*
 and will be retroactively determined (at the normal close) to have ended at some point in the past
 during the ending phase. So during the open phase, bids will continue to be accepted, but later bids
 have higher probability of losing since the retroactively determined close moment may be found to
 have preceded the time that a bid was submitted.
+
+### [Randomness](learn-randomness.md) in Action
+
+The following example will showcase the randomness mechanics of the candle auction for 
+the ninth auction on Kusama. Keep in mind that the candle phase has a uniform termination 
+profile and has an equal probability of ending at any given block, and the termination block 
+cannot be predicted before or during the auction.
+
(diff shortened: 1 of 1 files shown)


docs/learn/learn-auction.md
```diff
@@ -33,13 +33,35 @@ that point would win.
...
 
+### [Randomness](learn-randomness.md) in Action
+
+The following example will showcase the randomness mechanics of the candle auction for 
+the ninth auction on Kusama. Keep in mind that the candle phase has a uniform termination 
+profile and has an equal probability of ending at any given block, and the termination block 
+cannot be predicted before or during the auction.
+
+- Auction 9 starts at [`block 9362014`](https://kusama.subscan.io/extrinsic/0x7b67d653c9522b623a97e20a967b83a8517fe3821370475ddb6611cd37c29a03?event=9335014-26).
+  > The auction has a full duration equal to `block 9362014` + `72000`.
```
emresurmeli, line 51: There is good visualizations from Jonas's townhall presentation here: https://hackmd.io/@8F4MrJhQT32fynUEzuSsHA/H1NKcDAJt

emresurmeli: We can use some of them like the candle auction timeline viz

salmad3: Added references to resource section

PR #2711, opened 2021-10-11
Fix Runtime Link Lookup on Maintain Errors Page

### LINKED ISSUE #2687: Fix link for looking up errors in Polkadot
https://wiki.polkadot.network/docs/maintain-errors#polkascan-and-subscan

_By looking at the runtime code we see that the pallet at index 5 is Balances. Now we will check the Balances pallet's code which is hosted in the Substrate repository, and look for the fourth error in the Error enum. According to its source the error that we got is InsufficientBalance or in other words "Balance too low to send value"._

The runtime code does not link to the correct location (list of pallets).

We should link to a permanent location instead of a line number because any update to the code would change the line number.

salmad3:
resolves #2687


emresurmeli (commented):
lgtm

PR #2749, opened 2021-10-20
Remove Spacing Between Conditional Rendering

xnorly:
Removed spaces that were introduced by newline or space characters in between conditionals


salmad3:
Thanks @xnorly, we actually went ahead and updated the associated pages: https://github.com/w3f/polkadot-wiki/pull/2746

PR #2752, opened 2021-10-20
What can I do with my DOT?

### LINKED ISSUE #2702: New article request - What can I do with my DOT?
Shawn suggested an article "What can I do with my DOT?" that covers some tiers of DOT holdings, like:
- 1 DOT: ED
- 5 DOT: Basic TXs, Remarks
- 10 DOT: Democracy Voting
- 20 DOT: Register an Identity
- 100 DOT: Participate in Staking

This will allow new users to get a feeling of how DOT can be used within the ecosystem and introduce them to these terms through a short description and links to the relevant pages. It will also be linked to on the Gift UI after they've claimed their gift.

Probably should go under "Getting started"

DrW3RK:


DrW3RK:
@michalisFr Does this look good?

michalisFr:
That looks good. Any other tiers we can think of or additional things you can do? 

For example, we could add that with 20 DOT you can vote for Council. It's missing something with regard to democracy. Perhaps we can also add that to the 5 DOT tier (vote on referendums) or add a 10 DOT tier (not that there's a minimum but in the sense of a "meaningful" vote).

We could add another for 100 DOT: "You can bond at least this amount and submit a spending proposal to the Council receive funding for building on Polkadot"

Would it make sense to add 5000 DOT as: "You can set up a validator and participate in the Thousand Validators Programme"?

PR #2755, opened 2021-10-20
Update unbonding and rebonding page

### LINKED ISSUE #2753: Outdated information on Rebonding page
Community member noticed that the wiki hasn't been updated to reflect the change in UI when rebonding:(https://wiki.polkadot.network/docs/maintain-guides-how-to-unbond#rebonding-before-the-end-of-the-unbonding-period)

The image still shows the units in DOT when in fact you now need to enter the value in plancks.
unit

It also doesn't mention that you need to select the controller account when rebonding.

Is it possible to update this page to add this information?

a-jwc:
Fixes #2753

Will update with DOT units convertor in next PR.
Also, contains adding *.bak to .gitignore.


salmad3 (approved):
Thanks @a-jwc, lgtm

PR #2786, opened 2021-11-01
replace rebond instructs in plancks with DOT units

a-jwc:
Reverted back to previous instructions with minor clarification update.


salmad3 (approved):
lgtm

PR #2805, opened 2021-11-08
Update learn-staking.md

kl3eo:
I'm going to translate this help for the benefit of the community. This is an initial commit and I hope it will merge.


salmad3:
Hi @kl3eo, thanks for providing this translation. We actually have a method in place for translations that is explained [here](https://wiki.polkadot.network/docs/contributing#helping-with-translations). However, we would advise holding off on this for now as we plan to completely update and restructure the entire wiki. I will close this PR as a result.

PR #2832, opened 2021-11-24
Mark ECC memory a requirement for validator setup

sorpaas:
Single-bit flip can happen and [does happen](https://www.youtube.com/watch?v=aT7mnSstKGs) (a lot) in reality. Those who think otherwise either don't check their hardware error logs, or are not using ECC at all. For a secure validator setup, ECC is a must.


salmad3 (approved):
Thanks @sorpaas :)

PR #2837, opened 2021-11-26
Grammar checks for XCM

albertov19:
Some grammar checks that encountered while reading the XCM article :)


salmad3 (approved):
Thanks @albertov19 :)

PR #2857, opened 2021-12-05
Add prometheus expose command in doc and update the docker image vers…

charly37:
Add more info about exposing service especially Prometheus. I think when this doc was done prometheus exporter was not yet part of polkadot. It took me sometime to understand why i was not able to access the prometheus metric after i deploy on Kubernetes cluster. I hope this will help other users.


salmad3 (approved):
Thanks @charly37 :)

PR #2886, opened 2021-12-16
Missing `--public` parameter on public key "inspection"

davxy:
Without the `public` parameter input was mistakenly interpreted as the `Secret Key URI`.


salmad3 (approved):
Thank you @davxy

PR #2908, opened 2021-12-22
Update kusama-statemine-ledger.md

michalisFr:
- Added instructions on how to use an account on both Kusama and Statemine
- Removed reference to the support email


salmad3 (approved):
Thanks @michalisFr, we'll prettify after merging.

PR #2919, opened 2021-12-23
updated document location

jacobnash:
Both learn-parachain, and learn-crosschain are in the docs directory. They were looking for them in the learn directory.


salmad3 (approved):
Thanks @jacobnash

PR #2984, opened 2022-01-21
Fix Nominator Inject Values

salmad3:
image


DrW3RK:
Ran locally. The value rendered is 256. Merging this

PR #2996, opened 2022-01-28
Update kusama-statemine-ledger.md

### LINKED ISSUE #2997: Clarify Ledger Support of Statemine
Comments from @michalisFr 

Polkadot and Kusama accounts on Polkadot-JS UI are now visible on Statemint and Statemine as well, even if they're not enabled for all chains (Reference: https://github.com/polkadot-js/apps/issues/5439)

However, this has the following undesired side effect: Any Ledger accounts in the extension will also show up, but they cannot be used on Statemint and Statemine because the extension doesn't support these two chains for Ledger.

For Statemint that's expected because there's no Ledger app yet. However, if a user sends tokens to that account on Statemint they'll be inaccessible until an app is released.

For a Statemine account this is easily fixable if the user removes the account from the extension and adds it directly on Polkadot-JS UI, at least until the extension supports Statemine.

michalisFr:
Updated information on a Ledger account being available on both Kusama and Statemine

### DIFF
--- a/docs/general/kusama/kusama-statemine-ledger.md
+++ b/docs/general/kusama/kusama-statemine-ledger.md
@@ -103,9 +103,9 @@ prompt your ledger for confirmation when you initiate a transaction.
 
 ### Working on both Kusama and Statemine
 
-Polkadot and Kusama accounts will show up in Statemint and Statemine respectively, even if they're not set to be used on all chains. What this means is that if you already have a Kusama Ledger account, that account will also be available on Statemine. But the opposite is not the case: an account created specifically on Statemine or Statemint won't show up on Kusama or Polkadot respectively. 
+Polkadot and Kusama accounts from the [extension](https://wiki.polkadot.network/docs/learn-account-generation#polkadotjs-browser-extension) will show up in Statemint and Statemine respectively, even if they're not set to be used on all chains. The reasoning behind this is that Statemine, which is a common good parachain on Kusama uses KSM as its native token. Hence, the same accounts that were selected for Kusama will also be injected into Statemine. What this means is that if you already have a Kusama Ledger account configured on the extension, that account will also be available on Statemine. But the opposite is not the case: an account created specifically on Statemine or Statemint won't show up on Kusama or Polkadot respectively. 
 
-**However**, as mentioned above, the Polkadot extension does not support Statemine Ledger accounts at the moment. So, if your Kusama account is in the Polkadot extension, you'll need to remove it from there and re-add it on Polkadot-JS UI directly **as a Kusama account**, as described [here](https://guide.kusama.network/docs/kusama-ledger#using-on-polkadot-js-apps-ui).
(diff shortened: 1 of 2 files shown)


DrW3RK:
@michalisFr Please review the changes made. If they look good, this PR is ready to be merged

docs/general/kusama/kusama-statemine-ledger.md
```diff
@@ -103,9 +103,9 @@ prompt your ledger for confirmation when you initiate a transaction.
 
 ### Working on both Kusama and Statemine
 
-Polkadot and Kusama accounts will show up in Statemint and Statemine respectively, even if they're not set to be used on all chains. What this means is that if you already have a Kusama Ledger account, that account will also be available on Statemine. But the opposite is not the case: an account created specifically on Statemine or Statemint won't show up on Kusama or Polkadot respectively. 
+Polkadot and Kusama accounts from the [extension](https://wiki.polkadot.network/docs/learn-account-generation#polkadotjs-browser-extension) will show up in Statemint and Statemine respectively, even if they're not set to be used on all chains. The reasoning behind this is that Statemine, which is a common good parachain on Kusama uses KSM as its native token. Hence, the same accounts that were selected for Kusama will also be injected into Statemine. What this means is that if you already have a Kusama Ledger account configured on the extension, that account will also be available on Statemine. But the opposite is not the case: an account created specifically on Statemine or Statemint won't show up on Kusama or Polkadot respectively.
```
michalisFr: The additions that this only applies to extension accounts is incorrect. All accounts, regardless of type, will be available on Statemine and Statemint.

Also, this: 
`Statemine, which is a common good parachain on Kusama uses KSM as its native token. Hence, the same accounts that were selected for Kusama will also be injected into Statemine.`
should probably be:
`Statemint and Statemine, which are common good parachains on Polkadot and Kusama respectively, use DOT and KSM as their native tokens. Hence, accounts on Polkadot and Kusama makes sense to also be available on their respective common good parachains`

docs/general/kusama/kusama-statemine-ledger.md
```diff
@@ -103,9 +103,9 @@ prompt your ledger for confirmation when you initiate a transaction.
...
 
-**However**, as mentioned above, the Polkadot extension does not support Statemine Ledger accounts at the moment. So, if your Kusama account is in the Polkadot extension, you'll need to remove it from there and re-add it on Polkadot-JS UI directly **as a Kusama account**, as described [here](https://guide.kusama.network/docs/kusama-ledger#using-on-polkadot-js-apps-ui).
+**However**, as mentioned above, the [Polkadot extension](https://wiki.polkadot.network/docs/learn-account-generation#polkadotjs-browser-extension) does not support Statemine Ledger accounts at the moment, but the Polkadot-JS UI does. So, if your Kusama account is in the Polkadot extension, you'll need to remove it from there and re-add it on Polkadot-JS UI directly **as a Kusama account**, as described [here](https://guide.kusama.network/docs/kusama-ledger#using-on-polkadot-js-apps-ui). After which, you will be able to make transactions with that account.
```
michalisFr: `After that` instead of `After which`

docs/general/kusama/kusama-statemine-ledger.md
profradha: perfect!

PR #3013, opened 2022-02-04
Add Blockchair to the list of explorers

TokhaMyshkin:
Added Blockchair to the list of block explorers.


salmad3:
Thanks @LucasRoorda. At the moment, we limit the build tools that are from within the ecosystem. I will close this for now.

PR #3051, opened 2022-02-23
Use production profile for compiling validators

bkchr:


docs/maintain/maintain-guides-how-to-validate-polkadot.md
```diff
@@ -255,7 +255,7 @@ git checkout VERSION
 Build native code with the cargo release profile.
```
profradha: Probably need to update this sentence to -

```suggestion
Build native code with the production profile.
```

PR #3052, opened 2022-02-23
Add General Tools page and Block Time Estimator

a-jwc:
Created a tools page under "general" and a block time estimator component.

image

Please let me know if any other features or changes are desired. Thanks!


salmad3:
I changed the base to another branch so we can test on staging/locally

PR #3099, opened 2022-03-09
fix broken links in learn-architecture section

kelvinwelter:
Hello! While studying Polkadot's architecture, I ended up finding some broken links. There were 4 links that should lead to a certain term in the glossary, as well as one link that should lead to a section of the article. These 5 links were with a double #, for example, "glossary##validator". I just removed these double hashtags to make the links functional again.


DrW3RK:
Thank you!

PR #3151, opened 2022-03-29
Fix markdown anchors

sacha-l:
I noticed the anchors were broken. This PR should fix them. Also added some full-stops here and there.


DrW3RK:
Thank you @sacha-l

PR #3152, opened 2022-03-29
Fix conditional rendering in validator community overview page

cmppoon:
I noticed that the conditional rendering syntax was incorrect, so the page did not render correctly. I just fixed it by adding a space before the closing curly braces.


DrW3RK:
@cmppoon Thank you!

PR #3155, opened 2022-03-31
Add "Serialized transaction format" subsection to "Transaction Format"

IkerAlus:
- Added a subsection to "Transaction Format" section to add details about the actual format of a serialized transaction.
- Added the "Summary" subsection so it is clear the enumeration of the transaction process refers to the whole section.

The info for this update has been mainly take from [this example](https://github.com/paritytech/polkadot-interaction-examples-rs/blob/main/src/bin/05_transfer_balance.rs).

### DIFF
--- a/docs/build/build-transaction-construction.md
+++ b/docs/build/build-transaction-construction.md
@@ -51,6 +51,10 @@ Before being submitted, transactions are serialized. Serialized transactions are
   - 1 byte: the pallet index the transaction is calling into.
   - 1 byte: the function in the pallet the transaction is calling.
   - variable: the SCALE-encoded parameters required by the function being called.
+- 4 bytes: Spec version of the runtime.
+- 4 bytes: Transaction version of the runtime.
+- 32 bytes: Genesis hash of the chain.
+- 32 bytes: Block hash serving as the era reference. If the transaction is immortal, then this would be the genesis hash.
  
 The metadata provides you with all of the information required to know how to construct the serialized call data specific to your transaction. You can read more about the metadata, its format and how to get it in the [Substrate documentation](https://docs.substrate.io/v3/runtime/metadata/).
 
--- a/docs/build/build-transaction-construction.md
+++ b/docs/build/build-transaction-construction.md
@@ -36,6 +36,7 @@ function from the Balances pallet will take:
 - `#[compact] value`: Number of tokens (compact encoding)
 
 **Serialized transaction format**
+
 Before being submitted, transactions are serialized. Serialized transactions are hex encoded SCALE-encoded bytes with the following format:
 - Compact encoded number of SCALE-encoded bytes following this.
 - 1 bit: it is a 0 if no signature is present, or a 1 if it is.
@@ -44,8 +45,8 @@ Before being submitted, transactions are serialized. Serialized transactions are
   - a SCALE encoded `sp_runtime::MultiAddress::Id<AccountId32, u32>` indicating the signer(s) of the transaction.
   - a SCALE encoded `sp_runtime::MultiSignature::S225519` with the signature.
   - a SCALE encoded `sp_runtime::generic::Era` indicationg for how long this transaction will live in the pool.
-  - Compact encoded u32 with the nonce.
-  - Compact encoded u128 with the tip paid to the block producer.
+  - Compact encoded `u32` with the nonce.
+  - Compact encoded `u128` with the tip paid to the block producer.
 - The specific transaction parameters or call data, which consists of:
   - 1 byte: the pallet index the transaction is calling into.
   - 1 byte: the function in the pallet the transaction is calling.
--- a/docs/build/build-transaction-construction.md
+++ b/docs/build/build-transaction-construction.md
@@ -38,23 +38,25 @@ function from the Balances pallet will take:
 **Serialized transaction format**
 
 Before being submitted, transactions are serialized. Serialized transactions are hex encoded SCALE-encoded bytes. The specific serialization is defined in the runtime and can change if the runtime is upgraded, but in general the serialization format can be described as follows:
-- Compact encoded number of SCALE-encoded bytes following this.
+- Compact encoded number of SCALE encoded bytes following this.
 - 1 bit: it is a 0 if no signature is present, or a 1 if it is.
-- 7 bits: the extrinsic version, it is equal to 4 in decimal.
+- 7 bits: the extrinsic version, it is equal to 4 in decimal.\- 4 bytes: Spec version of the runtime.
+- 4 bytes: Transaction version of the runtime.
+- 32 bytes: Genesis hash of the chain.
+- 32 bytes: Block hash serving as the era reference. If the transaction is immortal, then this would be the genesis hash.
 - If there is a signature:
   - a SCALE encoded `sp_runtime::MultiAddress::Id<AccountId32, u32>` indicating the signer(s) of the transaction.
   - a SCALE encoded `sp_runtime::MultiSignature::{SigningScheme}` with the signature\*.
-  - a SCALE encoded `sp_runtime::generic::Era` indicating for how long this transaction is valid.
+  - a SCALE encoded `sp_runtime::generic::Era` indicating for how long this transaction is valid:
+    - If the transaction is immortal, the Era would be simply 0.
+    - Otherwise, it would be a `Vec[u64, u64]` comprising the period and the phase.  
   - Compact encoded `u32` with the nonce.
   - Compact encoded `u128` with the tip paid to the block producer.
+  - a SCALE encoded `sp_runtime::traits::SignedExtension<Vec<Text>>` with the additional data and logic associated with this transaction.
 - The specific transaction parameters or call data, which consists of:
   - 1 byte: the pallet index the transaction is calling into.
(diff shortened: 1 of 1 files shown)


docs/build/build-transaction-construction.md
```diff
@@ -35,6 +35,27 @@ function from the Balances pallet will take:
...
 
+**Serialized transaction format**
+
+Before being submitted, transactions are serialized. Serialized transactions are hex encoded SCALE-encoded bytes with the following format:
+- Compact encoded number of SCALE-encoded bytes following this.
+- 1 bit: it is a 0 if no signature is present, or a 1 if it is.
+- 7 bits: the transaction protocol version.
+- If there is a signature:
+  - a SCALE encoded `sp_runtime::MultiAddress::Id<AccountId32, u32>` indicating the signer(s) of the transaction.
+  - a SCALE encoded `sp_runtime::MultiSignature::S225519` with the signature.
```
joepetrowski: Polkadot supports sr25519, ed25519, and ecdsa. This is just an example using sr25519 (with a typo :)), so the text should either indicate that or just make the example generic.
```suggestion
  - a SCALE encoded `sp_runtime::MultiSignature::{SigningScheme}` with the signature.
```

docs/build/build-transaction-construction.md
```diff
@@ -35,6 +35,27 @@ function from the Balances pallet will take:
 - `dest`: Destination address
 - `#[compact] value`: Number of tokens (compact encoding)
 
+**Serialized transaction format**
+
+Before being submitted, transactions are serialized. Serialized transactions are hex encoded SCALE-encoded bytes with the following format:
+- Compact encoded number of SCALE-encoded bytes following this.
+- 1 bit: it is a 0 if no signature is present, or a 1 if it is.
+- 7 bits: the transaction protocol version.
```
joepetrowski: This is different from the `transaction_version`. It's confusing, but at least note it here.

joepetrowski (commented):
Good start. Would add more contextual info, like that the actual serialization is defined in the runtime and can change. And at the end format for line length.

docs/build/build-transaction-construction.md
IkerAlus: good point, updated and added a little clarification.

docs/build/build-transaction-construction.md
IkerAlus: I though these 7 bits was exactly `transaction_version` as described [here](https://docs.substrate.io/v3/runtime/upgrades/).

What are they then? I will update this accordingly.

joepetrowski: Nope, so this is a bit lower level, it's like the version of "what an extrinsic is" _to Substrate_, while `transaction_version` refers to backwards incompatible changes _to a specific runtime,_ namely when you break pallet/call indexing, function signatures (including types), or types in the serialized transaction itself (e.g. `AccountId` or `SignedExtension`s).

`transaction_version` came much later than `extrinsic_version`, and was a bit of a naming collision that most people just didn't notice. Unless there is a major major major change to Substrate, `extrinsic_version` will basically always `= 4`, and a change would mean basically needing to update everything.

You can see a few examples of its use in txwrapper:

- [Decoding extrinsics](https://github.com/paritytech/txwrapper-core/blob/d590b0912ee0bf9f357233024c064138cff36cd2/packages/txwrapper-core/src/core/decode/decodeSigningPayload.ts#L38) (notice `transaction_version` at L54)
- [The definition of a "method"](https://github.com/paritytech/txwrapper-core/blob/3cc7bf3e63f1e88d231454a98307afe4ab798b27/packages/txwrapper-core/src/core/method/defineMethod.ts#L106)
- [Explained a bit in an example](https://github.com/paritytech/txwrapper-core/blob/b3d7fc1ea49244cbda18d08f14a5d2cdd0b2e78c/packages/txwrapper-core/src/core/construct/createSigningPayload.ts#L27)

docs/build/build-transaction-construction.md
```diff
@@ -35,6 +35,29 @@ function from the Balances pallet will take:
...
+**Serialized transaction format**
+
+Before being submitted, transactions are serialized. Serialized transactions are hex encoded SCALE-encoded bytes. The specific serialization is defined in the runtime and can change if the runtime is upgraded, but in general the serialization format can be described as follows:
+- Compact encoded number of SCALE-encoded bytes following this.
+- 1 bit: it is a 0 if no signature is present, or a 1 if it is.
+- 7 bits: the transaction version.
+- If there is a signature:
+  - a SCALE encoded `sp_runtime::MultiAddress::Id<AccountId32, u32>` indicating the signer(s) of the transaction.
+  - a SCALE encoded `sp_runtime::MultiSignature::{SigningScheme}` with the signature\*.
+  - a SCALE encoded `sp_runtime::generic::Era` indicationg for how long this transaction will live in the pool.
```
joepetrowski: Its being kicked from the pool is just a byproduct of it being invalid from the runtime's perspective (the pool won't include transactions that will _obviously_ fail, e.g. nonce too low, too old, etc.)
```suggestion
  - a SCALE encoded `sp_runtime::generic::Era` indicating for how long this transaction is valid.
```

salmad3: @IkerAlus

joepetrowski: @DannyS03 - he's on vacation, he'll get back to it when he comes back.

IkerAlus: good point, changed as suggested

docs/build/build-transaction-construction.md
IkerAlus: thanks a lot for the detailed context around this. 

I updated the text to "extrinsic version" and added the point that it should be 4.

IkerAlus: I updated the description to "extrinsic version" also based on this PR: https://github.com/paritytech/substrate/commit/6bcc3851909284f934461e4dc45fec72ce1cb525

docs/build/build-transaction-construction.md
```diff
@@ -35,6 +35,29 @@ function from the Balances pallet will take:
...
+- If there is a signature:
+  - a SCALE encoded `sp_runtime::MultiAddress::Id<AccountId32, u32>` indicating the signer(s) of the transaction.
+  - a SCALE encoded `sp_runtime::MultiSignature::{SigningScheme}` with the signature\*.
+  - a SCALE encoded `sp_runtime::generic::Era` indicating for how long this transaction is valid.
+  - Compact encoded `u32` with the nonce.
+  - Compact encoded `u128` with the tip paid to the block producer.
+- The specific transaction parameters or call data, which consists of:
+  - 1 byte: the pallet index the transaction is calling into.
+  - 1 byte: the function in the pallet the transaction is calling.
+  - variable: the SCALE-encoded parameters required by the function being called.
```
joepetrowski, line 60:
```suggestion
  - variable: the SCALE-encoded parameters required by the function being called.
- 4 bytes: Spec version of the runtime.
- 4 bytes: Transaction version of the runtime.
- 32 bytes: Genesis hash of the chain.
- 32 bytes: Block hash serving as the era reference. If the transaction is immortal, then this would be the genesis hash.
```

docs/build/build-transaction-construction.md
```diff
@@ -35,6 +35,29 @@ function from the Balances pallet will take:
...
+**Serialized transaction format**
+
+Before being submitted, transactions are serialized. Serialized transactions are hex encoded SCALE-encoded bytes. The specific serialization is defined in the runtime and can change if the runtime is upgraded, but in general the serialization format can be described as follows:
+- Compact encoded number of SCALE-encoded bytes following this.
+- 1 bit: it is a 0 if no signature is present, or a 1 if it is.
+- 7 bits: the extrinsic version, it is equal to 4 in decimal.
+- If there is a signature:
+  - a SCALE encoded `sp_runtime::MultiAddress::Id<AccountId32, u32>` indicating the signer(s) of the transaction.
+  - a SCALE encoded `sp_runtime::MultiSignature::{SigningScheme}` with the signature\*.
+  - a SCALE encoded `sp_runtime::generic::Era` indicating for how long this transaction is valid.
```
joepetrowski: Can @Imod7 or @tarikgul comment on encoding for mortal vs. immortal transactions?

joepetrowski (commented):
Left some additions, then I think it's just missing signed extensions? Once those are addressed then I think it's ready.

docs/build/build-transaction-construction.md
TarikGul: For encoding between an `Immortal` and `Mortal` era, the following `impl` is useful: https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/generic/era.rs#L102

An immortal era will default to the encoded value of `0x00`. It will be just a single byte of `0`.

A mortal era on the other hand will be `Vec[u64, u64]`; Comprising of the Period, and Phase.

IkerAlus:
> Left some additions, then I think it's just missing signed extensions? Once those are addressed then I think it's ready.

I added the info for the era encoding and the signed extensions point in my last commit. It would be nice if @joepetrowski double checks the point about the signed extensions. It is not entirely clear to me how this info should be referenced here for serialization.

joepetrowski (approved):
Would just run `prettier` for formatting, then it's good!

PR #3183, opened 2022-04-06
Initialize Misc Page

### LINKED ISSUE #3137: Initialize Misc Page
Create a new page for miscellaneous content as per previous TechEd discussion.

salmad3:
- Initialized by adding [experimental] content, which should be considered cautiously. 
- The misc page is meant to be a place for additional info about Polkadot topics, a place for "other" topics, including additional perspectives.
- The team should outline topics that are missing but have no place as a follow up to this initialization.
- Only added to Polkadot Wiki for now
- Resolves #3137


emresurmeli:
I think we can call this page something that's more explanatory of its intention. Polkadot analogies, Polkadot outside the box, How to think about Polkadot, etc. Misc is a page where I would think are things that don't have a section, but this is more of a page that helps people think about Polkadot by explaining it with different thought approaches.

salmad3:
right, my thought was to have a misc doc available for content that doesn't currently have a place, and the content in this PR be a section of it 
> The team should outline topics that are missing but have no place as a follow up to this initialization

PR #3187, opened 2022-04-06
Update Kusama Max Nominations

### LINKED ISSUE #3186: Update Kusama Staking Page - Max Nominations to 24
image
https://guide.kusama.network/docs/maintain-guides-how-to-nominate-kusama/

DrW3RK:
Create variables for both Polkadot and Kusama max nominations with default values

As the docs of Kusama and Polkadot get indexed on both Polkadot Wiki and Kusama guide, fetching the value from the chain constants might render wrong values. An example scenario is that someone on Polkadot Wiki land on the Kusama Staking page through the search bar (Although it is inaccessible through the sidebar), they will end up seeing Polkadot's max nominations value instead. Hence, separate variables with default values were created.


emresurmeli (commented):
We can add a line about the max nominations being subject to change via governance/runtime upgrade.

Something along the lines of:
""... distributed up to {{ max_nominations }} validators. This value can be changed via runtime upgrade if the proposal passes on-chain governance..."

DrW3RK:
@emresurmeli It is a good info for the reader to be presented with. My only concern is that the nomination process is already info heavy and that would be a non-consequential info for someone that is trying to nominate (given we would be updating this value in the doc right after a runtime upgrade.)

The parameters page has the exact disclaimer at the top and the staking section can be updated. Let me do that there https://wiki.polkadot.network/docs/maintain-polkadot-parameters

emresurmeli (approved):

emresurmeli (approved):
Yup fair enough, lgtm

PR #3196, opened 2022-04-07
Update learn-polkadot-host.md

Noc2:
Add Kagome and remove the part about chainsafe, which is not relevant for polkadot host.


DrW3RK:
Thanks David!

Issue #3133, opened 2022-03-23
Add Admonitions Across Wiki


salmad3:
https://docusaurus.io/docs/next/markdown-features/admonitions

salmad3:
First as a default for some notes, warnings and info, and then custom ones where applicable

salmad3:
Admonitions added to all docs

PR #3198, opened 2022-04-08
Add Admonitions to Account Content

salmad3:
Part of https://github.com/w3f/polkadot-wiki/issues/3133

May change based on page reviewals, but these are to have them live now

### DIFF
--- a/docs/learn/learn-accounts.md
+++ b/docs/learn/learn-accounts.md
@@ -40,8 +40,12 @@ Polkadot's default address format to represent many different address types. Thi
 byte**, **32 byte**, and **arbitrary raw byte** variants. It also allows an enhancement to the
 original [`indices`](learn-accounts.md##indices) lookup.
 
-> Many wallets allow you to convert between formats. Stand-alone tools exist as well; you can find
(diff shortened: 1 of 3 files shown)


docs/learn/learn-accounts.md
```diff
@@ -465,13 +485,17 @@ For this example, we will be using the
...
 and send a transaction with it.
 
-> While Westend is meant to replicate the Polkadot mainnet as closely as possible, there are a few
-> notable differences:
->
-> - Existential deposit is equal to 0.01 WND (Westies; Westend's native coin) instead of 1 DOT.
-> - The multi-signature transaction deposit is equal to ~1 WND instead of ~20.2 DOT.
->
-> The photos below reflect values in WND, but instructions are the same for DOT.
+:::note Differences for Westend
```
profradha: Title probably not needed. Polkadot and Westend have the same code base, but may have different constants in the chain state. The title does not capture that, while the content does. Better to leave it empty

```suggestion
:::note 
```

DrW3RK (approved):
Looks good to me!

PR #3224, opened 2022-04-13
Update build-smart-contracts.md

Noc2:
My suggestion is to remove edgeware here, since it’s not a parachain and listed under parachains. Let me know if you want to add more smart contract projects here, like https://github.com/gear-tech/gear


salmad3 (approved):
thanks @Noc2

Yes, we can add other smart contract projects

PR #3237, opened 2022-04-20
Add Dedicated Slashing Page

paradox-tt:
Unformatted document


salmad3 (changes requested):
- would advise adding an instance or example where applicable
- maintain-slashing.md would be better as slashing is a key aspect of maintaining the network, so having information on slashing in general and then leading into best practices seems like a better approach where we can use content on slashing already available on the wiki and treat this doc as the main content for slashing

paradox-tt:
I'll get some examples posted in Polkassembly.

I agree with the advised document flow.

salmad3:
After the fact, we can add the page to both sidebars for the Polkadot Wiki and Kusama Guide

emresurmeli:
Let's add the page to the sidebars and add 1-3 examples

cc: @paradox-tt

paradox-tt (commented):
Sorry for the delay, please find examples tabulated below.

DrW3RK:
Thank you @paradox-tt Will review this soon!

DrW3RK (approved):
Thank you @paradox-tt Merging this PR and will create a new PR for linking appropriate Wiki docs and polish the content further

PR #3290, opened 2022-05-05
Max-width

emresurmeli:
Add a max width of 1920px (avg screen size), screen larger than 1920px wide will have white margin on left and right


DrW3RK:
The staging websites render as expected. Let's merge this fix for now.

DrW3RK (approved):

PR #3325, opened 2022-05-11
Add warning note to immortal tx mention

CurlyBracketEffect:
Just adding the same warning note that is found [here](https://wiki.polkadot.network/docs/build-protocol-info#transaction-mortality)

Based on conversations with the Integration team, we felt there needed to be some more visibility brought to the risks involved with making a transaction immortal.


joepetrowski (approved):
Just adjust for line length

PR #3363, opened 2022-05-23
Update Proxy doc

### LINKED ISSUE #3362: Update Proxies doc
https://wiki.polkadot.network/docs/learn-proxies#using-time-delayed-proxies

This section has misleading info on block delay that needs to be fixed. Block time is 6 seconds

"When creating a proxy through the PolkadotJS application, we are provided a delay field. In this example we are creating a proxy with a delay value of 100, which means 100 blocks. 100 * 6(minutes) = 600 minutes, or 10 hours."

Also, multiple UI snapshots are outdated. addProxy includes delay as one of the input arguments.

DrW3RK:
Add new snapshots, correct info about time delayed proxies and add relevant info


emresurmeli (approved):
``

PR #3420, opened 2022-06-20
Banner for Polkadot decoded

emresurmeli:
Quick hack, will refine later


DrW3RK (approved):
Needs refinement. The Polkadot banner disappears when scrolling down on page. Also, it would be wise to have this show up on all Wiki docs, as users may not navigate to the homepage often (as they would be directly clicking on indexed docs on search engines). Substrate docs already implemented this. Example: https://docs.substrate.io/main-docs/fundamentals/node-and-network-types/

We can work on the improvements in a different PR. This on is ready to be merged

PR #3424, opened 2022-06-23
[WIP] Fix Wiki Layout Bug

carinaluise:
Changed max-width value to accommodate for larger screen sizes. (Value comes from the largest breakpoint available on Browserstack)
It also works if we remove the --custom-content-max-width variable from the docs container all together as well. 
@rmnprkrl


rmnprkrl:
@carinaluise let's set the `max-width` to `none` so it goes full width, same like GH layout eg:
https://docs.github.com/en/repositories

rmnprkrl (approved):

PR #3436, opened 2022-06-24
Added reference to dedicated server for validating node and list of dedicated server providers

stakeworld:
My first addition to the wiki, adding some information about using dedicated servers for a node.

### DIFF
--- a/docs/maintain/kusama/maintain-guides-how-to-validate-kusama.md
+++ b/docs/maintain/kusama/maintain-guides-how-to-validate-kusama.md
@@ -66,6 +66,8 @@ choose whatever [VPS](#vps-list) provider that your prefer, and whatever operati
 comfortable with. For this guide we will be using **Ubuntu 18.04**, but the instructions should be
 similar for other platforms.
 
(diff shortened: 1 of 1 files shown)


docs/maintain/kusama/maintain-guides-how-to-validate-kusama.md
```diff
@@ -66,6 +66,8 @@ choose whatever [VPS](#vps-list) provider that your prefer, and whatever operati
 comfortable with. For this guide we will be using **Ubuntu 18.04**, but the instructions should be
 similar for other platforms.
 
+Most validators who are serious about setting up an active validator progress to a [dedicated server](#dedicated-server-providers) because the process of (parachain) validating gets more and more resource intensive and VPS providers seem to fall short in performance, despite their good looking specifications.
```
joepetrowski: One, this is a long sentence, I'd break it up. Two, it may be the case now but it's a bit of a generalization to say that validating will get more resource intensive (plus, it's relative to "standard" hardware, which also improves over time). I would do this like so:

```suggestion
Node hardware should run at a level at or above the standard benchmark machines. Many VPS providers
seem to fall short in practice, despite good looking specifications. Most validators who are
serious about setting up an active validator progress to a
[dedicated server](#dedicated-server-providers).
```

And run prettier for line length :).

stakeworld:
Thanks @joepetrowski for the tips

DrW3RK (approved):

PR #3450, opened 2022-06-28
Fetch values from the chain state correctly

### LINKED ISSUE #3447: Investigate 'No value found, seeking default for SOME_CONST_VAR' build warnings
This warning surfaces several times during the build process.  This isn't a major issue since the default values are active, but it should be investigated as to why we are unable to query the original targets.  If failures are due to simple API updates, attempt to restore the query and avoid the default value from registering.

```
No value found, seeking default for block_target_in_seconds
In query for query.staking.validatorCount
In const for consts.proxy.proxyDepositBase
No value found, seeking default for tip_deposit_amount
In const for consts.treasury.tipFindersFee
No value found, seeking default for tip_finders_fee
In const for consts.treasury.spendPeriod
No value found, seeking default for identity_reserve_funds
No value found, seeking default for identity_reserve_funds_ksm
In const for consts.identity.reserveFunds
No value found, seeking default for identity_sub_reserve_funds
No value found, seeking default for identity_sub_reserve_funds_ksm
In const for consts.identity.reserveFieldFunds
No value found, seeking default for identity_field_funds
In const for consts.balances.existentialDeposit
No value found, seeking default for min_dot_nominator_intention
In query for query.staking.maxNominatorsCount
In const for consts.electionProviderMultiPhase.maxElectingVoters
In const for consts.staking.maxNominatorRewardedPerValidator
In const for consts.staking.maxNominatorRewardedPerValidator
No value found, seeking default for kusama_max_nominations
No value found, seeking default for polkadot_max_nominations
```

DrW3RK:
Either the path is incorrect or the path has changed with changes to the API

For example, the variables for the constants of the Identity have changed

image

### DIFF
--- a/scripts/inject-dict.json
+++ b/scripts/inject-dict.json
@@ -52,7 +52,7 @@
   {
     "tpl": "identity_reserve_funds",
     "default": { "kusama": 0.033333, "polkadot": 20.258 },
-    "path": "consts.identity.reserveFunds",
+    "path": "consts.identity.basicDeposit",
     "filters": ["humanReadableToken"]
   },
   {
@@ -62,7 +62,7 @@
   {
     "tpl": "identity_sub_reserve_funds",
     "default": { "kusama": 0.006666, "polkadot": 20.053 },
-    "path": "consts.identity.reserveFunds",
+    "path": "consts.identity.subAccountDeposit",
     "filters": ["humanReadableToken"]
   },
   {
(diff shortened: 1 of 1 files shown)


DrW3RK:
No value found warning can also show up when there is no path provided for the variable. For instance, the newly added votingLockPeriod variables have no path specified. There must be a better way to incorporate these values to the dictionary I guess! or ensure such warning does not show up when no path is specified.

alfarok:
I need to dig into the docusaurus docs to get more familiar.  ~~I'm surprised they don't provide an out-of-the-box solution for fetching and setting global constants.~~  I see now it is encouraged to use plugins for something like this, which is what we are already doing.
https://github.com/facebook/docusaurus/issues/395#issuecomment-393377433

alfarok:
@DrW3RK I pulled these 3 changes in #3464 and fixed 2 more that needed updating.  Feel free to re-open if you wanted to add anything additional in this PR.

Issue #3446, opened 2022-06-28
Address build warnings regarding multiple versions of dependencies


alfarok:
Address build warnings regarding multiple versions of dependencies, **ensure that there is only one installed.**

I am seeing this for at least the following dependencies several times:
- @polkadot/util
- @polkadot/util-crypto

Example of console output when compiling:
```
@polkadot/util has multiple versions, ensure that there is only one installed.
Either remove and explicitly install matching versions or dedupe using your package manager.
The following conflicting packages were found:
        esm 9.5.1       node_modules/@polkadot/api/node_modules/@polkadot/util/
        esm 9.5.1       node_modules/@polkadot/rpc-augment/node_modules/@polkadot/util/
```

Further research is required to determine how many of these are resolvable verses having a logical reasoning for using different versions.

alfarok:
addressed in #3452

Issue #2415, opened 2021-08-02
Address conversion tool is not working


laboon:
If you enter an address, it does not change when you select a new network. 

https://wiki.polkadot.network/docs/learn-accounts#address-conversion-tools

jakehemmerle:
After spending about an hour trying to get this to work, I say we scrap it and tell them to redirect to one of the converters we have listed below. Thoughts?

jakehemmerle:
I think it's related to some static JS rendering differences between docusaurus v1 and v2 (since the script was created by Logan a year ago)

DrW3RK:
<img width="804" alt="Screenshot 2021-08-03 at 19 06 00" src="https://user-images.githubusercontent.com/86818441/128057115-ab898104-0c1a-4b33-bdfa-0e3def16abc5.png">

Needs a deeper look as to why this might be happening

DrW3RK:
Yup, indeed this is a migration issue. Not with docusaurus, but with the way we reorganized wiki content folders

https://wiki.polkadot.network/js/custom.js   - The js folder is outside the docs folder. Looks like this would break all the pages that are currently using javascript in them

salmad3:
@a-jwc said he would take a crack at this

DrW3RK:
Keep me posted. I can look into this again tomorrow

DrW3RK:
Figured out the issue. Just needed to reference to JS files appropriately in config file. Creating a PR and merging it to Master

DrW3RK:
![image](https://user-images.githubusercontent.com/86818441/129708905-ae5ecb08-4a0f-4170-adb8-6f1f0d672ef8.png)
Made progress with correctly referencing JS files in the latest repo structure, but looks like the address changer JS file itself has an issue. Debugging this is beyond our scope I guess. I will remove the address converter tool from the doc. If anyone wants to debug this, raise a new issue/PR

PR #3452, opened 2022-06-29
Resolve duplicate dependency issue with @polkadot/util and @polkadot/util-crypto

alfarok:
This PR address #3446.

I upgraded `@polkadot/util` and `@polkadot/util-crypto` in the package.json file to allow up to the latest released version which is 9.6.2 for both.  After clearing my existing node modules and re-running `yarn install`, the duplicate dependencies are removed (dropping version 8.2.2 for both).  Running `yarn polkadot:start` no longer shows the duplicate errors outlined in #3446.


alfarok:
Please let me know if anyone is aware of a legacy reason the two versions were used.  Everything appeared to build and function locally with the updated dependency tree but I will double check the functionality referenced in `addressChanger.js` tomorrow.  I believe this is the only location these modules are used.

DrW3RK:
I remember we had issues previously with address changer https://github.com/w3f/polkadot-wiki/issues/2415
In fact, we have stopped using it. The address convertor tool used to be here previously https://wiki.polkadot.network/docs/learn-accounts#address-conversion-tools

So, yeah. Nothing to worry about I guess!

alfarok:
Thanks for the link, it does look like it was remove but is still present in the translated docs in the repo.  I am not exactly sure how translations are handled.  

I am guessing that the address changer inline tool is present but broken in the translated versions.  I want to make sure swapping the dependency version doesn't cause any errors in those versions that still load `@polkadot/util` and `@polkadot/util-crypto` modules.

Example:
https://github.com/w3f/polkadot-wiki/blob/master/polkadot-wiki/translated_docs/it/learn-accounts.md#address-conversion-tools

@DrW3RK Do you know how I can view a translated version on localhost, such as:

http://localhost:3000/translated_docs/it/learn-accounts.md#address-conversion-tools

Since these modules aren't doing anything right now the solution might be to remove them completely after removing the translated references.

DrW3RK:
We have had problems with Translated docs as well in the past and it might make sense to reimplement that functionality from scratch. @emresurmeli do you have any input on this?

DrW3RK (approved):

PR #3464, opened 2022-07-01
Organize and unify formatting for fetched and default constant values

alfarok:
This PR addresses #3447 and accomplishes the following:
- [x] Cleans up conditional logic in existing `fetch.mjs`
- [x] Unifies the formatting of fetched and default values
- [x] Removes extra verbose logging for less unnecessary console noise
- [x] Refactors some naming for added clarity
- [x] Add more comments to outline how the constants are processed for future reference
- [x] Tested that output of `computed-dict.json` returns the same values for Polkadot and Kusama + additional formatting
- [x] Tested that all default values format to expected constants for both Polkadot and Kusama
- [x] Determine why default values are used when fetch paths are available
- [x] Fix broken fetch paths
- [x] Removes duplicate logic from `fetch.mjs` and `inject.mjs`

Example of 4 default values previously vs updated:
Formatting-Default-Values

Console output during compilation 
previously:
```
Connecting to node wss://rpc.polkadot.io
Connected
Working on the Polkadot wiki
No value found, seeking default for block_target_in_seconds
In query for query.staking.validatorCount
In const for consts.proxy.proxyDepositBase
Applying humanReadableToken to polkadot value 200080000000
In const for consts.proxy.proxyDepositFactor
Applying humanReadableToken to polkadot value 330000000
In const for consts.treasury.proposalBondMinimum
Applying humanReadableToken to polkadot value 1000000000000
In const for consts.treasury.proposalBondMaximum
Applying humanReadableToken to polkadot value 5000000000000
In const for consts.treasury.tipReportDepositBase
No value found, seeking default for tip_deposit_amount
In const for consts.treasury.tipFindersFee
No value found, seeking default for tip_finders_fee
In const for consts.treasury.spendPeriod
Applying blocksToDays to polkadot value 345600
In const for consts.identity.reserveFunds
No value found, seeking default for identity_reserve_funds
No value found, seeking default for identity_reserve_funds_ksm
In const for consts.identity.reserveFunds
No value found, seeking default for identity_sub_reserve_funds
No value found, seeking default for identity_sub_reserve_funds_ksm
In const for consts.identity.reserveFieldFunds
No value found, seeking default for identity_field_funds
In const for consts.balances.existentialDeposit
Applying humanReadableToken to polkadot value 10000000000
In const for consts.crowdloan.minContribution
Applying humanReadableToken to polkadot value 50000000000
In query for query.staking.minNominatorBond
No value found, seeking default for min_dot_nominator_intention
In query for query.staking.maxNominatorsCount
In const for consts.electionProviderMultiPhase.maxElectingVoters
In const for consts.staking.maxNominatorRewardedPerValidator
In const for consts.staking.maxNominatorRewardedPerValidator
No value found, seeking default for kusama_max_nominations
No value found, seeking default for polkadot_max_nominations
No value found, seeking default for vote_locking_period_0
No value found, seeking default for vote_locking_period_1
No value found, seeking default for vote_locking_period_2
No value found, seeking default for vote_locking_period_3
No value found, seeking default for vote_locking_period_4
No value found, seeking default for vote_locking_period_5
No value found, seeking default for vote_locking_period_6
In const for consts.democracy.voteLockingPeriod
Applying blocksToDays to polkadot value 403200
Applying humanReadableToken to polkadot value 100000000000
Save dict values successfully!
```
updated:
```
Connected to node at wss://rpc.polkadot.io
Active Project: Polkadot Wiki
Updated global constants in computed-dict.json
```
^If a constant has a path but fails to retrieve a value during the fetch, the event will be logged.  Nothing displays in the updated output because the 5 broken references are now resolved.  See comments below.


alfarok:
These are the constants that have fetch paths but fail to retrieve valid data, therefore fallback to defaults:

1) ~~tip_deposit_amount~~ `treasury.tipReportDepositBase` should be `tips.tipReportDepositBase`
2) ~~tip_finders_fee~~ `treasury.tipFindersFee` should be `tips.tipFindersFee`
3) ~~identity_reserve_funds~~ Addressed in #3450 
4) ~~identity_sub_reserve_funds~~  Addressed in #3450 
5) ~~identity_field_funds~~  Addressed in #3450 

~~Looking into the reasoning of this now that I understand the conditional flow.~~ ✔️

alfarok:
Sorry the diff is slightly confusing on this one but it should simplify adding and maintaining constants in the future.  It might be easier to look at the updated `fetch.mjs` file individually prior to looking at the diff.  Let me know if you have questions!

DrW3RK (changes requested):
Good work! Everything comiples as expected and the values seem correct, except for spend period which still shows in blocks in the log (but I see the correct value 24 on computed-dict.json. I'm seeing this when I run `yarn polkadot:inject`

image

Could you please double check and let me know if there is something to be addressed.

alfarok:
> Could you please double check and let me know if there is something to be addressed.

Interesting, I wasn't familiar with the yarn `polkadot:inject` command yet.  It looks like `inject.mjs` is responsible for this behavior.  It contains almost the same logic as `fetch.mjs` but there are a few differences.  It looks like even prior to these changes you could potentially get slightly different results, one example is when the human readable option is set for Kusama.

fetch.mjs:
```js
case "humanReadableToken":
      let decimals = 6
      if (wiki === "polkadot") {
        decimals = 3;
      }

      value = (value / values[wiki].precision).toFixed(decimals) + " " + values[wiki].symbol;
      break;
```

inject.mjs:
```js
case "humanReadableToken":
      value = (value / values[wiki].precision).toFixed(3) + " " + values[wiki].symbol;
      break;
``` 

I will spend a little time looking into how difficult it would be to consolidate the core logic in a consistent manner.

alfarok:
@DrW3RK I was able to achieve the expected results by removing the `inject.mjs` file and adding an additional command line argument to `fetch.mjs` (`--inject`).  The constants dictionary is calculated every time fetch is invoked but the injection flag must be set to `true` for the constants to be inserted into the html.  This should operate the same as having the 2 files, without having duplicate logic for the calculation of the constants.  The commands in the `package.json` were also updated so build, start, and inject should all work with the same legacy syntax.

image

scripts/fetch.mjs
```diff
@@ -1,120 +1,167 @@
...
+          to: to,
+        });
+        const changedFiles2 = results2
+          .filter((result) => result.hasChanged)
+          .map((result) => result.file);
+        console.log("Modified files for kusama/polkadot difference:", changedFiles2);
+      } catch (error) {
+        console.error("Error occurred:", error);
+        process.exit(1);
+      }
```
alfarok: This is the only logic that was unique to `inject.mjs` that wasn't already in `fetch.mjs`, everything else was duplicated.  I haven't verified this logic but it produces the same results as the previous state.

DrW3RK (approved):
Looks good to me! Thanks @alfarok

alfarok:
I was trying to validate the logic in `inject` and I think in the current state it only logs the calculated constants.  

The README shows this command was used after running `build` to visualize all the replacements of constants throughout the html files that are generated after building.  However, this never works because the replacements happen during the build so `inject` detects no differences.

```
Modified files: []
Modified files for kusama/polkadot difference: []
```

Long story short, I think we could have just scrapped the `inject` command as it appears to be a legacy feature.  I am going to merge this PR because it has already been reviewed and is pretty big.  I will file a separate issue and follow-up with a smaller PR that just removes the stale code.

PR #3466, opened 2022-07-04
Clarify NPoS Election Algorithms on Polkadot

### LINKED ISSUE #3461: Refactro: Don't mention Phragmen as if it is a core part of the Polkadot NPoS protocol.
On a related note, we should start a sweep and remove everything along the lines of 

> The way the [Nominated Proof of Stake (NPoS) consensus mechanism](https://wiki.polkadot.network/docs/learn-consensus#nominated-proof-of-stake) works is that it uses [Phragmén algorithm](https://wiki.polkadot.network/docs/learn-phragmen) to map the nominators to their selected validators and produce the active set of validators based on some criteria.

This is factually incorrect. There is nothing in the polkadot protocol that says phragmen. The main bottleneck in the new system is that the chain allows for any algorithm to come up with a solution, submitted to the chain, and verifies it. This is the main bottleneck and the reason that NPoS won't scale. 

More accurate would have been to state: The NPoS system allows an algorithm to efficiently distribute the nominator stakes among validators, such that the economic security of the chain is optimized, and properties such as PJR are satisfied. An example of this algorithm is the Phragmen algorithm.
_Originally posted by @kianenigma in https://github.com/w3f/polkadot-wiki/issues/3434#issuecomment-1172286864_

DrW3RK:
Clarify that Phragmen algorithm is not an integral part of Polkadot NPoS system, but is just one of the algorithms that can be used to compute the election solution


docs/learn/learn-phragmen.md
```diff
@@ -1,20 +1,35 @@
...
+description: Learn about the election method used on Polkadot's Nominated Proof of Staking.
+keywords: [phragmen, sequential phragmén method, elections, algorithm, phragmms]
 slug: ../learn-phragmen
 ---
 
+## NPoS Election Algorithms
+
+Since validators are paid almost equally in Polkadot in each era, it is important that the stake behind each validator 
+is uniformaly spread out. Any algorithm for Nominated Proof of Staking (NPoS) on Polkadot should try to optimize three 
+metrics when computing an election solution graph with nominators and validators:
```
emresurmeli:
```suggestion
is uniformly spread out. An election algorithm for Nominated Proof of Staking (NPoS) on Polkadot will try to optimize three 
metrics when computing a solution graph of nominators and validators:
```

docs/maintain/kusama/maintain-guides-how-to-validate-kusama.md
```diff
@@ -42,11 +42,11 @@ Riot.im) using [this link](https://matrix.to/#/#KusamaValidatorLounge:polkadot.b
 
 You can have a rough estimate on that by using the methods listed
 [here](../../general/faq.md#What-is-the-minimum-stake-necessary-to-be-elected-as-an-active-validator-?).
-Validators are elected based on [Phragmén's algorithm](../../learn/learn-phragmen.md). To be elected into the
-set, you need a minimum stake behind your validator. This stake can come from yourself or from
-[nominators](../../learn/learn-nominator.md). This means that as a minimum, you will need enough KSM to set up
-Stash and Controller [accounts](../../learn/learn-keys.md) with the existential deposit, plus a little extra for
-transaction fees. The rest can come from nominators.
+To be elected into the set, you need a minimum stake behind your validator. This stake can come from
```
emresurmeli, line 45: We should add the minimum stake amounts here, in the computed-dict.json I only see min_nominator_intention in KSM. So we will need to add the Polkadot amount as well.

profradha: Yeah, the minimum stake required to push a validator into the active set changes from era to era. What @alfarok is proposing here https://github.com/w3f/polkadot-wiki/issues/3533 might help render that value.

The minimum bond required by the validators on both Polkadot and Kusama is 0. That is why there are no variables for them in computed-dict.json I think.

emresurmeli (approved):

emresurmeli:
@DrW3RK Looks like there is a conflict in `maintain-guides-how-to-validate-kusama.md`

PR #3486, opened 2022-07-07
Remove "--unsafe-pruning" flag from the docs

vovacha:
```
// according to https://github.com/substrate/issues/8103;
warn!(
   "WARNING: \"--unsafe-pruning\" CLI-flag is deprecated and has no effect. \
   In future builds it will be removed, and providing this flag will lead to an error."
);
```


DrW3RK (approved):
Thank you 

Ref: https://github.com/paritytech/substrate/issues/8103

Issue #3448, opened 2022-06-28
Resolve multiple redirects are created with the same "from" pathname build errors


alfarok:
These errors can be seen when looking through the Polkadot or Kusama Staging build and deploy jobs.  The error is thrown several times stating **It is not possible to redirect the same pathname to multiple destinations**.

Console output of a single instance of the error:
```
Error:  @docusaurus/plugin-client-redirects: multiple redirects are created with the same "from" pathname: "/docs/en/build-build-with-polkadot"
It is not possible to redirect the same pathname to multiple destinations: 
- {"from":"/docs/en/build-build-with-polkadot","to":"/docs/build-build-with-polkadot"}
- {"from":"/docs/en/build-build-with-polkadot","to":"/docs/build-build-with-polkadot"}
```

alfarok:
Full log of conflicts:
```
[INFO] Duplicate routes found!
- Attempting to create page at /docs/build-build-with-polkadot, but a page already exists at this route.
- Attempting to create page at /docs/build-hackathon, but a page already exists at this route.
- Attempting to create page at /docs/build-oracle, but a page already exists at this route.
- Attempting to create page at /docs/build-pdk, but a page already exists at this route.
- Attempting to create page at /docs/build-smart-contracts, but a page already exists at this route.
- Attempting to create page at /docs/build-storage, but a page already exists at this route.
- Attempting to create page at /docs/build-tools-index, but a page already exists at this route.
- Attempting to create page at /docs/build-wallets, but a page already exists at this route.
- Attempting to create page at /docs/ens, but a page already exists at this route.
- Attempting to create page at /docs/learn-auction, but a page already exists at this route.
- Attempting to create page at /docs/learn-balance-transfers, but a page already exists at this route.
- Attempting to create page at /docs/learn-crowdloans, but a page already exists at this route.
- Attempting to create page at /docs/learn-governance, but a page already exists at this route.
- Attempting to create page at /docs/learn-identity, but a page already exists at this route.
- Attempting to create page at /docs/learn-parachains, but a page already exists at this route.
- Attempting to create page at /docs/learn-parathreads, but a page already exists at this route.
- Attempting to create page at /docs/learn-registrar, but a page already exists at this route.
- Attempting to create page at /docs/learn-treasury, but a page already exists at this route.
- Attempting to create page at /docs/maintain-errors, but a page already exists at this route.
[...]

Issue #411, opened 2020-04-28
Split out the Kusama guide


lsaether:
The proposal is to build the Kusama guide as a separate site from the Polkadot wiki. One idea is to put two docusaurus sites that both build from the `docs` directory but had different configurations. The mirror script would then work across both sites. 

Could be combined with #369 to solve the issue of where to host the second site. 

Need to be careful not to break CrowdIn during the process.

Swader:
Closing because of #612 et al being done

PR #3489, opened 2022-07-07
Consolidate mirror logic and append 'mirror-' to page slugs

alfarok:
### Description
This PR addresses #3448 

It was determined that the duplicate route errors we were seeing were due to the implementation of `mirror.mjs`.  This file runs during builds to created duplicate pages to avoid some [restrictions of Docusaurus](https://github.com/w3f/polkadot-wiki/blob/master/scripts/mirror.mjs#L4).  The issue is that these duplicate mirrored pages share the same `slug` as the original pages.

The top of each markdown doc contains some metadata.  One of the properties in this metadata is the `slug`, which determines the route that will be used for that page.  `mirror.mjs` updates the `id` property only and keeps the route the same.  This is the source of the errors, 2 pages now share the same `slug`.

The router detects this, generating one set of errors and the redirect code throws again for the same reason (100+ lines of errors in console).  Since these pages share the same `slug`, whichever page is compiled later gets rendered for that route.  This isn't a big issue because the content is the same but it will always trigger the console errors.

### Current Solution
Append the same 'mirror-' prefix to the slug of the mirrored page.  All errors go away but mirror pages now have their own routes.  For example:

`docs/learn-crowdloans` and `docs/mirror-learn-crowdloans` would both be valid routes.

~~It's a bit strange to have both pages.  For example, how does this play with analytics.~~
^After looking into this more, I don't think this is a big issues.  The mirror pages are only ever used for Kusama content.  So for example, a Polkadot route in the docs could be 

`https://wiki.polkadot.network/docs/build-pdk`

while the Kusama route would be:

`https://guide.kusama.network/docs/mirror-build-pdk`

but

`https://wiki.polkadot.network/docs/mirror-build-pdk`

and 

`https://guide.kusama.network/docs/build-pdk`

would technically exist but never be referenced in the docs so a user would have to manually type out these type of routes.  For the purposes of analytics, it's very low likelihood that both routes would get regular hits and cause any confusion.  The `mirror-` prefix could be changed to `kusama-` for our purposes but then `https://wiki.polkadot.network/docs/kusama-build-pdk` would exists too, although never referenced.

### Alternative
Turn-off duplicate page errors in the `docusaurus.config.js` files via the `onDuplicateRoutes` property and still deal with the errors surfacing again at the [redirect code](https://github.com/w3f/polkadot-wiki/blob/master/polkadot-wiki/docusaurus.config.js#L84).  This risk here is hiding real errors.

**With either solution we should merge this PR as it consolidates `mirror.mjs`.  If we don't want the additional routes, the `else if` condition in the `BuildMirrors` function can be removed.**

### DIFF
--- a/scripts/mirror.mjs
+++ b/scripts/mirror.mjs
@@ -1,163 +1,106 @@
-import fs from "fs";
+import { existsSync, readFileSync, writeFileSync} from "fs";
 
-/**
- * The wiki hosts both the Polkadot Wiki and the Kusama User Guide. This means
- * that we can "mirror" the same document between the two. However, due to
- * a peculiarity in Docusaurus, when we put the same document into both sub-wikis
- * in `website/sidebars.json` only the most recent entry is read.
- *
- * To solve this, we create an explicit mirror of the document. The process for
- * adding content to both wikis is as follows:
- *
- * 1) Create the source document and put it under the relevant sidebar in the
- *    Polkadot wiki configuration in `sidebars.json`.
- * 2) Add the source document's name to the `mirrored` array below.
- * 3) Run `node mirror.js`.
- * 4) Add `mirror-DOCUMENT_NAME` to the sidebar for Kusama in `sidebars.json`.
- *
+/*
+  The wiki hosts both the Polkadot Wiki and the Kusama User Guide. This means
+  that we can "mirror" the same document between the two. However, due to
+  a peculiarity in Docusaurus, when we put the same document into both sub-wikis
+  in `website/sidebars.json` only the most recent entry is read.
+
+  To solve this, we create an explicit mirror of the document. The process for
+  adding content to both wikis is as follows:
+ 
+  1) Create the source document and put it under the relevant sidebar in the
+     Polkadot wiki configuration in `sidebars.json`.
+  2) Add the source document's name to the `mirrored` array below.
+  3) Run `node mirror.js` which duplicates the original markdown file and
+     appends 'mirror-' to the file's id and slug to prevent duplicate routes.
+  4) Add `mirror-DOCUMENT_NAME` to the sidebar for Kusama in `sidebars.json`.
  */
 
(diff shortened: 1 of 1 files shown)


emresurmeli (approved):

alfarok:
I noticed the sidebar isn't showing on the routes with the `mirror-` prefix.  Looking into it now.  

@emresurmeli & @DrW3RK Do you guys have a preference on the current solution vs the alternative I outlined in the PR description?  There is a trade-off with either option.

DrW3RK:
I have not run this on my machine yet. @emresurmeli What was the output at your end?

DrW3RK:
I don't think I have the bandwidth to review this PR today, but to give you a backdrop of where this all started https://github.com/w3f/polkadot-wiki/issues/411 

Also, search for "mirror" in the repo's issues and PRs and you'll find the work that has been put into this

alfarok:
No rush at all, I just want to discuss it a little before merging since it can be handled a couple different ways with different repercussions.

I know some of these infrastructure related PRs aren't very exciting but we almost have all the errors/warnings from the build process handled.  It should make it a lot easier to catch new bugs and issues earlier moving forward.

alfarok:
I spent a little more time trying to understand how the mirrored pages are currently used.  I now think we can remove the mirror functionality completely.  I should have tried this before attempting to salvage the existing system.  I realized some changes between Docusaurus v1 and v2 that I think eliminated the need for this work around.  The history you pointed me to @DrW3RK shows this was implemented when we were still using v1.

TLDR: removing all the mirror code and removing any `mirror-` prefixes from existing links doesn't seem to be an issue or break anything.  Despite the cleanup effort here, I am thinking about closing this PR and opening another that removes mirrors and all references to mirrors.

Let me know what you guys think.  In general, I think the more legacy code we can remove the better.  The more lean the repo is, the easier it is for a new potential contributor to get familiar with the code.

DrW3RK:
Yeah, sounds good to me!

PR #3500, opened 2022-07-11
add info on needed reserve for opening crowdloan

brenzi:
If this explicit info isn't welcome because it has to be maintained, there should be info on how to find out. The error just says insufficient funds, and I had to browse the Events of previous crowdloans to find the Balances:reserved Event to learn how high the deposit must be


DrW3RK (approved):
Thank you @brenzi

PR #3512, opened 2022-07-11
Remove legacy mirroring functionality

alfarok:
After the previous research conducted in #3489, it seems like the mirroring functionality is no longer required to maintain both the Polkadot and Kusama docs from the same code-base.

I tested these changes locally for both docs and didn't see any issues with the navigation at the updated urls.


kusama-guide/docusaurus.config.js
```diff
@@ -48,10 +48,6 @@ module.exports = {
         docs: {
           showLastUpdateAuthor: true,
           showLastUpdateTime: true,
-          editUrl: ({ docPath }) => {
-            const githubDocPath = `https://github.com/w3f/polkadot-wiki/edit/master/docs/`;
-            return githubDocPath + docPath.replace("mirror-", "");
-          },
```
alfarok, line 54: This line was actually removing the 'mirror-' prefix anyway to point at the same original doc, not the mirror.

DrW3RK (approved):
Looks good to me!

PR #3515, opened 2022-07-12
Update deploy-kusama-staging.yml

DrW3RK:


DrW3RK:
not very useful. closing it

PR #3520, opened 2022-07-12
Update build-parachains.md

lisa-parity:
The link to the "parachain tutorials" had the wrong file name and was throwing a 404. The tutorials are still under construction, though, so maybe it makes sense to remove that link to docs.substrate.io for now.


DrW3RK (approved):
Thanks! We will monitor the tutorials section and update the links when they are live

Issue #3511, opened 2022-07-11
Substrate links update script may be invalid


DrW3RK:
https://github.com/w3f/polkadot-wiki/blob/master/scripts/update-substrate-links.mjs

This script should replace substrate rustdoc links in the md files for the latest version of substrate.
Looks like a lot of changes happened since this script has been written and now http://substrate.dev/ redirects to https://docs.substrate.io/quick-start/ and http://substrate.dev/docs redirects to https://paritytech.github.io/substrate/master/sc_service/

This script needs to be revisited as it may no longer be relevant

DrW3RK:
This script runs as a part of CI check, which can be disabled if there is no use.

alfarok:
Yeah I don't think this is doing much in the current state.  On my last PR it logs the following:
```
NOTE: 'monthly-2022-07' and variants of its prefixes are not valid Substrate rustdoc version numbers. No links to update!
```

Link verification is already happening (to some extent) during the build process within Docusaurus [here](https://github.com/w3f/polkadot-wiki/blob/master/polkadot-wiki/docusaurus.config.js#L45-L46).  It could be valuable to have a CI check that verifies no broken links but it probably shouldn't prevent merging as it could block PR's that are unrelated to the failures.  When I completed the audit a few months ago I saw some intermittent failures but it turned out they were caused by 3rd party links being temporarily down but not actually broken when I returned say the next day.  

Maybe there is a way to performs these type of checks once a day and report failures on riot.  Something like [scheduled events](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#scheduled-events-schedule) on GitHub actions.

Also like you pointed out in the description with the current redirects, it doesn't look like a version number is included in the url's anymore for `https://docs.substrate.io` or `https://paritytech.github.io/substrate/master/` (later always points to master).  So hopefully these updates will be required less frequently.  They are usually pretty easy to update with a VS Code search and replace if required.

My proposal would be to remove [update-substrate-links.mjs](https://github.com/w3f/polkadot-wiki/blob/master/scripts/update-substrate-links.mjs) and [check-substrate-release.yml](https://github.com/w3f/polkadot-wiki/blob/master/.github/workflows/check-substrate-release.yml).  If we feel like this check is a priority to have, look into adding scheduled events.

@DrW3RK let me know your thoughts and I can start moving on this one.

DrW3RK:
[...]

PR #3540, opened 2022-07-18
Remove legacy substrate link updater (not active)

alfarok:
This PR addresses #3511.

Although the job is still active, the current code will not actively find or update any links when run.


DrW3RK (approved):
About time! Although, we should keep an eye on the Substrate docs through the link checker

Issue #3370, opened 2022-05-28
Update Cross-Consensus Messaging Content


salmad3:
- XCM Page doesn't really explain XCM under the hood; material is meant to be advanced
- Breakdown the tech stack
- More specifics: cross-consensus protocols as transport protocols, multilocation, etc.
- Should provide real examples

emresurmeli:
Things to add:
- [ ] XCM high level blurb
- [ ] XCM analogies and philosophy
- [ ] XCMP
- [ ] VMP
- [ ] DMP
- [ ] UMP
- [ ] XCVM
- [ ] Encoding rules

PR #3554, opened 2022-07-20
Add XCM v2 updates

emresurmeli:
Targeting #3370 this PR is a general revamp of /learn-crosschain page to be /learn-xcm, and include all updates/changes to naming and function of XCM. 

Things to add:
- [x] XCM high level blurb
- [x] XCM analogies and philosophy
- [x] XCMP
- [x] VMP
- [x] DMP
- [x] UMP
- [x] XCVM


docs/learn/learn-cross-consensus.md
```diff
@@ -4,7 +4,7 @@ title: Cross-Consensus Message Format (XCM)
 sidebar_label: Cross-Consensus Message Format (XCM)
 description: Learn about the messaging format at the forefront of interoperability.
 keywords: [cross-consensus, XCM, XCMP, interoperability, communication]
-slug: ../learn-crosschain
+slug: ../learn-xcm
```
profradha, line 7: There are articles in circulation with the Polkadot Wiki links. Just ensure that they are not broke by implementing appropriate redirects, if you are changing the slugs.

emresurmeli: I've updated in all places in the wiki. I think even though it might break some links out there in the wild, the rename is necessary.

profradha: The suggestion was to implement redirects. Not to stop appropriate renaming the link.
https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects

We do not know how many websites have linked our Wiki article, and as a rule of thumb, we need to ensure every linked Wiki doc is accessible.

image

emresurmeli:
Rebased mater into branch early on - starting a new branch and PR to keep things clean #3736

Issue #3536, opened 2022-07-15
Start leveraging new React components to replace existing global constants or other dynamic values


alfarok:
This issue builds off of #3533 which was addressed in #3535.

The first step was to propose a implementation and this task involves leveraging the previous work to start replacing the existing constants or add new constants that have greater short-term ranges with these new components.

@DrW3RK suggested to start with these values:
```
Min Active Nomination (very useful for users landing on our Staking page)
Min stake backing an active validator in an era (resourceful for new node operators eyeing our ecosystem)
Number of parachains on Polkadot/Kusama
number of active nominators and Total number of nominators
number of active validators/ waiting validators
era number
Staked DOT/KSM percentage
```

alfarok:
I started working on this by initially trying to swap out as many global constants as possible for components.  The work-in-progress branch can be found [here](https://github.com/w3f/polkadot-wiki/compare/%233536?expand=1).

alfarok:
Addressed with #3557

PR #3557, opened 2022-07-21
Use React components to retrieve chain state values

alfarok:
### Description

This PR addresses #3536.

Historically the project stores a series of commonly referenced values as global constants for reference throughout the project. 
 These global constants were calculated at the time that the project was built/deployed.  This means if the Polkadot Wiki or Kusama Guide is not frequently redeploy, the values have the potential to become stale.  This isn't a big deal if the constant has a relatively small short-term range in value but is much more apparent for values that have larger ranges.

The new solution is to use inline React components instead which allow us to either fetch values when each page is navigated to (on component mount) or we can subscribe to the chain and update them in real-time (see #3535).  I created an example for the later but have not yet done this for the existing values as it would be overkill based on their short-term ranges.  @DrW3RK has already suggested several areas where the real-time updates would be very impactful (coming soon).

So for example, when you navigate to any given page the values are retrieved once asynchronously from the live state and still fallback to defaults if required.  If you leave a page and the values happens to change on-chain, they will be updated when you navigate back and does not require refreshing the web app as it occurs on-mount.

### Features
- [x] Uses React components to pull live data instead of referencing hard-coded dictionary values
- [x] Several global constants had hard-coded values are now retrieved from the chain instead
- [x] Filters are consistently applied to default values and retrieved values (`humandReadable` & `blockToDays`)
- [x] Improved error handling
- [x] Adds a new component for the `Voluntary Locking Table` (can't render a React component inside an existing markdown table - this also serves as a good example for implementing more custom content in the future that may be outside the scope of what traditional markdown is capable of)
- [x] Removes a bunch of legacy code

### Formatting Notes
- React components should be imported at the top of each markdown file under the header
- When using conditional rendering it is best to try and put each component on a newline for clarity
  - The one exception to this is when the component does not have a trailing space before the next character.  Because of how the markdown is parsed a newline will result in a space being added.  So, when this is the case both components must be on the same line.  (I will point out an example in the code below for clarity)
- When not using conditional rendering placing a component inline is less overwhelming
- Try to re-use existing components as much as possible instead of making a new component for every task.  This PR using one component for retrieving every global constant (except the custom table)

### Testing
As I replaced each previously existing global constant reference, I tried to navigate to that page both locally, at `wiki.polkadot.network` and at `guide.kusama.network` to make sure the values matched their previous state on the old system.  In some cases I found units (DOT or KSM) were duplicated because of how the human readable filter was applied to defaults vs retrieved values and addressed the issue.  There were also several cases were Polkadot values were used for both Polkadot and Kusama docs which is also now fixed.

### All build errors are now gone for both Polkadot and Kusama builds! 🎉
image


components/Voluntary-Locking.jsx
```diff
@@ -0,0 +1,112 @@
...
+		// can't be used to render a table (can't put a <table> in a <p>).
+		// So, we use the same component for Polkadot and Kusama and figure it out here.
+        const title = document.title;
+		if (title === "Governance · Polkadot Wiki") {
+			updateTable("polkadot")
+		} else if (title === "Governance · Guide") {
+			updateTable("kusama");
+		} else {
+			console.log("Unknown wiki/guide type");
+		}
```
alfarok, line 95: This took me way too long to find a solution :sweat_smile:.  I can further explain what the issue was if desired.  There also may be a better way to check which doc is active but this works for now.

docs/general/glossary.md
```diff
@@ -361,16 +362,17 @@ online, even if they haven't published a block this epoch. This is sometimes ref
 ## Origin
 
 The initiator of an extrinsic. A simple origin would be the account that is sending a token to
-another account. Polkadot also supports more complex origin types, such as the [root origin](#Root Origin), from
+another account. Polkadot also supports more complex origin types, such as the [root origin](#root-origin), from
```
alfarok, line 365: broken link

docs/learn/learn-crowdloans.md
```diff
@@ -76,7 +77,8 @@ runtime upgrades (as determined through its own local governance).
 :::info Minimum Crowdloan Contribution
 
 The minimum balance for contributions for a crowdloan campaign is 
-currently set to {{ crowdloan_min_contribution }}. 
+currently set to 
+{{ polkadot: <RPC network="polkadot" path="consts.crowdloan.minContribution" defaultValue={50000000000} filter="humanReadable"/> :polkadot }}{{ kusama: <RPC network="kusama" path="consts.crowdloan.minContribution" defaultValue={100000000000} filter="humanReadable"/> :kusama }}.
```
alfarok, line 81: Here is an example of what I mention in the description.  Both conditionally rendered components need to be on the same line otherwise the trailing `.` would have a space added for Polkadot only.

docs/learn/learn-governance.md
```diff
@@ -246,19 +248,12 @@ votes = tokens * conviction_multiplier
...
-| :----------: | :-------------: | :------------: |
-|      0       |       0.1       |     {{ vote_locking_period_0 }}  |
-|      1       |        1        |     {{ vote_locking_period_1 }}  |
-|      2       |        2        |     {{ vote_locking_period_2 }}  |
-|      4       |        3        |     {{ vote_locking_period_3 }}  |
-|      8       |        4        |     {{ vote_locking_period_4 }}  |
-|      16      |        5        |     {{ vote_locking_period_5 }}  |
-|      32      |        6        |     {{ vote_locking_period_6 }}  |
-
+<VLTable network="polkadot"/>
```
alfarok, line 251: Here is the custom `Voluntary Locking Table` component.  Required because you can't load a component inside of the markdown table.

docs/learn/learn-proxies.md
```diff
@@ -106,9 +107,9 @@ Learn more about anonymous proxies from our
 ### Time Delayed Proxies
 
 We can add an additional layer of security to proxies by giving them a delay time. The delay will be
-quantified in number of blocks (blockNumber). Polkadot and Kusama both have
-{{ block_target_in_seconds }} second blocks, hence a delay value of 10 will mean 10 blocks which
-will equal 1 minute of delay. The proxy will announce it's intended action and wait for the number
+quantified in number of blocks (blockNumber). Polkadot and Kusama both have 6
```
alfarok, line 110: I was having trouble finding how to retrieve this value on-chain and asked about how to calculate it in `W3F Dumb questions and answers`.

docs/maintain/kusama/maintain-guides-how-to-validate-kusama.md
```diff
@@ -46,7 +46,7 @@ To be elected into the set, you need a minimum stake behind your validator. This
 yourself or from [nominators](../../learn/learn-nominator.md). This means that as a minimum, you will 
 need enough KSM to set up Stash and Controller [accounts](../../learn/learn-keys.md) with the existential 
 deposit, plus a little extra for transaction fees. The rest can come from nominators. To understand how 
-validators are elected, check the [NPoS Election algorithms](../learn/learn-phragmen.md) page.
+validators are elected, check the [NPoS Election algorithms](../../learn/learn-phragmen.md) page.
```
alfarok, line 49: broken link

package.json
```diff
@@ -1,17 +1,15 @@
 {
   "scripts": {
     "kusama:clean": "shx rm -rf node_modules/.cache; docusaurus clear kusama-guide",
-    "kusama:start": "yarn fetch --isPolkadot=false; docusaurus start kusama-guide",
-    "prekusama:build": "npm-run-all -ln kusama:clean -p \"fetch --isPolkadot=false\" kusama:copy-assets --aggregate-output",
```
alfarok, line 5: As far as I could tell these `pre:build` commands aren't used but please let me know if this is not the case.

scripts/injectPlugin.js
```diff
@@ -97,27 +73,17 @@ const conditionalReplace = (node, file, options) => {
...
   };
 }
 
 function injectPlugin(param) {
-  const dictPath = path.resolve(__dirname, "computed-dict.json");
-  const dict = JSON.parse(fs.readFileSync(dictPath, "utf8"));
   return [
     transform,
-    { dict, isPolkadot: param.isPolkadot, dryRun: false, verbose: false, debug: false },
+    { isPolkadot: param.isPolkadot, dryRun: false, verbose: false, debug: false },
```
alfarok, line 86: all the changes in this file basically remove the old functionality that swapped in the dictionary values but I left the functionality for conditional rendering.

components/Voluntary-Locking.jsx
alfarok: I was thinking about this a little more and had another thought.  For the future we could add a small metadata section to the Polkadot Wiki and Kusama Guide that contains a couple variables we could verify to avoid this hack.  It would also allow us to remove all the duplicate components that use `{{ polkadot/kusama: :polkadot/kusama}}`.  The single component could complete the check internally.  Not a priority, but something to think about down the line for less code and more clarity.

docs/general/faq.md
```diff
@@ -109,7 +110,7 @@ However, increasing validators above one thousand remains a goal for later itera
 It is also worth mentioning that one thousand validators is more than the number of validators of
 similar PoS chains with comparable levels of economic security as Polkadot. The closest contenders
 are operating with around 150 validators, while Polkadot is already securely running with
-{{ num_validators }}.
+<RPC network="polkadot" path="query.staking.validatorCount" defaultValue={150}/>.
```
profradha: Formatting seems a bit off

image

docs/learn/learn-crowdloans.md
profradha: Got it

docs/learn/learn-governance.md
profradha: Yeah, I am aware of this issue. Not a problem.

docs/learn/learn-governance.md
```diff
@@ -246,19 +248,12 @@ votes = tokens * conviction_multiplier
...
 
 The maximum number of "doublings" of the lock period is set to 6 (and thus 32 lock periods in
-total), and one lock period equals {{ vote_locking_period }} days. 
+total), and one lock period equals 
+{{ polkadot: <RPC network="polkadot" path="consts.staking.maxNominations" defaultValue={403200} filter="blocksToDays"/> :polkadot }}
```
profradha:
```suggestion
{{ polkadot: <RPC network="polkadot" path="consts.democracy.voteLockingPeriod" defaultValue={403200} filter="blocksToDays"/> :polkadot }}
```

docs/learn/learn-governance.md
```diff
@@ -246,19 +248,12 @@ votes = tokens * conviction_multiplier
...
+{{ kusama: <RPC network="kusama" path="consts.staking.maxNominations" defaultValue={115200} filter="blocksToDays"/> :kusama }} days.
```
profradha:
```suggestion
{{ kusama: <RPC network="kusama" path="consts.democracy.voteLockingPeriod" defaultValue={115200} filter="blocksToDays"/> :kusama }} days. 
```

docs/learn/learn-proxies.md
profradha: Alright. The solution mentioned can be explored in a different PR

docs/learn/learn-teleport.md
```diff
@@ -66,9 +67,10 @@ Here are a couple of things that you need to understand before submitting a tele
 
 :::caution Existential Deposit
 
-  After the teleport, if your account balance drops below the Existential Deposit which is 
-{{ existential_deposit }}, that balance will be burnt and lost. Either plan on teleporting all of 
-  your balance or leave enough balance such that the account does not get reaped.
+After the teleport, if your account balance drops below the Existential Deposit which is 
+{{ polkadot: <RPC network="kusama" path="consts.balances.existentialDeposit" defaultValue={33333300} filter="humanReadable"/> :polkadot }}{{ kusama: <RPC network="kusama" path="consts.balances.existentialDeposit" defaultValue={33333300} filter="humanReadable"/> :kusama }},
```
profradha:
```suggestion
{{ polkadot: <RPC network="polkadot" path="consts.balances.existentialDeposit" defaultValue={10000000000} filter="humanReadable"/> :polkadot }}{{ kusama: <RPC network="kusama" path="consts.balances.existentialDeposit" defaultValue={33333300} filter="humanReadable"/> :kusama }},
```

DrW3RK (changes requested):
Added a few suggestions and fixes. Might need another round of review

docs/learn/learn-proxies.md
alfarok: Just for reference 

This was the solution provided by Philipp:
```
Iirc the blockheader metadata has a timestamp
```

joe elaborated with:
```
timestamp.now.at(<block_hash>)

Or if you want it from a block (instead of chainstate) then you can just match on the set timestamp extrinsic

{
  "method":{
    "pallet":"timestamp","method":"set"
  },
  "signature":null,
  "nonce":null,
  "args":{"now":"1658382702006"}
}

I don't think the block time is in chainstate anywhere (perhaps under some BABE config, but I don't have time to look at the moment). But it's easy to get empirically from (time(B) - time(A)) / (B - A)
```

alfarok:
> Added a few suggestions and fixes. Might need another round of review

Thank you!  Working through the updates now.

docs/learn/learn-teleport.md
alfarok: Great catches!

DrW3RK (approved):

Issue #3551, opened 2022-07-19
Remove all console (runtime) errors from project


alfarok:
Currently there are 2 real console errors that can be seen when building/running locally or in prod.  They don't appear to be causing any major issues but we should get them fixed to help identify newly introduced errors.

1) `GET https://apisa.web3.foundation/latest.js net::ERR_BLOCKED_BY_CLIENT`
2) `Uncaught SyntaxError: Identifier 'e' has already been declared (at addressChanger.js:25079:15)`

![image](https://user-images.githubusercontent.com/13341935/179791997-6eda9acc-279a-4a9f-931d-ec540ab66e8f.png)

alfarok:
The 2nd error comes from this file: https://github.com/w3f/polkadot-wiki/blob/master/polkadot-wiki/static/js/packaged/addressChanger.js

`e` is param in the function and redefined as a const within the function.  I need to look more into what this file does.  There are 2 files with the same name (`addressChanger.js`) with different content.  They also seem to get copied during the build process.  It a little hard to make any changes in the file listed above because it looks like it was auto-generated elsewhere.  Most variables are a single letter and its a large file.

![image](https://user-images.githubusercontent.com/13341935/180272144-a186011f-79ce-4820-8da1-c4f6f8181185.png)

alfarok:
The first error comes from what looks like an old analytics script.  Not sure if it is still intended to be active yet, but it is definitely failing in the current state.  The script is imported [here](https://github.com/w3f/polkadot-wiki/blob/master/polkadot-wiki/docusaurus.config.js#L26) for Polkadot and [here](https://github.com/w3f/polkadot-wiki/blob/master/kusama-guide/docusaurus.config.js#L27) for Kusama but never referenced anywhere else.

Removing it doesn't seem to have an impact but I want to double check with the team to make sure there is no intended use.

DrW3RK:
I thought we decomissioned addresschanger.js

That can be removed completely

PR #3563, opened 2022-07-22
Resolve remaining run-time errors that populate in console

alfarok:
### Description

This PR addresses #3551.

As long as `https://apisa.web3.foundation/latest.js` isn't intended to be doing anything (doesn't look like it has worked for a while) this PR should be good to go.

There are still references to address checker in the translated docs, which to my knowledge are no longer officially supported to be in sync.  A lot of the address changer code was already removed prior to this PR but a few files we left behind.  This means any of these references in the translated docs should already be broken prior to this removal.

### TODO

- [x]  Double check the status of `https://apisa.web3.foundation/latest.js` with dev ops team


alfarok:
Confirmed with Philipp that the analytics reference I am removing is legacy no longer used.  We now use fanthom analytics instead.

DrW3RK (approved):

Issue #3544, opened 2022-07-18
Audit and remove all unused files from the repo


alfarok:
The team has recently removed several legacy files and scripts that are no longer used or required but there appears to be several more that can also be removed.

For example, at a quick glance it doesn't look like `removeUneeded.mjs` and `converter.rb` are referenced anywhere else in the repo to name a few.

This helps keep the repo lean and less complex for new contributors.

alfarok:
also possibly `computed.mjs`?

alfarok:
This should be pretty minimal after #3557 merges but there are a couple more that weren't related to that content so I will address them after separately to avoid confusion.

alfarok:
Addressed in #3573.

PR #3573, opened 2022-07-25
Remove additional legacy files

alfarok:
This PR addresses #3544.

There may be a few more files that can be removed but these are the most obvious as they are no longer referenced at all or are in a state that they no longer produce a usable result.  I would propose we start a new unit testing harness that runs on the CI automatically to replace the intentions of `generateErrorTable.mjs` (see #3574).


DrW3RK (approved):
Need to figure out a better way to generate the error table

Issue #3574, opened 2022-07-25
As a developer, I would like a unit testing solution that runs automatically on the CICD pipeline


alfarok:
The solution to this issue involves creating a test harness that runs automatically on the CICD pipeline for new PRs.

An exhaustive set of tests is not required for this but something we can work towards as we add new content or update existing content.  Anytime a PR involves fixing something that "broke" a unit test should be added to cover the issue and help prevent future cases.

Examples:
- Ensure RPC values are successfully being retrieved
- Check new submissions for broken links
- Are console errors being generate for an updated page
- Etc.

alfarok:
addressed with #3576

PR #3576, opened 2022-07-26
Add jest testing to CI

alfarok:
### Description
This PR addresses #3574.

image

- [x] Add Jest testing framework to project and CI
- [x] Add several test samples covering the new RPC component
- [x] Fix a couple bugs in the RPC component discovered from the new tests

### TODO
- [ ] Add section on Jest testing to README


components/RPC-Connection.jsx
```diff
@@ -51,10 +51,16 @@ function RPC({ network, path, defaultValue, filter=undefined }) {
...
+				const newValue = await syncData(network, path, setReturnValue);
+				if (newValue === undefined) {
+					 // There was an issue with the request, use default
+					return;
+				} else if (filter !== undefined) {
+					// Apply filter to retrieved value if a filter is provided
+					applyFilter(newValue, filter, network, setReturnValue)
+				} else {
+					// Apply value as-is
+					setReturnValue(newValue);
```
alfarok, line 63: This was actually causing an edge case where some 'const' values without filters were always using the default value even though a value was successfully being retrieved from the chain.  I thought the test framework wasn't working at first but it turned out the test was correct.  Unit testing already paying off. :partying_face:

package.json
```diff
@@ -66,8 +72,8 @@
   "dependencies": {
     "@algolia/client-search": "^4.14.1",
     "@mdx-js/runtime": "^1.6.22",
-    "@polkadot/util": "^9.6.2",
-    "@polkadot/util-crypto": "^10.1.1",
+    "@polkadot/util": "10.1.1",
```
alfarok, line 74: All these dependencies should align in version

tests/rpc.test.js
```diff
@@ -0,0 +1,128 @@
...
+import RPC from "../components/RPC-Connection";
+import { ApiPromise, WsProvider } from "@polkadot/api";
+
+// Set max test duration before failing (2 min)
+jest.setTimeout(120000);
+
+test("Retrieves and applies a 'const' RPC value", async () => {
+	render(<RPC network="polkadot" path="consts.balances.existentialDeposit" defaultValue={0} />);
+	await waitFor(() => expect(screen.getByText("10000000000")).toBeInTheDocument(), { timeout: 2000 });
+});
```
alfarok, line 13: I want to try loading the entire app and write more generic tests that can potentially find errors in the docs, but I will attempt that in another PR.

DrW3RK (approved):

DrW3RK (changes requested):
holding this from merging. Please check why this fails

alfarok:
The tests are passing again with no changes.  This makes me think that it is timing related with the delay to receive the RPC response.  I am going to **bump the timeout from 2 seconds to 5 seconds** to see if that helps.  The tests will still complete faster than that if the data is available but wait a little longer if it's not yet.

If any of the tests prove to be more annoying that useful with intermittent failures we can remove or adjust further accordingly.

DrW3RK:
Makes sense. We can merge now

DrW3RK (approved):

alfarok:
I didn't push the change yet haha but I'll open in a new PR since it's minimal.

PR #3580, opened 2022-07-28
Activate SimpleAnalytics and remove Fathom

DrW3RK:


DrW3RK:
@Nexus2k I think these changes should do. wdyt?

Nexus2k (approved):

Issue #3491, opened 2022-07-08
Move Polkadot Open Stack doc to the Wiki from Grants repo


DrW3RK:
Add this doc (or split it into multiple docs) to the Wiki

https://github.com/w3f/Grants-Program/blob/master/docs/polkadot_stack.md

The badges are generated through https://github.com/w3f/status-badges

alfarok:
There is some type of formatting issue with this page that docusaurus doesn't like.  When I copy the raw markdown directly into a new page on the wiki I get this formatting error:

![image](https://user-images.githubusercontent.com/13341935/181782184-a8440840-a08d-4f65-a1fc-5144a3d42494.png)

Looking through it now to see if there is a missing closing tag somewhere but it is a little challenging to quickly see in such a large table.

alfarok:
Got it, there are 4 instances of `<>` used in link titles.  The markdown renders fine as-is in the standard GitHub flavor but not in JSX.  The `<>` gets read as the opening/closing of an element tag causing the rest of the doc to shift tags.  I will replace these with a `-`.

alfarok:
addressed in #3584

PR #3584, opened 2022-07-29
Move Polkadot Open Stack doc to the Wiki from Grants repo

alfarok:
Addresses #3491

This PR accomplished the following:
- [x] Migrates the Open Stack doc from Grants to the Polkadot Wiki repo (build-open-source.md)
- [x] Adds the new page the the Polkadot Wiki and Kusama Guide + sidebars
- [x] Leverages update status badges GitHub action managed by w3f


alfarok:
We are hitting some kind of rate limiter on GitHub actions so I can't finish testing this yet

alfarok:
Having a hard time testing this one.  It makes many requests to the GitHub public API so running it too many times in a short period of time results in hitting the rate limiter.  I am still working on sorting out the following error:
```
Validation error creating new PR (existing PRs will be force pushed to): Validation Failed: {"resource":"PullRequest","field":"base","code":"invalid"}
```

DrW3RK:
The current setup refreshes every week. We probably should schedule this workflow to run that way.

alfarok:
> The current setup refreshes every week. We probably should schedule this workflow to run that way.

I agree but it makes it challenging to debug since I can only run it once and have to wait a while for the API rate limiter.

Everything is ready to go except for the badge updating.  A `422` error is being returned when attempting to open the pull request.  Still looking into the source.

alfarok:
Just reached out to @lucasvanmol via Riot to see if he has any feedback before I dig too deep into debugging.  Debugging actions hosted in other repos is tricky.  The error comes from the result of [this try/catch block](https://github.com/lucasvanmol/status-badges/blob/main/index.js#L93).  The [GitHub Token](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) is generated by actions so it should have the correct permissions to open the PR.  Maybe there is an additional setting that needs to be updated on the repo to permit actions from opening a PR?

alfarok:
Based on the error code, it seems like the `base` branch being set is invalid.  This looks like it is programmatically set in the action and isn't a parameter we can provide.  Something unexpected could be happening [here](https://github.com/w3f/status-badges/blob/main/index.js#L66) maybe?

```js
const base = (await git.branchLocal()).current;
```

I need to modify the index.js file in the source repo to get more information regarding what is wrong with the base provided as the `422` is too vague.

lucasvanmol:
I haven't seen a 422 error like this - although I know it happens when a PR already exists (hence the message about force pushing). I've added extra debug logging to `w3f/status-badges@v1.1` (you shouldn't need to update the tag) - you can try running the workflow again with a debug logging enabled, either by setting the secret ACTIONS_STEP_DEBUG to true in the repository that contains the workflow, or by re-running the workflow through the UI and checking "Enable debug logging".

Regarding rate limiting - I agree it can get annoying because the document contains a lot of links, so you might want to run it on a smaller version of the file with only a handful of them just for testing purposes. IIRC the limit is 1000 requests per hour and the doc contains 350+ links, so getting rate limited means you have to wait an hour for it to reset.

alfarok:
Thanks @lucasvanmol, this is very useful!  I will dig more into debugging today.

alfarok:
Hey @DrW3RK, after some conversation with @lucasvanmol today, we think the issue could be related to the action being triggered by the open pull request instead of the cron job.  In testing, the new branch is successfully created but it fails to open the pull request.  The new branch also contains all the PR code since it's not merged yet which could be another source of the issue.

I am thinking lets try to merge it and see what happens when the cron job runs.
Worst case nothing happens, the new branch get created but no PR open.  We can revert or turn-off the job until its resolved.
Best case it works and we are done.

~~I sped up the cron job to happen once per hour to test.  Once it's confirmed failing or working I will update accordingly.~~

lucasvanmol:
Fyi, you won't need to wait for the cron job in order to test it. With `workflow_dispatch` you can run the action manually from the actions tab in the github UI

lucasvanmol:
fwiw it works in my fork: 

The run:
https://github.com/lucasvanmol/polkadot-wiki/actions/runs/2795155863
created the PR properly: https://github.com/lucasvanmol/polkadot-wiki/pull/1

README.md
```diff
@@ -146,6 +146,8 @@ To format markdown pages, run the following in the `docs` folder:
 
 ```bash
 npx pretty-quick --staged
+or
+npx pretty-quick --branch
```
profradha, line 150: What is this doing here :)

DrW3RK (approved):
Very cool!

PR #3603, opened 2022-08-04
Run Jest tests weekly instead of on PRs

alfarok:
This PR puts the Jest tests on a cron job instead of the PR trigger.  This makes more sense for the type of tests we have in the suite right now.  I hope we can add more general tests in the future and can revisit this topic.  The tests can also be invoked manually from the actions UI if desired for testing a PR:

image

 It also comments-out one long running test.  In general Jest encourages writing short running operations.  Running over a minute starts to get a bit excessive/flaky.


DrW3RK (approved):
Looks good to me!

PR #3604, opened 2022-08-05
Update build-wallets.md

gmauboussin-ledger:
Because Ledger Live now supports the Polkadot blockchain, it will be more exhaustive to add Ledger Live in the list of compatible wallet


DrW3RK:
Thank you! We are planning a separate section in the Wiki doc to add references to wallets like Ledger live. Will keep you posted.

DrW3RK:
As this section only lists the wallets officially supported by https://support.polkadot.network/support/home we will close the PR for now. We plan to create a page or section listing third-party wallets and will include this suggestion.

Issue #3371, opened 2022-05-30
Conditional Render - Nomination Pools to Kusama Guide


DrW3RK:
With Nomination pools going live on Kusama, perform conditional render for https://wiki.polkadot.network/docs/learn-nomination-pools on Kusama guide

alfarok:
addressed in #3625

PR #3625, opened 2022-08-09
Add Nomination Pool chain data preview for Kusama Guide

alfarok:
Addresses #3371 

Adds the following on-chain data to the Nomination Pools doc:

image

image

image


DrW3RK (commented):
I really like the idea of having dashboards in the Wiki docs showing key on-chain information. Have a couple of comments regarding the information presented. 

1. Can we abridge this further by showing, Pools - 59/64 etc. rather than sticking to the table format? The reason I mention this is because the first two rows are comparative and then the next two rows are not. 
2. Not having it in table format lets us expand on the descriptions. For instance, the min create bond and minPoints to balance ratio are not very intuitive. We could explain what they are and display the values.
3. Add Markdown links to the sections in the Wiki document that explains the terminology in the dashboard

Let me know your thoughts. I am okay with presenting the info in tabular format, but just being nit-picky that it can be restricting in the ways mentioned above.

alfarok:
> 

Thanks @DrW3RK, this is great feedback.  I also felt iffy on the table format.  Left me make another pass over this today and see if I can come up with an alternative layout.

DrW3RK (approved):
Looks great!

docs/learn/learn-nomination-pools.md
```diff
@@ -135,6 +137,12 @@ vertical dots and click on Withdraw unbonded.
 - In order for a member to switch pools they must wait for the normal 28 day unbonding process.
 - A member can partially unbond the staked funds in the pool (at most 16 partial unbonds).
 
+:::info Kusama Pool Sizing
```
profradha:
```suggestion
:::info Kusama Pool Stats
```

PR #3626, opened 2022-08-09
Add minStake info for Validators

DrW3RK:
Populate minStake of validator in the active set. Default to historic values while loading.

image

### DIFF
--- a/components/Minimum-Stake.jsx
+++ b/components/Minimum-Stake.jsx
@@ -28,9 +28,9 @@ async function CalcValidatorMinStake(defaults, setReturnValue) {
   const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[0])
   let validatorMinStake = parseInt(validatorStake['total'].toString())
 
+  // TODO: this operation takes too long - try and see if we can batch the requests or use async
   // For all current validators
   for (let i = 1; i < currentValidators.length; i++) {
-    console.log(i);
     // Get the validators stake
     const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[i])
     const validatorTotalStake = parseInt(validatorStake['total'].toString())
@@ -40,8 +40,6 @@ async function CalcValidatorMinStake(defaults, setReturnValue) {
     }
   }
   const output = validatorMinStake.toString();
-
-  console.log(output);
 
   setReturnValue(output);
 }
--- a/components/Minimum-Stake.jsx
+++ b/components/Minimum-Stake.jsx
@@ -44,9 +44,8 @@ async function CalcValidatorMinStake(network, wsUrl, setReturnValue) {
     }
   });
   
-  const output = validatorMinStake.toString();
-  setReturnValue(output);
-  HumanReadable(output, network, setReturnValue);
+  const result = validatorMinStake.toString();
+  HumanReadable(result, network, setReturnValue);
 }
 
 export default MinimumStake;
--- a/components/Minimum-Stake.jsx
+++ b/components/Minimum-Stake.jsx
@@ -7,7 +7,6 @@ function MinimumStake({ network, defaultValue }) {
   const [returnValue, setReturnValue] = useState('');
 
   useEffect(async () => {
-    console.log(defaultValue);
     // Set defaults based on network
     let wsUrl = undefined;
     if (network === "polkadot") { wsUrl = "wss://rpc.polkadot.io" }
@@ -18,7 +17,7 @@ function MinimumStake({ network, defaultValue }) {
     HumanReadable(defaultValue, network, setReturnValue);
     // Calculate a more accurate approximation using on-chain data
     await CalcValidatorMinStake(network, wsUrl, setReturnValue);
-  });
+  }, []);
 
   return (returnValue);
 }
@@ -36,18 +35,17 @@ async function CalcValidatorMinStake(network, wsUrl, setReturnValue) {
   const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[0])
   let validatorMinStake = parseInt(validatorStake['total'].toString())
 
-  // TODO: this operation takes too long - try and see if we can batch the requests or use async
-  // For all current validators
-  for (let i = 1; i < currentValidators.length; i++) {
-    // Get the validators stake
-    const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[i])
-    const validatorTotalStake = parseInt(validatorStake['total'].toString())
-    // Compare against current minimum
+  // Iterate era validators
+  const validators = await api.query.staking.erasStakers.entries(currentEra.toString());
+  validators.forEach(([key, validator]) => {
+    const validatorTotalStake = parseInt(validator.total);
     if (validatorTotalStake < validatorMinStake) {
       validatorMinStake = validatorTotalStake;
     }
-  }
+  });
+  
   const output = validatorMinStake.toString();
+  setReturnValue(output);
   HumanReadable(output, network, setReturnValue);
 }
 
--- a/components/Minimum-Stake.jsx
+++ b/components/Minimum-Stake.jsx
@@ -15,6 +15,24 @@ const KusamaDefaults = {
   currentEra: 4058
 }
 
+function MinimumStake({ network }) {
+  const [returnValue, setReturnValue] = useState('');
+
+  useEffect(async () => {
+    // Set defaults based on network
+    let defaults = undefined;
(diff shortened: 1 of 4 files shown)


components/RPC-Connection.jsx
```diff
@@ -103,6 +103,7 @@ async function syncData(network, path, setReturnValue) {
 		const wsProvider = new WsProvider(wsUrl);
 		let api = await ApiPromise.create({ provider: wsProvider });
 
+
```
alfarok: extra newline

components/RPC-Connection.jsx
```diff
@@ -117,6 +118,32 @@ async function syncData(network, path, setReturnValue) {
 				chainValue = api.toString();
 				break;
 			case "query":
+				switch (pathParameters[2]) {
```
alfarok: I don't think we should rely on this condition at such a low level in the RPC component.  Every time the component is instantiated elsewhere in the docs and references a `validators` query, this logic will run.  That being said I think all the logic below here is great.  I think moving this to it's own component makes the most sense to keep the RPC component as general as possible.

I made a confusing comment yesterday about adding parameters to the switch statement.  I thought you were trying to use the RPC component to make a query that required you to pass input parameters to the function, such as 

```js
api.query.getBalanceFor(SOME_WALLET_ID)
```

Right now we don't support this because there is no way to pass `SOME_WALLET_ID`.  So I was thinking you were trying to extend the RPC component to accept these input parameters, such as 

```js
 <RPC network="polkadot" path="api.query.getBalanceFor" parameters="[SOME_WALLET_ID]"  /> 
```

If you had this functionality you could then use several `<RPC />` components in conjunction to do the calculations below, such as

```js
a = <RPC network="polkadot" path="api.query.session.validators"  /> 
b = <RPC network="polkadot" path="api.query.staking.currentEra"  /> 
c = <RPC network="polkadot" path="api.query.getBalanceFor" parameters="[b, a]"  />
```

This sudo code could be accomplished by creating a new parent component specific to this functionality, that wraps the child components outlined above.  When all the children have valid values you could raise an event that notifies the parent that the display value is ready to be updated.

```js
<MinimumStake>
    <RPC network="polkadot" path="api.query.session.validators"  /> 
    <RPC network="polkadot" path="api.query.staking.currentEra"  /> 
    <RPC network="polkadot" path="api.query.getBalanceFor" parameters="[b, a]"  />
</MinimumStake>
```

Now that I see what you meant, I wouldn't worry about the approach I just described.  It would probably be a pain to synchronize the child components, especially if some have to wait for others to complete before running etc.  I would just create a new component and copy the logic from below into it.

The RPC component is good for retrieving results from a single endpoint and has the ability to post-process the results using a filter function.  Anything that is more complex probably requires a new component, even if that new component wraps or extends RPC through inheritance.

Sorry for the long-winded reply but I wanted to try and explain my thinking from yesterday and why it probably pushed you in this direction :grin:   Good conversation for our meeting today!

components/RPC-Connection.jsx
```diff
@@ -117,6 +118,32 @@ async function syncData(network, path, setReturnValue) {
...
+						  api.query.staking.currentEra(),
+						]);
+
+
+						const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[0])
+						let validatorMinStake = parseInt(validatorStake['total'].toString())
+						
+
+						for (let i = 1; i < currentValidators.length; i++) {
+							const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[i])
```
alfarok: How many validators does this execute for?  This could be a big bottleneck if there are many validators, as they are synchronous requests.

We don't have an example of this yet but you could fire them all off async and wait for say an array of results to have the expected number of items (or time-out) before proceeding.  Still might not be super fast but it would save considerable time.

components/RPC-Connection.jsx
```diff
@@ -172,6 +199,9 @@ function applyFilter(value, filter, network, setReturnValue) {
 		case "blocksToDays":
 			value = (value * 6) / 86400;
 			break;
+		case "lowStaked":
+			value = (value * 6) / 86400;
+			break;
```
alfarok: I don't think this filter is being used?

alfarok (commented):
Hey @DrW3RK, the calculation logic here looks great.  My comments are mainly around where this code should live.  I added some overly verbose comments :laughing:, building off our discussions from yesterday.  Let's chat about it in our 1:1 today and decide the best place for it.

components/RPC-Connection.jsx
profradha: Yeah, makes sense. I'll move this to a dedicated component.

components/RPC-Connection.jsx
profradha: We have 1K validators in the active set on Kusama and 297 on Polkadot. That is why I intend to refresh the default values every month or so to ensure it reflects the trend with min stake.

components/RPC-Connection.jsx
profradha: Yeah, clearly I did not delete the unused code :)

components/RPC-Connection.jsx
profradha: Could you try that mechanism to speed up things after I create a dedicated component?

alfarok: Yeah we can try running them async.  I'm wondering if the Polkadot JS api has a rate-limiter?  Making 1000+ requests each time the page loads could push that limit if it exists. 

I guess the alternative would be to cache these values in a json file in the repo and have a CI weekly action that updates the json (similar to the status-badge system).  Also not ideal considering we are trying to keep the values more up-to-date.  If we start finding many similar application where large requests are being made for a more complex calculation, a system like that may make more sense.

alfarok: @DrW3RK This also might be useful for retrieving batches of data in a single request: https://polkadot.js.org/docs/api/start/api.query.other#map-keys--entries

components/Minimum-Stake.jsx
```diff
@@ -0,0 +1,52 @@
+import { useState, useEffect } from "react";
+import React from "react";
+import { ApiPromise, WsProvider } from "@polkadot/api";
+import { HumanReadable } from "./utilities/filters";
+
+function MinimumStake({ network, defaultValue }) {
+  const [returnValue, setReturnValue] = useState('');
```
alfarok, line 7: This is the new component

components/Minimum-Stake.jsx
```diff
@@ -0,0 +1,52 @@
...
+    // Set default value to render on component
+    HumanReadable(defaultValue, network, setReturnValue);
+    // Calculate a more accurate approximation using on-chain data
+    await CalcValidatorMinStake(network, wsUrl, setReturnValue);
+  }, []);
+
+  return (returnValue);
+}
+
+async function CalcValidatorMinStake(network, wsUrl, setReturnValue) {
```
alfarok, line 25: this is basically all the logic migrated from the switch statement

components/Minimum-Stake.jsx
```diff
@@ -0,0 +1,52 @@
...
+    api.query.staking.currentEra(),
+  ]);
+
+  // Get validators stake for current error and first validator
+  const validatorStake = await api.query.staking.erasStakers(currentEra.toString(), currentValidators[0])
+  let validatorMinStake = parseInt(validatorStake['total'].toString())
+
+  // Iterate era validators
+  const validators = await api.query.staking.erasStakers.entries(currentEra.toString());
+  validators.forEach(([key, validator]) => {
```
alfarok, line 40: In this single api call we can get all the values that were being fetched from the 1000+ calls in the previous iteration.  For more details, check this out: https://polkadot.js.org/docs/api/start/api.query.other#map-keys--entries

Even though we still iterate all the validators, considerable time is saved because the bulk of the wait time was coming from waiting on the synchronous web requests to complete.

components/utilities/filters.js
```diff
@@ -0,0 +1,54 @@
+const Polkadot = "polkadot";
```
alfarok, line 1: Moved filters to an independent js module so they can be used by components other than just the RPC

alfarok (commented):
@DrW3RK I think this should be working as expected as an independent component now.  It also only takes a couple seconds to sync with the on-chain data.  Please double-check that the calculation logic is correct & that the default and calculated values are valid for both networks.

components/Minimum-Stake.jsx
profradha: Cool! Thanks for the info

DrW3RK (commented):
Looks good to me! Thanks

PR #3629, opened 2022-08-10
Updated voting Councillor voting constants

paradox-tt:
I updated the base amounts for both Kusama and Polkadot using the constant phragmenElection.votingBondBase .  To avert users from encountering insufficient funds errors I rounded up.  

As an addition, I've added amounts for each vote sourced from phragmenElection.votingBondFactor.  I tried to keep the significant numbers to the same degree but in some cases the numbers were too round to allow for this.

### DIFF
--- a/docs/maintain/maintain-guides-how-to-vote-councillor.md
+++ b/docs/maintain/maintain-guides-how-to-vote-councillor.md
@@ -16,8 +16,8 @@ councillors in the elections.
 ## Voting for Councillors
 
 Voting for councillors requires you to reserve 
-{{ polkadot: ~20 DOT :polkadot }}{{ kusama: ~0.063 KSM :kusama }} 
-(a base amount + a per vote amount). You can then bond whatever amount you wish to put 
+{{ polkadot: ~20.064 DOT :polkadot }}{{ kusama: ~0.067 KSM :kusama }} as a base amount
+and an amount per vote {{ polkadot: ~0.032 DOT :polkadot }}{{ kusama: ~0.00011 KSM :kusama }}. You can then bond whatever amount you wish to put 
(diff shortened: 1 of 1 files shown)


laboon:
Looks good, but we probably should pull these numbers off-chain directly if we can. @alfarok is that possible?

DrW3RK:
Yes @laboon Adding the required code for it

paradox-tt:
Oh, so we can extract these from the chain constants, is this something that the public can do as well?

DrW3RK:
Yes @paradox-tt We added this brand new feature to the Wiki. Thanks to @alfarok  :)

https://github.com/w3f/polkadot-wiki#inline-react-components

DrW3RK (approved):

paradox-tt:
Excellent, I'll make use of it.  May I make a suggestion? It would be nice to have a parameter for significant figures.

DrW3RK:
> It would be nice to have a parameter for significant figures.

Could you elaborate?

Now, the values are fetched directly from the chain state and are rendered like below -

image
image

paradox-tt:
I was thinking some flexibility to round to 4 decimal places or in some cases 4 significant figures etc but this works.

alfarok:
Hey @paradox-tt, check out the `humanReadable` filter on the RPC component:

https://github.com/w3f/polkadot-wiki/blob/master/components/RPC-Connection.jsx#L153

It is how Polkadot or Kusama values are currently formatted after being pulled off-chain.

```js
<RPC network="polkadot" path="consts.identity.fieldDeposit" defaultValue={660000000} filter="humanReadable"/>
```

Let me know if you have any additional questions about the usage or changes.

paradox-tt:
Hi @alfarok,

Thanks for sharing and great work on this.  I understand the mechanism now.

PR #3671, opened 2022-08-19
Remove "undetected" when discussing the Monero inflation bug

kayabaNerve:
The presence of torsion is detectable, and the Monero blockchain has 
been confirmed to have not been exploited. This statement, as-is, casted 
doubt onto the current legitimacy of its monetary status with a faulty 
premise.

I do not speak Chinese and used Google Translate to remove characters in 
a similar location until Google Translate no longer said "undetected". I 
cannot confirm I did it properly, nor can I confirm the Chinese is 
correct in general. Google Translate claims it's all present tense, 
despite this being a historical footnote (and a valuable one at that).

### DIFF
--- a/polkadot-wiki/translated_docs/ar/learn-keys.md
+++ b/polkadot-wiki/translated_docs/ar/learn-keys.md
@@ -61,7 +61,7 @@ But ultimately the benefits of using Schnorr signatures outweigh the tradeoffs, 
 
 ### What is `sr25519` and where did it come from?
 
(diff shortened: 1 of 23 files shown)


laboon:
Thanks @kayabaNerve , you are correct. It's undetectable unless you know to look for it, but at this point in time Monero has been checked and re-checked for this specific bug.

DrW3RK (approved):

Issue #3346, opened 2022-05-16
Migration of how-to guides in the General and Basics sections to the Support page


michalisFr:
We can use this issue to find and settle on the how-to guides in the wiki that need to migrate to the Support page and track the progress of the migration. I'll update this post with a task list for each guide as we find them.

- [X] Find all how-to guides in the General and Basics wiki sections that need to migrate to the support page
- [X] Settle on the list of articles to migrate below
- [x] Prepare (create/update) all articles below
- [x] Remove/update corresponding wiki articles

### Articles to migrate
---
- [x] [Polkadot claims](https://wiki.polkadot.network/docs/claims): The whole article can be removed. Update the support article with steps of the claim process. Remove account creation material and instead link to relevant articles. (Low priority) ✅
---
- [X] [Using the Polkadot Ledger Application](https://wiki.polkadot.network/docs/learn-ledger): I propose this article be removed entirely from the wiki (the devs release isn't really supposed to be used). 

Each section has/needs its own article basically:
- [X] Ledger Live: [Article](https://support.polkadot.network/support/solutions/articles/65000175822-how-to-use-polkadot-and-stake-with-ledger-live) exists ✅
- [X] Loading your account: Articles exist for the [UI](https://support.polkadot.network/support/solutions/articles/65000170812-how-to-add-ledger-account-through-the-polkadot-js-ui) and the [extension](https://support.polkadot.network/support/solutions/articles/65000175387-how-to-add-your-ledger-through-the-polkadot-js-extension) ✅
- [x] Confirming the address on your device: We need to **create** this article. **[New article](https://support.polkadot.network/support/solutions/articles/65000181854-how-to-confirm-your-account-address-on-your-ledger-device)** ✅
- [X] Checking the balance: [Article](https://support.polkadot.network/support/solutions/articles/65000169332-where-can-i-see-the-balance-of-my-account-) exists (for all types of accounts) ✅
- [X] Sending a transfer: Article exists. **Update** with instructions on how to sign based on the specific account type (see Balance Transfers below). **[New article](https://support.polkadot.network/support/solutions/articles/65000181994)** ✅
- [x] Receiving a transfer: We need to **create** an article for types of accounts. **[New article](https://support.polkadot.network/support/solutions/articles/65000181866-how-to-receive-dot-to-my-account-on-polkadot-js-ui)** ✅
[...]

PR #3698, opened 2022-08-24
Remove migrated support content for statemine/statemint

alfarok:
### Description
This PR addresses the `Statemint` removal task outlined in https://github.com/w3f/polkadot-wiki/issues/3346.

### Actions
- [x] Links to new support content after general introduction
- [x] `Creating assets` should remain on the wiki
- [x] `Transferring assets` removed with link to new support page displaying same process and risks
- [x] Remove images associated with removal

### Support Articles Referenced
- https://support.polkadot.network/support/solutions/articles/65000181800-what-is-statemint-and-statemine-and-how-do-i-use-them-
- https://support.polkadot.network/support/solutions/articles/65000181118-how-to-transfer-tether-usdt-on-statemine


filippoweb3 (approved):
Looks good to me :)

docs/general/kusama/kusama-statemine.md
```diff
@@ -5,87 +5,80 @@ sidebar_label: Statemine
...
+Statemine is a generic assets parachain which provides functionality for deploying and transferring
+assets — both Fungible and Non-Fungible Tokens (NFTs). It is a common good parachain on Kusama (not
+to be confused with [Statemint](../../learn/learn-common-good-chains.md#statemint), which is a
+parachain on Polkadot). The native token of Statemine is KSM. Balance transfers and the Existential
+Deposit (ED) of KSM on Statemine are about 1/10th of the values on the Relay chain. For example, the
+Existential Deposit of a Statemine account is
+<RPC network="statemine" path="consts.balances.existentialDeposit" defaultValue={3333333} filter="humanReadable"/>,
+when compared to
+<RPC network="kusama" path="consts.balances.existentialDeposit" defaultValue={33333333} filter="humanReadable"/>,
+on Kusama. Apart from the core protocol token KSM, the assets held on Statemine can be broadly
```
alfarok: @filippoweb3 
I was curious how prettify would handle the RPC components.  This is going to cause an issue with formatting:

image

The behavior is pretty strange, both instance exhibit slightly different formatting (newline vs no space).  I need to spend a few minutes looking into it.  We can probably make a change on the RPC component to always force inline behavior, unless maybe a double newline is used.

filippoweb3: Good catch! Let me know if it is possible to fix it!

alfarok: Looks like the markdown parser expects the newline to start with a character, if it doesn't it assumes end of paragraph and adds a bottom margin to that paragraph:

image

If you put a character infront of `<RPC` it formats correctly.  I wonder if the same behaviour happens when using the `{{ polkadot: :polkadot }}` syntax.

alfarok: Since the formatting comes from the parent component, not the RPC component, I am hesitant to mess with any higher-level CSS.  A small change that works here could incorrectly format other content on the site.

One not so great option is to ignore the prettify command on blocks that contain RPC components such as:

image

It works but redundant and not ideal.  For now I am going to manually fix these couple occurrences and file another issue around finding a better long-term solution.  I am hoping there is a way to add a rule to prettier to ignore lines that contain HTML elements so we don't have to manually mark each occurrence.

PR #3623, opened 2022-08-09
Staking Wiki Doc - Reorg + writeup

filippoweb3:
Added Intro, and overview of NPoS

PR #3734, opened 2022-09-01
Remove Ambiguity on Staking Rewards calculation

DrW3RK:
Removing the statement that is ambiguous and adds no real contribution regarding staking rewards calculation.

This change should also be reflected in https://github.com/w3f/polkadot-wiki/pull/3623


filippoweb3 (approved):
Good to me 👍

PR #3781, opened 2022-09-12
Update Delegate Voting section

### LINKED ISSUE #3770: Delegate a Vote - Wiki section needs update
Update the screenshots in the article

> After you send the delegate transaction, you can verify it went through by navigating to the "Chain State" tab and selecting the "democracy" and "delegations" options.

This option seems to have disappeared and you can no longer see that option on Polkadot-JS UI

Update the section appropriately

DrW3RK:
We have to query `democracy.votingOf` now, which returns a JSON with both votes and delegations.


filippoweb3 (approved):
Looks good to me. I just run grammarly on the whole page and found a couple of minor grammatical errors. Let me know if it works!

Issue #3768, opened 2022-09-07
Update auctions pages to calculate estimated auction end times when page loads


alfarok:
### Description

Currently, all the information provided around auctions on [polkadot.network/auctions](https://polkadot.network/auctions/) & [kusama.network/auctions](https://kusama.network/auctions/) is hard-coded.  The auction end times are particularly an issue as these hard-coded values are just estimates that change moment to moment.  The more time that passes from the time the changes were authored, the greater likelihood the estimates are less accurate.

This content is to be moved to the Wiki/Guide and may be referenced by the original sources,  [polkadot.network/auctions](https://polkadot.network/auctions/) & [kusama.network/auctions](https://kusama.network/auctions/).

We can run these calculations when the given page loads either using the RPC component with a new filter for post-processing or by creating a new component (if that is a cleaner solution).

### Requirements

- [x] Create a react version of the migration tool that can be rendered inline on the wiki/guide ([#3785](https://github.com/w3f/polkadot-wiki/pull/3785))
- [x] Estimate block date/times based on the current chain state

alfarok:
Previous work on https://github.com/w3f/polkadot-wiki/pull/3065 may be useful for reference.

alfarok:
Another similar implementation for reference: https://kusama.subscan.io/block/14515200

alfarok:
This new content/table should be added to the wiki and will be mirrored on the original [polkadot.network/auctions](https://polkadot.network/auctions/) location.

alfarok:
First pass at the migration (#3785):

![auction-tool-wip](https://user-images.githubusercontent.com/13341935/191439335-8fee3765-663a-484e-a2c5-2a5dc18fa20f.gif)

alfarok:
Adding https://github.com/w3f/polkadot-wiki/pull/3814 as a follow-up task to fully automate this process end-to-end

PR #3785, opened 2022-09-13
Create react auction data component to replace existing externally hosted tool

alfarok:
### Description

This PR addresses #3768.

The task linked above calls for two main actions.  The first being the migration of the auction tool functionality from https://polkadot.network/auctions/ and https://kusama.network/auctions/ to a new react component that can be hosted on the wiki/guides in-line.  I wrote a scrapper that pulls the original hard-coded values from the original tool and now host this info in the `auctions.js` file in this PR.  The second step (which I already have some progress on) is pulling as much of this data as possible on-chain.  This PR covers steps 1.

### Previews

**Polkadot Original Tool:**
image

**Polkadot React Tool:**
image

**Kusama Original Tool:**
image

**Kusama React Tool:**
image

### Next Steps

As a follow-up, we should try to pull as much data from on-chain as possible and use the hard-coded estimates as a fallback/default (similar to how the RPC component treats defaults).  Pulling all this data, for all auctions on page load will likely not be very efficient so we should just pull the first selection from the drop-down live.  When the user changes the selection the next live retrieval should occur.  I will file a new issue for this work once this PR is approved and merged.


components/Auction-Schedule.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+		// TODO - estimates should only be made for future block, otherwise use on-chain timestamp
+		auction.startDate = EstimateBlockDate(date, currentBlockNumber, auction.startBlock);
+		auction.endPeriodDate = EstimateBlockDate(date, currentBlockNumber, auction.endPeriodBlock);
+		auction.biddingEndsDate = EstimateBlockDate(date, currentBlockNumber, auction.biddingEndsBlock);
+
+		// TODO - how to get his value?
+		auction.startOnBoard = "December 17th, 2021";
+		auction.endOnBoard = "October 20th, 2023";
+		//auction.endOnBoard = auction.weeksLeased * 7 + auction.startOnBoard;
```
alfarok: Hey @DrW3RK, I made some good progress on this and it renders exactly like the original hard-coded version however I am still a little confused on the timing of some auction events.

I am mainly confused regarding what values from the original table will always be hard-coded vs what can be calculated on-chain.  I made several assumptions that I know are incorrect, but want to pause and confirm which values can't be determined on-chain.  For example, I don't know what happens if not auctions are scheduled or how I could query that information on-chain.

I am also concerned that to generate all the auction estimates for say 25-50 auctions will be slow.  I definitely can optimize from what I am currently doing once I verify the logic but I am pretty sure it will still require at least a couple RPC invocations per auction (maybe can be done in parallel).

I am thinking we take a small step back and load the hard-coded values from json.  This will be 1:1 with the current tool but will now live on the wiki (step 1).  I will follow up with replacing all the possible on-chain values once I have a better understanding of different edge-cases (step 2).

Let me know what you think and I will continue working it out tomorrow.

profradha: Having a hardcoded auctions schedule on the Wiki for a start sounds good to me. You can implement step 1 in a PR (this one?) and then work on step2 later

alfarok: Yes exactly, I can get the hard-coded version prepared today.  Then I will outline the couple questions I have for our 1:1 or different meeting to finish part 2.  I think there may always be a couple hard-coded values so it will probably just involve removing a few properties from the json schema and pulling them on-chain async.

alfarok:
Continuing the on-chain data version on a [new branch](https://github.com/w3f/polkadot-wiki/compare/%233768-on-chain?expand=1) which includes all of these commits so this can be merged first if desired.

alfarok:
Closing this in favor of https://github.com/w3f/polkadot-wiki/pull/3814, which now includes steps 1 and 2.  The code has been refactored and condensed significantly from this version and makes estimates based on the current chain state.

PR #3790, opened 2022-09-14
Run "prettier" on all files in docs directory to achieve baseline state

alfarok:
This PR runs `npx prettier --write .` in the `/docs` sub-directory, which applies the same rules as `npx pretty-quick` but to unmodified files to achieve a baseline state.  This should ensure small PRs that run `npx pretty-quick` as a final step will not introduce large GitHub diffs, complicating what should be a simple review.  As long as we keep up with running the command with new PRs this should be a one-time effort.


docs/learn/learn-governance.md
```diff
@@ -7,8 +7,8 @@ keywords: [governance, referenda, proposal, voting, endorse]
 slug: ../learn-governance
 ---
 
-import RPC from "./../../components/RPC-Connection"
-import VLTable from "./../../components/Voluntary-Locking"
+import RPC from "./../../components/RPC-Connection" import VLTable from
+"./../../components/Voluntary-Locking"
```
profradha, line 11: This threw an error when rendering locally

```suggestion
import RPC from "./../../components/RPC-Connection" 
import VLTable from "./../../components/Voluntary-Locking"
```

DrW3RK (changes requested):

DrW3RK (approved):
Apart from the RPC component issue (already fixed), the rendering in all the changed docs looks good to me. Did a quick skim through.

filippoweb3 (approved):
Looks good to me :)

PR #3732, opened 2022-08-31
Apply prettier formatting to all docs containing RPC components ensuring rendering format is maintained

alfarok:
### Description

This PR seeks to address https://github.com/w3f/polkadot-wiki/issues/3704.

### Conclusion
**Anytime a RPC component is used in the docs, it should be wrapped in a conditional rendering statement**, such as:

```
{{ polkadot: <RPC network="polkadot" path="query.staking.validatorCount" defaultValue={297}/> :polkadot }}
{{ kusama: <RPC network="kusama" path="query.staking.validatorCount" defaultValue={297}/> :kusama }}
``` 

The RPC is always associated with a specific network so it makes sense that is should be rendered conditionally, accordingly.  This also helps prevent many of the formatting issues we were experiencing.  I was able to adequately fix every edge-case, ensuring the prettier command will not disturb the formatting.  We can now safely run prettier without worry about it incorrectly formatting existing RPC components.

Anytime a new component it added, prettier should be run and the author is responsible for making sure their syntax is correct (which shouldn't be modified when prettier runs) - see the README update in the PR for more details on this.

### Requirements
- [x] Address all existing instances by:
  a) Manually viewing every doc that contains any RPC components
  b) Running prettier
  c) Making sure rendered formatting is correct
  d) If formatting is modified or wrong, update the syntax to something stable by prettier standards
  e) Re-running prettier, ensuring no further changes to formatting have occurred on Polkadot or Kusama iterations
- [x] README should outline best practices for avoiding formatting issues

### Follow-up Tasks
- Run prettier on all docs
- Update [Gov2 PR](https://github.com/w3f/polkadot-wiki/pull/3605) by running prettier to simplify the diff against master

PR #3794, opened 2022-09-18
Update prettier configuration and restore running prettier on pre-commit hooks

alfarok:
This PR restores automated usage of the `npx pretty-quick --staged` command.  This means anytime a new commit is made, the command will automatically run forcing all new PRs to abide by the prettier standards but **only on markdown files**.

Since the entire docs section was already processed in #3790 this should not cause major changes when creating a new PR for legacy formatting.  All RPC were also properly formatted in #3732 so as long as new instances are wrapped in conditional rendering brackets and placed on newlines this shouldn't cause issues with formatting either.

Lastly, since #3790 only ran prettier within the `/docs` sub-directory, I re-ran the command at the root level with the new `.prettierignore` config as to ignored the unsupported translated markdown files.  The entire repo should now be up-to-date.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -6,14 +6,12 @@ labels: docs
 assignees: ""
 ---
 
-**Is your feature request related to a problem? Please describe.**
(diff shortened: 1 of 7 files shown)


filippoweb3 (approved):
Looks good to me 👍

Issue #3761, opened 2022-09-07
Update deposit values for Kusama


Imod7:
## Issue
Under the section [Making Transactions with a Multi-signature Account](https://wiki.polkadot.network/docs/learn-accounts#making-transactions-with-a-multi-signature-account) of the wiki, there is a table with the values of `DepositBase`and `DepositFactor` for the Polkadot and Kusama networks.
 
I was just double checking the values against Polkadot-js apps (_Developer > Chain State > Constants_) and it looks like the values for **Kusama** are different from what it is mentioned in the wiki.

### Deposit values for Kusama from p-js apps 
<img width="699" alt="Screenshot 2022-09-07 at 10 46 36" src="https://user-images.githubusercontent.com/12569221/188847808-dfe2368d-8707-468a-8c02-7b59ce068634.png">

### Deposit values for Kusama from the wiki
<img width="792" alt="Screenshot 2022-09-07 at 10 43 33" src="https://user-images.githubusercontent.com/12569221/188848041-6b7d2a37-f982-4775-a672-f40e25ac2930.png">

laboon:
Thanks for noticing. These were updated on Kusama last February on-chain but we missed updating this page.

@alfarok can you update this page to use on-chain data directly?

alfarok:
Addressed in #3797

PR #3797, opened 2022-09-19
Use on-chain values via RPC for deposit base/factor table

alfarok:
This PR address #3761.

Before:
image

After:
image

### DIFF
--- a/docs/learn/learn-accounts.md
+++ b/docs/learn/learn-accounts.md
@@ -489,14 +489,13 @@ console.log("DepositBase", deposit(1, 88));
 console.log("DepositFactor", deposit(0, 32));
 ```
 
-Thus the deposit values can be calculated as shown in the table below.
+Thus the deposit values can be calculated as shown in the table below. They are also shown
+in [plancks](learn-DOT.md#polkadot) for convenience.
 
 |          | Deposit Base                                                                                                    | Deposit Factor                                                                                                  |
 | -------- | --------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
-| Polkadot | <RPC network="polkadot" path="consts.multisig.depositBase" defaultValue={200880000000} filter="humanReadable"/> | <RPC network="polkadot" path="consts.multisig.depositFactor" defaultValue={320000000 } filter="humanReadable"/> |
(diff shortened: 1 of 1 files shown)


docs/learn/learn-accounts.md
```diff
@@ -489,10 +491,12 @@ console.log("DepositFactor", deposit(0, 32));
...
+| Polkadot | <RPC network="polkadot" path="consts.multisig.depositBase" defaultValue={200880000000} filter="humanReadable"/> | <RPC network="polkadot" path="consts.multisig.depositFactor" defaultValue={320000000 } filter="humanReadable"/> |
+| Polkadot | <RPC network="polkadot" path="consts.multisig.depositBase" defaultValue={200880000000}/> Planck                 | <RPC network="polkadot" path="consts.multisig.depositFactor" defaultValue={320000000 }/> Planck                 |
+| Kusama   | <RPC network="kusama" path="consts.multisig.depositBase" defaultValue={66959996400} filter="humanReadable"/>    | <RPC network="kusama" path="consts.multisig.depositFactor" defaultValue={106665600} filter="humanReadable"/>    |
+| Kusama   | <RPC network="kusama" path="consts.multisig.depositBase" defaultValue={66959996400}/> Planck                    | <RPC network="kusama" path="consts.multisig.depositFactor" defaultValue={106665600}/> Planck                    |
```
alfarok: This formatting is a bit overwhelming to look at but doesn't get reformatted by prettier which is good.  In the past we noticed you can't put tables in conditional rendering  so we are pretty much stuck with something that looks like this unless we create a custom react component just for this small table which seems like overkill.

@DrW3RK If you think the long line formatting is not adequate I can break it out to a custom component, let me know how you feel about it.  I also chose to transpose the table because the Deposit Base/Factor are the focus and the RPC filters provide the units automatically so I thought it look cleaner but it can also be reversed back if desired.

DrW3RK:
image

What do you think of this?

alfarok:
> image
> 
> What do you think of this?

I like it, looks better than the original and my option :+1: , will get this updated today.

DrW3RK:
Cool. I already pushed the commit. Let me approve this PR

DrW3RK (approved):

PR #3809, opened 2022-09-22
Fix ambiguous Ethereum Classic attack date

iTranscend:
Ethereum Classic's January 10, 2019 attack was ambiguously described to have happened on January 10, without indicating the year.

This PR adds the year of the attack so that readers don't think the attack happened in the current year the doc is being read.


alfarok (approved):
Thanks @iTranscend!

PR #29, opened 2019-09-30
Revamping the Maintain section

lsaether:
This pull request makes a major revamp to the maintain section that re-organizes it and updates it with Kusama information. The motivation of this revamp is to add a stronger narrative to the order of the pages in this section. The way I'm updating the validator guide is to maintain two pages, one for Alexander and one for Kusama.

- [X] Reorganizes the guides and information pages.
- [X] Splits and update the How To Validate guides.
- [X] Splits and update the How To Nominate guides.
- [X] Bring in the secure validator guide.

PR #3814, opened 2022-09-26
Create auction schedule component  that estimates block date/times based on chain data

alfarok:
### Description

This PR addresses #3768.

The task linked above calls for two main actions.  The first being the migration of the auction tool functionality from https://polkadot.network/auctions/ and https://kusama.network/auctions/ to a new react component that can be hosted on the wiki/guides in-line.  The second step was to make block date/time predictions based on the current chain stateon-chain.

sample-dot

### Strategy

In order to reduce latency the tool currently reads auctions blocks from the local `auctions.js` file.  When the page loads we only estimate the block date/times for the currently selected option in the drop-down.  When the user makes a new selection we connect to the chain and make an estimate for the new selected value.  If the user navigates back to a previously selected option we use a cache so it will be instant.  If the user refreshes or navigates away and back to the page all values will once again be retrieved live.  Attempting to make these RPC calls for 50+ auctions on every page load would be slow and expensive even async.

### Next Steps

#3814

Although we still rely on hard-coded block value, these values will not have to be frequently updated.  At this point it would still be the last manual step left in the workflow.  My proposal to remove this would be to add a new GitHub action that once a month checks for new future auction block, appends them to the existing auctions.js file and opens a new PR.  This would require the team to simply approve at max a once a month PR to keep the tool up-to-date far into the future.  Once a month may even be more than required.  I will file an additional task for the work if the team agrees.

Also pulling timestamps from finalized blocks on-chain  via PolkadotJS is slow, these values should probably also be cached and updated with the GitHub action.  Each time a drop-down is selected, if all blocks are finalized, 5 invocations will be made for data that does not change.  This is unnecessary overhead.


alfarok:
Hey @DrW3RK, I think this one is finally ready for review.  I am not thrilled with the latency required to retrieve all block timestamps for a given selection from on-chain.  If the block is already finalized it is probably a better idea to cache that info using the GitHub action in the backing JSON files.  I added more details around this to the description.

I also asked a few questions on this [here](https://github.com/polkadot-js/docs/issues/359) to the PolkadotJS team.

I don't want to spend too much more time nitpicking in this PR.  I think we can follow-up with some performance enhancements and enhancements around how future block data is handled but this should be a good starting point.  Make it work, make it work better.

Let me know if you have any questions.

components/Auction-Schedule.jsx
```diff
@@ -0,0 +1,260 @@
...
+		</a>
+		<hr />
+		<b>Auction Starts:</b>
+		<br />
+		{`${auctions[index].startDate} - `}
+		<a href={`${explorerUrl}${auctions[index].startBlock}`}>
+			Block #{auctions[index].startBlock}
+		</a>
+		<hr />
+		<b>Auction Ends:</b>
```
profradha: This is when bidding starts. It might be misleading to say Auction Ends. This is actually the end of the auction opening period when the bidding starts. https://wiki.polkadot.network/docs/learn-auction#mechanics-of-a-candle-auction

```suggestion
		<b>Bidding Starts:</b>
```

DrW3RK (changes requested):
Brilliant work!

I have a few suggestions to improve UX. Let me know how feasible it is to implement these suggestions

At the beginning of the Auctions Schedule section, we need to display the details of the current auction in this format. If there is an auction in progress, it should display

"Auction #28 is in progress on Polkadot which started on XXXX and ends on XXXX. For the full schedule of the auctions on Polkadot, use the dropdown below."

If there is no auction in progress, it should display

"Auction #29 on Polkadot will start on XXXX and ends on XXXX. For the full schedule of the auctions on Polkadot, use the dropdown below."

We can then eliminate the "Current Chain State" row from the displayed info.

Also, is it possible to make the default auction number in the dropdown to match with the current or next auction number?

components/Auction-Schedule.jsx
```diff
@@ -0,0 +1,260 @@
...
+		return auctions;
+	} else {
+		return (<div>Loading auction data...</div>)
+	}
+}
+
+// Loads drop-down selections
+async function LoadOptions(auctions, wsProvider) {
+	for (let i = 0; i < auctions.length; i++) {
+		const option = <option value={i} key={i}>{`Auction #${i + 1} at Block #${auctions[i]["startBlock"]}`}</option>
```
profradha: As we display the block numbers in the content. Makes it look less cluttered

```suggestion
		const option = <option value={i} key={i}>{`Auction #${i + 1}`}</option>
```

alfarok:
Thanks for the feedback @DrW3RK!  I think everything you outlined is possible.  The tricky aspect is we don't calculate the date/times for the blocks in the cache until they are selected.  One workaround/strategy might be iterating the `Bidding Ends:` block numbers until we find the first one that is in the future (compared to the current block).  From this object in the the JSON, check if the supplemental `Auction Starts:` block is also in the future.  If so, this is the next auction that will occur.  If not, this auction is currently taking place.  Let me know if that makes sense.  I will start experimenting with this approach.

alfarok:
> Thanks for the feedback @DrW3RK! I think everything you outlined is possible. The tricky aspect is we don't calculate the date/times for the blocks in the cache until they are selected. One workaround/strategy might be iterating the `Bidding Ends` block numbers until we find the first one that is in the future (compared to the current block). From this object in the the JSON, check if the supplemental `Auction Starts:` block is also in the future. If so, this is the next auction that will occur. If not, this auction is currently taking place. Let me know if that makes sense. I will start experimenting with this approach.

This returns `Auction #29 will be the next auction starting.` for Polkadot which seems to be accurate :+1: 

```js
function GetCurrentOrNextAuction(blockData, currentBlock) {
  for (let i = 0; i < blockData.length; i++) {
    // Still loading
    if (currentBlock === 0) {
      console.log(`Current block is still loading`);
      return;
    }
    // Check for active or next auction
    if (blockData[i].biddingEndsBlock > currentBlock) {
      if (blockData[i].startBlock > currentBlock) {
        console.log(`Auction #${i + 1} will be the next auction starting.`);
      } else {
        console.log(`Auction #${i + 1} is the current auction taking place.`);
      }
      return;
    }
  }
}
```

Will get it cleaned up with the formatting discussed above.  We can run it before setting the default selection, passing the resulting index.

alfarok:
@DrW3RK, this should be ready for another review.  The messaging and formatting should be in-sync with what we discussed above.  I am going to work on some additional caching and minor improvements to future block data as part of #3815.

Updated default loading state:
image

DrW3RK:
This seems incorrect. Also, for the current auction schedule, why can't we display the parachain onboarding date?

image

alfarok:
> This seems incorrect. Also, for the current auction schedule, why can't we display the parachain onboarding date?

Which part looks incorrect?  Here is similar data from https://parachains.info/auctions for auction #27.  The date/time will differ slightly based on the local computers timezone.  The times are also similar to https://polkadot.network/auctions/ but those are less accurate because they were predictions calculated a while ago, while the new values are based on finalized blocks or more much more recent 6 second block predictions.

image

If you are referring to the on-boarding start block being lower than the auction start block, I believe this is because the on-boarding period displays the maximum range and doesn't necessarily mean the selected auction was on-boarded at the start of that range (but within that range).  I kept this consistent with the [original auction tool](https://polkadot.network/auctions/).  For example, if you view auctions `#1 - #5` they all have the same range of `Fri Dec 17 2021 - Block #8179200 to Sun Oct 22 2023 - Block #17856000`.   I think the range is based on the full possible 96 week lease period but teams don't need to elect for the maximum.  Let me know if that makes sense.  Zach had a similar comment on this when I asked about the on-boarding range on the original tool.

image

I would have to do more research into the extrinsics to understand how to get these values on-chain using PolkadotJS.

In terms of the on-boarding data for the current auction or future auctions, it isn't displayed yet because you can't retrieve future block data from PolkadotJS AFAIK.  The latest auction began after I last manually updated the JSON files.  You can see [here](https://github.com/w3f/polkadot-wiki/blob/a5bd951915cc9b05bcd4f601af7fd426e47ca2b3/components/utilities/data/PolkadotAuctions.json#L369) the `startHash` is `0x0000000000000000000000000000000000000000000000000000000000000000` which indicates this was a future block at the time the JSON was generated.  Once we have the CI job this should be less of an issue.

DrW3RK:
Thanks! this is a misunderstanding on my part. Typically,  the winning parachains are onboarded a few weeks after they win the auction. But that onboarding schedule is up to the one who creates the proposal

I agree. Let us merge this PR and make improvements.

DrW3RK (approved):

DrW3RK:
I will make an update to the text 

"Winning parachain(s) onboarded" to "Lease period" to avoid ambiguity

alfarok:
Thanks for the review, all good questions!  This PR was getting pretty large so I think it's good we get it in before addressing everything.  If we are concerned with the speed or up-to-date accuracy at this point we can also hide the component from the markdown until I finish #3815.  Merging at least the JS in this PR will make the next review a lot less complicating.

PR #3845, opened 2022-09-29
New Banner implementation

je-boska:
@DrW3RK This implementation of the banner uses the built-in announcement bar feature of Docusaurus. The banner appears at the top of every page, and is closeable. Once closed, the closed state is saved to local storage, so the banner will not show again to the same visitor. Let me know if we prefer not to make the banner closeable.

I also deleted the previous banner implementation and all utilities and styling associated with it.


DrW3RK:
Thanks! Just tested it and it looks great! I think we should make the banner closeable.

Is it trivial to add it to https://guide.kusama.network/ as well? It is attracting increased traffic over the last few months. If it is not a priority, we can simple merge this.

je-boska:
I don't know, will look into it!

DrW3RK:
Kusma guide's config file https://github.com/je-boska/polkadot-wiki/blob/content/sub0-banner/kusama-guide/docusaurus.config.js

je-boska:
@DrW3RK This is almost the exact same banner, just removed the bottom border. Lmk if you're ok with keeping the same design. We could use the same implementation as https://kusama.network/ instead.

We'll probably want to create a unique UTM link for the Kusama guide, so hold off on merging until I have that.

DrW3RK:
Got it. Let us use the Kusama network implementation for the Kusama guide, for uniform styling.

je-boska:
UTM link is correct now, and styling is adjusted. For various reasons it's not 100% consistent with kusama.network, but it's pretty close.

DrW3RK (approved):

PR #3850, opened 2022-09-30
Minor edits to Intro to Staking

### LINKED ISSUE #3849: Minor edits Intro to Staking

filippoweb3:
Pretty quick caused a formatting issue at the end of the page, this has been reverted.

### DIFF
--- a/docs/learn/learn-staking.md
+++ b/docs/learn/learn-staking.md
@@ -476,7 +476,7 @@ nominal return rate will be higher, encouraging more users to use their tokens f
 contrary, when the system staking rate is higher than the ideal staking rate, the annual nominal
 return will be less, encouraging some users to withdraw.
 
-### Cons of Staking?
+### Cons of Staking
 
 - Tokens will be locked for about 28 days on Polkadot after unbonding, and seven days on Kusama.
 - Punishment in case of validator found to be misbehaving (see [slashing](#slashing)).
@@ -484,8 +484,7 @@ return will be less, encouraging some users to withdraw.
 
 ## How many Validators does Polkadot have?
 
-Polkadot currently has
-<RPC network="polkadot" path="query.staking.validatorCount" defaultValue={297}/> validators. The top
+Polkadot currently has <RPC network="polkadot" path="query.staking.validatorCount" defaultValue={297}/> validators. The top
 bound on the number of validators has not been determined yet, but should only be limited by the
 bandwidth strain of the network due to peer-to-peer message passing. The estimate of the number of
 validators that Polkadot will have at maturity is around 1000. Kusama, Polkadot's canary network,
--- a/docs/learn/learn-staking.md
+++ b/docs/learn/learn-staking.md
@@ -484,11 +484,12 @@ return will be less, encouraging some users to withdraw.
 
 ## How many Validators does Polkadot have?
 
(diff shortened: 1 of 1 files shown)


docs/learn/learn-staking.md
```diff
@@ -476,20 +476,20 @@ nominal return rate will be higher, encouraging more users to use their tokens f
...
-bound on the number of validators has not been determined yet, but should only be limited by the
-bandwidth strain of the network due to peer-to-peer message passing. The estimate of the number of
-validators that Polkadot will have at maturity is around 1000. Kusama, Polkadot's canary network,
-currently has 900 validator slots in the active set.
+{{ polkadot: Polkadot currently has <RPC network="polkadot" path="query.staking.validatorCount" defaultValue={297}/> validators :polkadot }}.
+The top bound on the number of validators has not been determined yet, but should only be limited by
+the bandwidth strain of the network due to peer-to-peer message passing. The estimate of the number
+of validators that Polkadot will have at maturity is around 1000.
+{{ polkadot: Kusama, Polkadot's canary network,
+currently has <RPC network="kusama" path="query.staking.validatorCount" defaultValue={1000}/> validator slots in the active set. :polkadot }}
```
alfarok: This will only display properly for Polkadot.  If you visit the same page for Kusama you will get the following formatting:

image

Every time you want to add an RPC component I would think about adding 1 for each network.  Most the time it makes sense because they will each point to different chains, `network="polkadot"` or `network="kusama"`.  In this case, you are pointing at the same chain for both given the context.  It may seem a little redundant but if we omit the conditional rendering brackets prettier will sometimes try and break up the component, causing it to fail.  Prettier ignores trying to format anything inside the `{{ }}`.

This should render properly for both and will not be modified by prettier:

```suggestion
Polkadot currently has
{{ polkadot: <RPC network="polkadot" path="query.staking.validatorCount" defaultValue={297}/> :polkadot }}
{{ kusama: <RPC network="polkadot" path="query.staking.validatorCount" defaultValue={297}/> :kusama }}
validators. The top bound on the number of validators has not been determined yet, but should only
be limited by the bandwidth strain of the network due to peer-to-peer message passing. The estimate
of the number of validators that Polkadot will have at maturity is around 1000. Kusama, Polkadot's
canary network, currently has
{{ polkadot: <RPC network="kusama" path="query.staking.validatorCount" defaultValue={1000}/> :polkadot }}
{{ kusama: <RPC network="kusama" path="query.staking.validatorCount" defaultValue={1000}/> :kusama }}
validator slots in the active set.
```

Let me know if you have any questions! :slightly_smiling_face:

alfarok (changes requested):

docs/learn/learn-staking.md
filippoweb3: I see; thank you! So basically, I use conditional rendering to avoid problems with pretty, even if I do not need it.

filippoweb3:
@alfarok, thank you for the help. Still need approval to merge the PR.

DrW3RK (approved):

DrW3RK:
Once @alfarok approves, this can be merged

docs/learn/learn-staking.md
alfarok: Exactly

alfarok (approved):

PR #3854, opened 2022-09-30
Update bagsList to voterList

### LINKED ISSUE #3751: bagsList pallet has been renamed voterList
https://wiki.polkadot.network/docs/learn-nominator#bags-list

This refers to the bagsList pallet (e.g. the extrinsics) but it has been renamed to voterList.

You can see a reference to it here: https://github.com/paritytech/polkadot/blob/b91e8efdde0a01981367cc18462d3613b1b8d3b7/runtime/kusama/src/lib.rs#L1059 and check it out via the extrinsics tab on Polkadot-JS UI, there is no bagsList.

There may be other references to bagsList that should be renamed to voterList elsewhere on the wiki.

DrW3RK:
Add note mentioning that voterList is an instance of Bags-List pallet

### DIFF
--- a/docs/learn/learn-nominator.md
+++ b/docs/learn/learn-nominator.md
@@ -69,7 +69,7 @@ occur often, so you should almost always see only a single active nomination per
 :::info
 
 On Polkadot and Kusama, the instance of the pallet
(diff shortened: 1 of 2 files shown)


filippoweb3 (approved):
Thanks Radha!

PR #3856, opened 2022-09-30
Add Caution about undelegation flaw

### LINKED ISSUE #3842: Add a cautionary note regarding Undelegation
Check this issue reported by @michalisFr https://substrate.stackexchange.com/questions/5067/delegating-and-undelegating-during-the-lock-period-extends-it-for-the-initial-am

Delegating and undelegating again within the lock period starts it over from the last undelegation.

emresurmeli:
For issue #3842

### DIFF
--- a/docs/maintain/maintain-guides-democracy.md
+++ b/docs/maintain/maintain-guides-democracy.md
@@ -285,9 +285,9 @@ through `democracy.votingOf` extrinsic.
 
 :::caution
 
-If there is an existing lock due to a previous change or undelegation, any new change or
+If there is an existing lock due to a previous delegation change or undelegation, any new change or
 undelegation will restart the lock period for the larger DOT amount and the longest conviction
-period, between the existing and the new lock.
+period between the existing and the new lock.
 
 This will only matter to accounts with conviction, as no conviction doesn't instill a lock period.
 
--- a/docs/maintain/maintain-guides-democracy.md
+++ b/docs/maintain/maintain-guides-democracy.md
@@ -285,9 +285,23 @@ through `democracy.votingOf` extrinsic.
 
 :::caution
 
-Currently, the undelegation logic has a flaw in it. If you choose to delegate and undelegate two
-times, and the second undelegation happens before the lock period of the initial delegation has
-ended. The initial undelegations lock period will be reset to the second undelegation lock period.
+If there is an existing lock due to a previous change or undelegation, any new change or
+undelegation will restart the lock period for the larger DOT amount and the longest conviction
+period, between the existing and the new lock.
+
+This will only matter to accounts with conviction, as no conviction doesn't instill a lock period.
+
+Examples:
+
+1. Delegate 500 DOT with 1x conviction, then change delegation to 1000 DOT with 1x conviction, the
+   lock period will reset for 1000 DOT with 1x conviction.
+
+2. Delegate 500 DOT with 3x conviction, then change the delegation to 1000 DOT with 1x conviction,
+   the lock period will reset for 1000 DOT with 3x conviction.
+
+3. Delegate 500 DOT with 1x conviction, then change the delegation to 200 DOT with 1x conviction,
+   the lock period will reset for 500 DOT with 1x conviction.
+
 To understand this in further detail checkout
 [this](https://substrate.stackexchange.com/questions/5067/delegating-and-undelegating-during-the-lock-period-extends-it-for-the-initial-am)
 stackexchange post.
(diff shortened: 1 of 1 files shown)


DrW3RK:
The formatting for the caution is off. Check https://docusaurus.io/docs/markdown-features/admonitions for the correct way to use admonitions.

image

docs/maintain/maintain-guides-democracy.md
```diff
@@ -283,6 +283,17 @@ through `democracy.votingOf` extrinsic.
 
 ## Undelegate a Vote
 
+:::caution
+
+Currently, the undelegation logic has a flaw in it. If you choose to delegate and undelegate two
```
profradha: In the scenario you mentioned, where someone delegates and undelegates, and then delegates and undelegates, it would not be a flaw to reset the lock period to the second undelegation.

I think the issue outlined is that the user first reduces the number of DOT they delegate. Say a user is delegating 500 DOT and wanted to reduce it to 200 DOT. They can issue an extrinsic to do that and the 300 DOT should start unlocking. During the period of unlock, the user chooses to undelegate the current voting power of 200 DOT. This resets the unlock period of the 300 DOT that was unlocking in the background.

@michalisFr Is my understanding correct?

michalisFr: Technically, when you reduce your delegation, the whole of the initial amount is set to unlock. If then, after the unlock period, you remove the lock, only the difference will become transferable because there is the lock of the current delegation.

So, if you delegate 500 DOT and then change it to 200 DOT, the lock set to expire is for the 500 DOT. When that lock expires and you remove it, 300 DOT will become transferable, because there is still a lock in existence for 200 DOT. It's not that when you do this you unlock only 300 DOT. 

The issue is that if you do this (the simplest case):
1. Delegate 500 DOT
2. Delegate 200 DOT at time t -> This creates a lock for 200 DOT and sets the lock for 500 DOT to expire at t + unlocking period`. So, the "overall" lock is for 500 DOT.
3. Undelegate at `t1 < t + unlocking period` (i.e. before the previous lock expired) -> Resets the lock for 500 DOT to expire at `t1 + unlocking period`

The expected behaviour would be that 500 DOT still unlocks at t + unbonding period and the 200 DOT at t1 + unbonding period.

Obviously, this only affects accounts that **delegated with a conviction**. 

Also, I wouldn't call this a "flaw". It's an edge case that wasn't considered.

DrW3RK (changes requested):
Please check the feedback above and update the caution

docs/maintain/maintain-guides-democracy.md
emresurmeli: > The formatting for the caution is off. Check https://docusaurus.io/docs/markdown-features/admonitions for the correct way to use admonitions.
> 
> image

Ah, this seems to be an issue with prettier:
If the closing colons are on the next line, it puts it at the end of the sentence.

docs/maintain/maintain-guides-democracy.md
```diff
@@ -283,6 +283,31 @@ through `democracy.votingOf` extrinsic.
 
 ## Undelegate a Vote
 
+:::caution
+
+If there is an existing lock due to a previous delegation change or undelegation, any new change or
+undelegation will restart the lock period for the larger DOT amount and the longest conviction
+period between the existing and the new lock.
+
+This will only matter to accounts with conviction, as no conviction doesn't instill a lock period.
```
profradha:
```suggestion
This will only matter to accounts with conviction, as the accounts with no conviction don't have any lock period.
```

docs/maintain/maintain-guides-democracy.md
```diff
@@ -283,6 +283,31 @@ through `democracy.votingOf` extrinsic.
...
+
+2. Delegate 500 DOT with 3x conviction, then change the delegation to 1000 DOT with 1x conviction,
+   the lock period will reset for 1000 DOT with 3x conviction.
+
+3. Delegate 500 DOT with 1x conviction, then change the delegation to 200 DOT with 1x conviction,
+   the lock period will reset for 500 DOT with 1x conviction.
+
+To understand this in further detail checkout
+[this](https://substrate.stackexchange.com/questions/5067/delegating-and-undelegating-during-the-lock-period-extends-it-for-the-initial-am)
+stackexchange post.
```
profradha:
```suggestion
[this stackexchange post.](https://substrate.stackexchange.com/questions/5067/delegating-and-undelegating-during-the-lock-period-extends-it-for-the-initial-am)
```

DrW3RK (approved):
The examples are a good addition.

PR #3857, opened 2022-10-01
Replace Parachain Tutorials links

DrW3RK:
https://docs.substrate.io/tutorials/connect-relay-and-parachains/  is the latest link


DrW3RK:
The original link is correct.

PR #3896, opened 2022-10-10
Use Non-Transfer Proxies to join Nomination Pools

DrW3RK:
https://github.com/paritytech/polkadot/blob/3d6b56342e5e1313685aae6f4e222cbf91ba790f/runtime/kusama/src/lib.rs#L1041

Add a note that says Staking proxies are not supported


filippoweb3 (approved):
Looks good to me ;)

PR #3897, opened 2022-10-10
Minor fixes - Anonymous Proxy Rename

DrW3RK:
The term "redenomination" is associated with money/currency. Added minor fixes to the section about renaming anonymous proxies to pure proxies.


filippoweb3 (approved):
Hey Radha, I did not know about this. Thanks for correcting it.

PR #3911, opened 2022-10-12
Deactivate banners

je-boska:
@DrW3RK To be merged tomorrow, Oct. 13th. 

I chose to comment out the banners, so that future users will have an easy time using the template. Lmk if you would prefer a different solution.


DrW3RK (approved):
Will be merged, first thing tomorrow morning!

Issue #3815, opened 2022-09-26
Use GitHub Actions to add new auction blocks to the auction schedule tool


alfarok:
https://github.com/w3f/polkadot-wiki/pull/3814 introduces a new react component that visualizes auction data for Polkadot and Kusama on the wiki/guide.  The block date/times are estimated based on the live chain state when the page loads. 

 https://github.com/w3f/polkadot-wiki/issues/3768 was the original task that called for this migration and move to using on-chain data for better time estimates.  The last step in fully automating this process would be to add a new GitHub action that updates the `auctions.js` file to occasional check for new auction blocks that will occur in the future.  We currently have hard-coded values pretty far into the future but to make this tool completely hands-off we can leverage GitHub actions to occasionally perform a check and open a new PR with the additions.

This tool will not have to run frequently (maybe once a month) but it will prevent us from having to manually remembering to do this on occasion.  The tool should simple check if any new auctions are scheduled that do not exist in the `PolkadotAuctions.json` and/or `KusamaAuctions.json` files, if so append the blocks and create a new PR.

It would be inefficient to try and pull all of this data live every time the page loads, which is why the blocks are cached in `PolkadotAuctions.json` and `KusamaAuctions.json` files.  It keeps the tool snappy and avoids making 100+ RPC calls every time the page loads.

Lastly, it may be worth adding block timestamps to the schema if the block has already reach finality.  This will further reduce latency from on-chain calls where the values will never change.

alfarok:
Addressed in #3946

PR #3946, opened 2022-10-18
Auction schedule component should use cache for finalized blocks and use GitHub action for updating cache

alfarok:
This PR addresses #3815.  The result is the removal of almost all latency in interacting with the tool.  We still need to wait to pull the current block data from on-chain but once loaded everything else is just about instant now.

Done:
- [x] Add node script capable of running via GitHub actions that updates the auction caches for Polkadot & Kusama from finalized on-chain data
- [x] Removes several inaccuracies in previous hard-coded data
- [x] Script runs once a week or via GitHub actions web UI
- [x] A pull request  should be made prior to the node script terminating (JSON files are already updated locally and ready to be committed)

Testing:
https://github.com/w3f/polkadot-wiki/pull/3947 (test PR created by the new GitHub action)

Also if no updates are found, the action will not attempt to commit/push and exit gracefully:
image

Sample:
image


DrW3RK (approved):
@alfarok you can merge this PR.

Issue #3915, opened 2022-10-13
Add the info about rewards earned by a Staking miner


DrW3RK:
Add info about the approximate rewards earned by the Staking miner in https://wiki.polkadot.network/docs/learn-staking-miner

DrW3RK:
Here is the chat from the Watercooler for reference

<img width="941" alt="image" src="https://user-images.githubusercontent.com/86818441/195604035-b42a041a-8e9b-4a74-910d-fb602f44991a.png">

<img width="974" alt="image" src="https://user-images.githubusercontent.com/86818441/195658799-70ce874b-95db-44a0-a0fd-9d31a40c797e.png">

Linked code - https://github.com/paritytech/substrate/blob/241b0d0455453499763d0db0b4ea4188012b372f/frame/election-provider-multi-phase/src/signed.rs#L453

DrW3RK:
https://github.com/w3f/polkadot-wiki/pull/3954

Issue #3853, opened 2022-09-30
MEGA: unified graphics


kianenigma:
I know that the priority of no one working on the wiki is aesthetics, but I presume that w3f has the resources to gradually move towards fixing this:

The graphics used in the wiki have two main flaws:

1. they look rather simple and sometimes not so good.
2. Even worse, they clearly have different styles, based on who made them. Some look a bit better, some worse. 

(The same comment applies to the videos linked to the wiki).

I think we need a unified theme for all of this. Probably one light and one dark. And this is my personal opinion, but I don't think it is fair to actually expect the people currently maintaining the wiki to make these graphics. This is probably the job for a technical illustrator that can be contracted/hired. 

I am sure most of the people maintaining these docs are excellent technical writers, but I think we need dedicated illustrators and content creators as well.

filippoweb3:
Today the Teched Team submitted a request to address this PR to the Marcomms Team.

CrackTheCode016:
This should be revisited after the new IA and content refactor is in place.

PR #3954, opened 2022-10-18
Staking miner deposit and reward section

emresurmeli:
Add a deposit and reward mechanics section, and use "deposit" instead of "bond" when talking about the locked-up DOT required by the staking miners. Targeting issue #3915


emresurmeli:
cc: @kianenigma

docs/learn/learn-staking-miner.md
```diff
@@ -48,7 +51,28 @@ order to submit solutions to the NPoS election. When the correct time comes, it
...
+{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedDepositBase" defaultValue={16} filter="humanReadable"/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedDepositBase" defaultValue={16} filter="humanReadable"/> :kusama }}
+
+Current deposit per byte(**SignedDepositByte**) is
+{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={16} filter="humanReadable"/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={16} filter="humanReadable"/> :kusama }}
+
+Current rewards(**SignedRewardBase**) is
+{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={16} filter="humanReadable"/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={16} filter="humanReadable"/> :kusama }}
```
alfarok:
```suggestion
Current deposit(**SignedDepositBase**) is
{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedDepositBase" defaultValue={400000000000} filter="humanReadable"/> :polkadot }}
{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedDepositBase" defaultValue={133333332000} filter="humanReadable"/> :kusama }}

Current deposit per byte(**SignedDepositByte**) is
{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={97656} filter="humanReadable"/> :polkadot }}
{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={32551} filter="humanReadable"/> :kusama }}

Current rewards(**SignedRewardBase**) is
{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={10000000000} filter="humanReadable"/> :polkadot }}
{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={100000000000} filter="humanReadable"/> :kusama }}
```

The RPC formatting you used here is perfect and won't be modified by prettier in any negative way!  Modifications are required for the default values to display properly before the on-chain values kick-in.  In the current state the page will initially loading with 0 for all these values until the RPC values are retrieved.  The proper defaults should be the raw values (without commas) you would get when calling those endpoint paths say via PolkadotJS, prior to applying the filter. The `humanReadable` filter will take that raw value of say `400000000000` Plancks and convert it to `40 DOT` prior to rendering.  So for example:

image

Let me know if you have any further questions on this!

components/RPC-Connection.jsx
```diff
@@ -139,7 +139,7 @@ function applyFilter(value, filter, network, setReturnValue) {
     	break;
 		case "blocksToDays":
 			BlocksToDays(value, setReturnValue);
-			break;
+		break;
```
alfarok, line 142:
```suggestion
    	break;
```
The original indentation is correct for the switch case

alfarok (commented):
Hey @emresurmeli, everything looks good to me!  I just have 2 minor formatting related suggestions.  If my suggestions look okay to you, I think you are good to merge pending any additional feedback from @DrW3RK.

docs/learn/learn-staking-miner.md
profradha: With the current precision of the rendered DOT values, the deposit byte value will be shown as 0.000 DOT. Not sure if we have to increase the precision which effects the whole WIki or make an exception just for this one case!

docs/learn/learn-staking-miner.md
```diff
@@ -48,7 +51,28 @@ order to submit solutions to the NPoS election. When the correct time comes, it
 solution and submits it to the chain. The default miner algorithm is sequential Phragmén with a
 configurable number of balancing iterations that improve the score.
 
-![NPoS election optimization](../assets/staking-miner/NPoS-election-optimization.png)
+## Deposit and reward mechanics
+
+The staking miners are required to pay a deposit in order to post their solutions. The deposit is a
+sum of the `SignedDepositBase`, which is a fixed amount and `SignedDepositByte`, a variable amount
+per KB of solution data. All good solutions are subject to recieving a `SignedRewardBase`.
```
profradha: fixed typo of 'receiving' and a bit of rephrasing. Also, the deposit collected seems to include a per-weight deposit as well https://github.com/paritytech/substrate/blob/49734dd1d72a00b9d3b87ba397661a63e0e17af3/frame/election-provider-multi-phase/src/signed.rs#L501

```suggestion
The staking miners are required to pay a deposit to post their solutions. This deposit is a
sum of the fixed `SignedDepositBase` amount and the variable`SignedDepositByte` amount, which depends on the size of the solution data. All good solutions are subject to receiving a `SignedRewardBase`.
```

docs/learn/learn-staking-miner.md
```diff
@@ -48,7 +51,28 @@ order to submit solutions to the NPoS election. When the correct time comes, it
...
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={16} filter="humanReadable"/> :kusama }}
+
+Current rewards(**SignedRewardBase**) is
+{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={16} filter="humanReadable"/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={16} filter="humanReadable"/> :kusama }}
+
+Once the staking miner with a good solution is ready to be rewarded,
+[`finalize_signed_phase_accept_solution`](https://github.com/paritytech/substrate/blob/master/frame/election-provider-multi-phase/src/signed.rs#L453-L474)
+will be called, which will submit the solution to the queue of solutions, reward the staking miner
+and refund the deposit.
```
profradha: Just a question that might come up from the readers -

It is unclear why the good solution is submitted to the queue of solutions here. What does this queue contain? Why can't the best solution that is being rewarded be directly accepted?

```suggestion
Once the staking miner with a good solution is ready to be rewarded,
[`finalize_signed_phase_accept_solution`](https://github.com/paritytech/substrate/blob/master/frame/election-provider-multi-phase/src/signed.rs#L453-L474)
will be called, which will submit the solution to the queue of solutions, reward the staking miner
and refund the deposit.
```

DrW3RK (changes requested):
Thanks @alfarok Your suggestions can be committed to the PR

A few minor issues to address. The newly added content is very resourceful to understand the reward mechanics. Thanks! @emresurmeli

docs/learn/learn-staking-miner.md
alfarok: Good point!

image

It would impact every RPC component that uses the `humanReadable` filter.  If we want something more precise for rare instances like this, I would suggest adding a new filter.  Maybe something like `humanReadablePrecise` which returns an appropriate value with the increased decimal precision.  It can be added in [filters.js](https://github.com/w3f/polkadot-wiki/blob/master/components/utilities/filters.js).

@emresurmeli Let me know if you go this route and need any assistance.

docs/learn/learn-staking-miner.md
```diff
@@ -35,11 +36,13 @@ starts, only the off-chain workers can provide election results.
...
 Running the staking miner requires passing the seed of a funded account in order to pay the fees for
 the transactions that will be sent. The same account's balance is used to reserve deposits as well.
-The best solution in each round is rewarded. All correct solutions will get their bond back and the
-ones that submit invalid solutions will lose their bond.
+The best solution in each round is rewarded. All correct solutions will get their deposit back and
+the ones that submit invalid solutions will lose their deposit.
 
 ## NPoS election optimization
 
+![NPoS election optimization](../assets/staking-miner/NPoS-election-optimization.png)
```
kianenigma, line 48: #3853 another different style, it does look good though to be fair :D

docs/learn/learn-staking-miner.md
kianenigma:
```suggestion
per KB of solution data. 
```

Last sentence is wrong. Only the winner gets this reward.

docs/learn/learn-staking-miner.md
kianenigma: Would be good if you also show an example: a solution that is 200KB of data to post onchain, costs xxx. 

You can look at subscan and old block to find the size of a typical transaction.

docs/learn/learn-staking-miner.md
```diff
@@ -48,7 +51,28 @@ order to submit solutions to the NPoS election. When the correct time comes, it
...
+Current deposit per byte(**SignedDepositByte**) is
+{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={16} filter="humanReadable"/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedDepositByte" defaultValue={16} filter="humanReadable"/> :kusama }}
+
+Current rewards(**SignedRewardBase**) is
+{{ polkadot: <RPC network="polkadot" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={16} filter="humanReadable"/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.electionProviderMultiPhase.signedRewardBase" defaultValue={16} filter="humanReadable"/> :kusama }}
+
+Once the staking miner with a good solution is ready to be rewarded,
+[`finalize_signed_phase_accept_solution`](https://github.com/paritytech/substrate/blob/master/frame/election-provider-multi-phase/src/signed.rs#L453-L474)
```
kianenigma: These links will be useless in no time, please use a permalink!

kianenigma: This gives the false impression that some account calls `finalize_signed_phase_accept_solution` which is far from true. I think it is good that you are trying to back statements with code, but it need to be accurate. 

Here's a lead to follow: the miner works in the `SignedPhase`, and at the end of signed phase, the runtime automatically calls `finalize_signed_phase`. 

also, this is a good lead: https://github.com/paritytech/substrate/blob/3bbdba405d42d7023eeb297efb03a3ec4ccf9563/frame/election-provider-multi-phase/src/lib.rs#L69-L74

docs/learn/learn-staking-miner.md
```diff
@@ -124,3 +148,4 @@ resources section below.
 
 - [Election Pallet definition](https://crates.parity.io/pallet_election_provider_multi_phase/index.html)
 - [Staking Miner repository](https://github.com/paritytech/staking-miner-v2)
+- [Signed phase parameter configuration on Polkadot](https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L389:L397)
```
kianenigma: Permalink please.

kianenigma (commented):
Has a few small mistakes, looks good otherwise.

docs/learn/learn-staking-miner.md
emresurmeli: Yeah, good point. It's well explained at the bottom of the page, I'll link.

docs/learn/learn-staking-miner.md
emresurmeli: Yeah, this is all mentioned below in the signed phase section. Actually makes more sense to move the deposit and reward mechanics below that and it will flow better.

docs/learn/learn-staking-miner.md
emresurmeli: I'm going to create another issue for adding this feature, meanwhile, for the `signedDepositByte` value I will put the right output into the default answer and leave a TODO above the code block.

DrW3RK:
I guess it is better to hardcode this value and modify it after the issue https://github.com/w3f/polkadot-wiki/issues/3986 is resolved. I will be pushing the changes shortly

image

docs/learn/learn-staking-miner.md
```diff
@@ -117,10 +121,40 @@ Queue
...
+which is a fixed amount.
+
+Current deposit per byte(`SignedDepositByte`) is
+{{ polkadot: 0.0000097656 DOT :polkadot }}{{ kusama: 0.00000032551 KSM :kusama }} and the total is
+variable depending on the size of the solution data. For example a solution weighing 200KB would
+yield {{ polkadot: 200 x 0.0000097656 = **0.00195312 DOT**. :polkadot }}
+{{ kusama: 200 x 0.00000032551 = **0.000065102 KSM**. :kusama }}
+
+And the weight deposit(`SignedDepositWeight`) is calculated by the runtime once a winning solution
+has been elected.
```
profradha: The entire deposit amount is calculated and collected to store this solution at the time of deposit. So, calculating the deposit value after the solution is elected does not seem right.

https://github.com/paritytech/substrate/blob/f2bc08a3071a91b71fec63cf2b22c707411cec0e/frame/election-provider-multi-phase/src/signed.rs#L512


```suggestion
And the weight deposit(`SignedDepositWeight`) is currently set to `0` and has no effect.
```

profradha: @kianenigma Could you please confirm if this is the case and the role of `SignedDepositWeight` in general?

image

DrW3RK (approved):

PR #3971, opened 2022-10-21
Small change in Phragmen algorithm explainer

SBalaguer:
If I understood the explanation correctly, I believe point (1) of the algorithm is wrong. Voters cast their vote to candidates, and not candidates to candidates. If this is wrong, please dismiss this PR.


DrW3RK (approved):
Thank you @SBalaguer

PR #4025, opened 2022-11-07
Restore conditional check bool to string

alfarok:
This PR basically reverts #4024 as #4024 fixed the conditional check.  The boolean values are stored in env variables, which converts them to strings.  This can be seen in the final line of the logs below:

image


alfarok:
Fixed:
image

PR #4051, opened 2022-11-16
Remove ambiguity regarding annual inflation

DrW3RK:
Add permalinks to the inflation on Polkadot and Kusama networks


filippoweb3 (approved):
Lgtm, noticed that there was a problem with prettify changing the position in text of an RPC component. Not sure if I fixed it since there might be an automatic call of pretty function after we merge.

PR #4058, opened 2022-11-16
Update README

alfarok:
General README update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,13 +73,6 @@ Using yarn, run:
 yarn install
 ```
 
-> The site is built using Docusaurus: you may need to install Docusaurus before running the Wiki
-> locally.
-
-```bash
-yarn upgrade @docusaurus/core@latest @docusaurus/preset-classic@latest
-```
-
 ### Build
 
 > The Wiki is currently in the process of being deployed onto IPFS and will follow the same
@@ -111,20 +104,6 @@ yarn kusama:start
 yarn polkadot:start
 ```
 
-### Test
-
-✅❌ Running tests locally:
-
-```bash
-yarn polkadot:test
-
-or
-
-yarn kusama:test
-```
-
-There is no need to run both as the tests are unified.
-
 ### Publish
 
 :bird: Publishing the Kusama Guide:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,13 +73,6 @@ Using yarn, run:
 yarn install
 ```
 
-> The site is built using Docusaurus: you may need to install Docusaurus before running the Wiki
-> locally.
-
-```bash
-yarn upgrade @docusaurus/core@latest @docusaurus/preset-classic@latest
```
alfarok, line 80: You shouldn't need to run this as long as the previous `yarn install` command was run as outlined.

README.md
```diff
@@ -111,20 +104,6 @@ yarn kusama:start
...
-```bash
-yarn polkadot:test
-
-or
-
-yarn kusama:test
-```
-
-There is no need to run both as the tests are unified.
-
```
alfarok, line 127: Right now this runs [5 tests](https://github.com/w3f/polkadot-wiki/blob/master/tests/rpc.test.js) relating to react RPC components.  Given the limited coverage I don't think this needs to be on the main README.

README.md
```diff
@@ -266,15 +245,8 @@ spacing, especially after running prettier.
 
 ## Internationalization
 
-| ❗ Currently the Wiki is being reorgnaized and updated. It is recommended to resume working on translations after the Wiki revamp is completed. |
-| ----------------------------------------------------------------------------------------------------------------------------------------------- |
-
-We are using Crowdin to manage all different translations. You can go to the
```
alfarok, line 272: We can restore this if required but given that it is currently inactive I removed the details.

I was also wondering if it makes sense to remove everything under https://github.com/w3f/polkadot-wiki/tree/master/polkadot-wiki/translated_docs?  If this information is out-of-date or incorrect does it make sense to keep it?  We can always recover it through the git history.  Removing these files also removes the requirement of having to use the `files to excludes: polkadot-wiki/translated_docs` filter in vs code when searching.

profradha: I agree

DrW3RK (approved):

README.md
alfarok: Awesome, I will follow up with another PR that just removes that directory :+1:

PR #4072, opened 2022-11-21
Remove all non-referenced images from assets

### LINKED ISSUE #4067: Order Figures in the /assets folder
The Figures in the /assets Folder look a bit messy. It would help to go through them and see which one is still used in the wiki and which one is not. I created an "old-figures" folder and we could start to move them there cleaning up a bit.

alfarok:
This PR addresses #4067.

I was able to leverage [this line](https://github.com/w3f/polkadot-wiki/blob/master/scripts/auditLinks.py#L149) in the link auditing script to extract all markdown links that contain images with the following extensions:
```js
[".png", ".jpg", ".svg", ".gif", ".jpeg"]
```
This covered all items in the `./assets` folder minus a couple `.mp4` and `.ico` files which I manually verified.

Once I had a list of all assets and all locally linked images, I diff'd the 2 sets to reveal 

1. Which items where in `./assets` but not referenced in the docs (280 items)
2. Which items are referenced in the docs but don't have a supporting asset (0 items)

I then manually verified the removal candidates with a VSCode project search.  I was able to identity some false positives that evaded the python script such as:
- images referenced in `.js` files
- images referenced in `.md` files outside of `./docs`
- images referenced in the markdown using an `` tag as opposed to the markdown flavor of `[linkname](link)`

**To reiterate all of these removals returned 0 references when searched at the root of the project.**

When running `yarn polkadot:build`, the pre-build process checks for broken image links and prevents building if any are found.  I identified 2 false positives that I accidentally removed using this check.


filippoweb3 (approved):
This is awesome Keith! Such a smart routine! 

For me everything looks fine, just not sure if deleting those images is the best way to proceed. I created with the issue a folder "old-figures" thinking that maybe it would be more cautious to put those figures there just in case. I would like to have Radha's opinion here (added him as a reviewer).

DrW3RK:
Yeah, although we can retrieve the images from GitHub history, I think placing them in an old figures folder makes sense. At times, I've used some of the legacy images in my presentations or in different places on the Wiki, and having them in the master repo is handy. 

What we could do is that this workflow PR moves images to the unused images folder and we create another PR to remove unwanted images. Hope this is okay.

alfarok:
Not a problem to move all the archived images to a folder instead.  Should it still be on the wiki though or would it be better to add an archive folder on the google drive for these assets? @DrW3RK

alfarok:
I added [this archive folder](https://drive.google.com/drive/folders/1Hd3r9b178vupm-E_Qq_o9OC_7mQIVW46?usp=share_link) to the existing `Visual Assets` folder on the shared TechEd drive.

If you guys think it's better to keep them in the repo still we can drag this folder back into `./assets`

alfarok:
As discussed in the team sync, the files will remain in the repo but have been moved to a new `./_legacy` directory under `./assets`.  Additional PRs can be made to remove content in this directory if the legacy content is deemed no longer have value to the project/community or the media displays dated information that is no longer accurate.

PR #4106, opened 2022-11-29
Reorg and Move Cryptography Page

### LINKED ISSUE #4105: The cyptography pages can be grouped and moved to Advanced
The info in "Keys" page and "Cryptography Explainer" page can be grouped into one page. That page can be moved under the Advanced menu. The current "Cryptography Explainer"  is very short and does not make sense to keep it as a standalone page.

filippoweb3:
See linked issue


DrW3RK:
Added comments to the linked issue https://github.com/w3f/polkadot-wiki/issues/4105

filippoweb3:
@DrW3RK thank for the suggestion. I think that the redirect option is better as it keeps the wiki structure clean allowing us to delete old pages or pages where we moved content. Let me know your opinion about this. I added the redirect for the learn-keys page that I deleted. Let me know if this is ok, if yes I will then do the same for the [other PR](https://github.com/w3f/polkadot-wiki/pull/4108).

DrW3RK (approved):

PR #65, opened 2019-10-07
Minor fixes in the smart contract file

Swader:

PR #4138, opened 2022-12-02
Update auction caches by running the update script locally with fix to Kusama auction 63 & 64

alfarok:
We saw a failure in the automated auction cache updating script this past week.  **It appears Auction `#63` and `#64` did not start during the regular cadence**.

Looking at the Governance Dispatch Queue for Kusama it can be determined that auction `#65` is scheduled as the cadence and cache would expect.  Also checking some block explorers it appears `#63` & `#64` have not started:

https://kusama.subscan.io/auction

or

https://parachains.info/auctions/kusama

For now, I removed `#63` and `#64` from the cache to prevent the automation script from failing.  

It was determined from https://kusama.polkassembly.io/motion/582 in the discussion below that there was an enactment issue due to an upgrade that caused this auction to not be scheduled.  It appears they will be re-scheduled at a later time.  When this happens we can re-add `#63` and `#64` at their original indices or new sequential indices.  

The cache stores the index number so we don't rely on a looping count to determine the auction numbers.  This means removing `#63` and `#64` doesn't cause `#65` to become `#63`, preserving the expecting numbering.

The remaining updates in this PR were generated once I fixed the failure points at `#63` and `#64`.  A new try/catch statement was also added to help identify this issue in the future.

image


laboon:
FYI - https://kusama.polkassembly.io/motion/582

alfarok:
@DrW3RK I think this should be good to go for now.  We can monitor the situation and determine how the missing auction gets re-added as more details become available.  The automation script will tell us if future auctions continue as expected or if they don't begin and need to be addressed.

components/utilities/updateAuctions.js
```diff
@@ -140,38 +140,43 @@ async function LoadAPI(chain) {
...
-      const onboardEndBlock = onboardStartBlock + DaysToBlocks(PolkadotLeasePeriodPerSlot * 12 * 7);
-      return [biddingStarts, auctionEndBlock, onboardStartBlock, onboardEndBlock]
-    }
-    else if (chain === "Kusama") {
-      const biddingStarts = startBlock + KusamaStartingPhase;
-      const auctionEndBlock = biddingStarts + KusamaEndingPeriod;
-      const onboardStartBlock = auctionLeasePeriod * KusamaSlotLeasePeriod + KusamaSlotLeaseOffset;
-      const onboardEndBlock = onboardStartBlock + DaysToBlocks(KusamaLeasePeriodPerSlot * 6 * 7);
-      return [biddingStarts, auctionEndBlock, onboardStartBlock, onboardEndBlock]
+  try {
```
alfarok, line 143: The only change here is wrapping all this logic in a try/catch for better error messaging related to auctions have inaccurate starting blocks defined in the cache.

DrW3RK (approved):

PR #3851, opened 2022-09-30
Update maintain-endpoints.md

PolinaShanti:
Mainnet
Hello Polkadot Team!
I added NOWNodes to your providers' list.
We provide a dedicated and mainnet node for the Polkadot network. Thank you!

PR #4145, opened 2022-12-05
Update third party node providers

AnnaNodes:
Hello,

We are a Polkadot node provider NOWNodes.io.
May you add us to the Polkadot third party providers?


DrW3RK:
Hi @AnnaNodes, As mentioned in https://github.com/w3f/polkadot-wiki/pull/3851 the Wiki lists the public endpoints (which do not require API keys to access).

AnnaNodes:
Hi @DrW3RK !

Thank you for the explanation! May you suggest the right page in Polkadot docs for full node providers?

DrW3RK:
There is no such section on the Wiki. The wiki includes only public endpoints. If you create one and add it to the Polkadot-JS UI, you will get automatically added to the list here.

Issue #4110, opened 2022-11-30
Improve coverage of Kusama docs in comparison to Polkadot


alfarok:
### Description
Many of the Polkadot docs are applicable to Kusama and the minor differences are normally handled with conditional rendering.  I completed a rough audit to gauge the gap between the sets by comparing `polkadot` vs `kusama` word frequency in each doc.  The results are outlined below and can serve as a starting point for bridging the gap between the sets.

#### Key:

⬜️ = not yet reviewed

⏩ = skipped (likely the Polkadot/Kusama references are example code or titled Polkadot/Kusama specific - almost all maintain docs)

✅ = reviewed, validated count discrepancy, no changes required

✔️ = reviewed, modifications made, PR linked (hover check)

:red_circle: = delta of 15 or more

:yellow_circle: = delta of 5 - 10

:green_circle: = delta of less than 5

#### Audit:

|Reviewed|File|Polkadot Count|Kusama Count|Difference|Severity|
|---|---|---|---|---|---|
|⏩|[maintain-sync.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//maintain/maintain-sync.md)|51|1|50|🔴|
|⏩|[maintain-guides-how-to-validate-polkadot.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//maintain/maintain-guides-how-to-validate-polkadot.md)|64|14|50|🔴|
|✅|[build-open-source.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//build/build-open-source.md)|55|5|50|🔴|
|✅|[build-tools-index.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//build/build-tools-index.md)|34|0|34|🔴|
|✅|[getting-started.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//general/getting-started.md)|35|4|31|🔴|
|✅|[ledger.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//general/ledger.md)|30|0|30|🔴|
|[✔️](https://github.com/w3f/polkadot-wiki/pull/4150)|[build-parachains.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//build/build-parachains.md)|86|58|28|🔴|
|⏩|[maintain-guides-how-to-nominate-polkadot.md](https://github.com/w3f/polkadot-wiki/tree/master/docs//maintain/maintain-guides-how-to-nominate-polkadot.md)|33|6|27|🔴|
[...]

PR #4155, opened 2022-12-07
Extend Kusama coverage for learn-staking-advanced doc

alfarok:
Part of the #4110 initiative.


filippoweb3 (approved):
lgtm. can you make sure the rendering is ok when you run it locally as well? Thanks for bringing these additions to Kusama! :)

alfarok:
> lgtm. can you make sure the rendering is ok when you run it locally as well? Thanks for bringing these additions to Kusama! :)

Yes I ran both Polkadot and Kusama on this one locally to validate since there was a lot of conditional rendering.  What I have noticed is large paragraphs of conditionals tend to cause more formatting issues because of newlines and occasional symbols that can disturb the `{{ }}` syntax.

PR #4168, opened 2022-12-09
Auction cache fixes

alfarok:
I noticed there was an incorrect hash for auction `#33` on Polkadot likely from the previous hard-coding.  With the auction automation updates being offline the last few weeks with the Kusama dispatch queue issue it wasn't updated or noticed.

I also noticed an intermittent timing bug in the auction cache updating GitHub action which can lead to the Polkadot update not completing before the Kusama update begins.  Since this async behavior can occur, the referenced chain API can be swapped (from Polkadot to Kusama) which can cause the Polkadot update to not finish.  I will file this in a separate issue and try to address it ASAP.


alfarok:
@DrW3RK merging this in as the currently displayed data for Polkadot auction `#33` is wrong.

alfarok:
Fixed the async issue w/ #4170

PR #4184, opened 2022-12-13
Manual Kusama auction cache update.

alfarok:
Same issue outlined in #4138, removing auction `#65` for now as it did not start as scheduled.

This PR also updates previous block values that are in-progress (correctly) as the automation is not running for Kusama as long as invalid starting blocks are still being provided.

Polkadot is already up-to-date.

### DIFF
--- a/components/utilities/updateAuctions.js
+++ b/components/utilities/updateAuctions.js
@@ -184,7 +184,6 @@ async function GetAuctionBlocks(api, currentBlock, startBlock, chain) {
     }
   } catch (error) {
     console.log(`Failure updating auction blocks on ${chain} at starting block ${startBlock}.`)
-    throw error.message;
   }
 }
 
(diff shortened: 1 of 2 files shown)


components/utilities/updateAuctions.js
```diff
@@ -184,7 +184,6 @@ async function GetAuctionBlocks(api, currentBlock, startBlock, chain) {
     }
   } catch (error) {
     console.log(`Failure updating auction blocks on ${chain} at starting block ${startBlock}.`)
-    throw error.message;
```
alfarok, line 187: Error messaging/handling is actually more clear without this throw:
```
Failure updating auction blocks on Kusama at starting block 15724800.
```
as opposed to 
```
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Cannot read properties of null (reading '0')".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

PR #4187, opened 2022-12-14
Change refs of Gov2 to OpenGov

### LINKED ISSUE #4183: Change references to "gov2" to "OpenGov"
This is the new name for it. We should start using OpenGov, although in the first mention of it on a page, we can say something like "OpenGov, formerly known as Gov2 or Gov v2,..." etc.

DrW3RK:
Renaming the doc id and the doc as well. Fairly new, so no real need to implement redirects.

### DIFF
--- a/polkadot-wiki/sidebars.js
+++ b/polkadot-wiki/sidebars.js
@@ -88,7 +88,7 @@ module.exports = {
             },
             {
               type: "category",
-              label: "Architecture and Components",
+              label: "Network Components",
               items: [
                 "learn/learn-architecture",
                 "learn/learn-collator",
--- a/polkadot-wiki/sidebars.js
+++ b/polkadot-wiki/sidebars.js
@@ -88,7 +88,7 @@ module.exports = {
             },
             {
               type: "category",
-              label: "Components",
+              label: "Architecture and Components",
               items: [
                 "learn/learn-architecture",
                 "learn/learn-collator",
--- a/polkadot-wiki/sidebars.js
+++ b/polkadot-wiki/sidebars.js
@@ -72,7 +72,7 @@ module.exports = {
               items: [
                 "learn/learn-staking",
                 "learn/learn-nomination-pools",
-                "learn/learn-staking-faq"
+                "learn/learn-staking-faq",
                 "learn/learn-nominator",
               ],
             },
--- a/polkadot-wiki/sidebars.js
+++ b/polkadot-wiki/sidebars.js
@@ -43,7 +43,7 @@ module.exports = {
           items: [
             {
               type: "category",
-              label: "Accounts",
+              label: "Accounts and Identity",
               items: [
                 "learn/learn-accounts",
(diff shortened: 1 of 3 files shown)


polkadot-wiki/sidebars.js
```diff
@@ -66,31 +68,35 @@ module.exports = {
...
-                "learn/learn-architecture",
-                "learn/learn-collator",
-                "learn/learn-consensus",
-                "learn/learn-governance",
-                "learn/learn-gov2",
-                "learn/learn-identity",
+                "learn/learn-staking",
+                "learn/learn-nomination-pools",
+                "learn/learn-staking-faq",
                 "learn/learn-nominator",
```
filippoweb3: The nominator page contains some advanced topics such as bags list. This topic is already explained in the advanced staking page. I see the basics page as something containing really easy info that all user can understand if the want use polkadot. probably everything related to the UI must be put in advanced, and leave the dashboard-realted things in basics?

polkadot-wiki/sidebars.js
```diff
@@ -43,14 +43,16 @@ module.exports = {
...
-              label: "Accounts",
+              label: "Accounts and Identity",
               items: [
                 "learn/learn-accounts",
                 "learn/learn-account-generation",
                 "learn/learn-balance-transfers",
                 "learn/learn-extrinsics",
                 "learn/learn-account-multisig",
                 'learn/learn-proxies',
+                "learn/learn-identity",
```
filippoweb3: Isn't it confusing having identity here? I see this as not a basic information, and having accounts and identity together can be confusing. What do you think?

polkadot-wiki/sidebars.js
```diff
@@ -66,31 +68,35 @@ module.exports = {
...
-                "learn/learn-polkadot-host",
-                "learn/learn-polkadotjs",
-                "learn/learn-registrar",
-                "learn/learn-runtime-upgrades",
-                "learn/learn-transaction-fees",
+              ],
+            },
+            {
+              type: "category",
+              label: "Governance",
```
filippoweb3: Maybe this can be put in Advnaced?

filippoweb3 (changes requested):
In my opinion it would be good to discuss intrnally how to do with the content organization. I see the Basics menu as somthing with info about really super basic things. I am asking myself: "I am not part of Web3F and I would like to start using Polkadot to stake some tokens and understand how the system works, what do I need to know without being overwhelmed?". This was actually my case before joining the team. I don't need to know about Gov, Identity & Registrars e maybe a couple of other things. I see some of the current pages distracting and discouraging. If there are people that want to learn in depth they can go to advanced. What is your opinion on this?

DrW3RK:
You're right. This discussion needs to be on a PR for Wiki reorg. I reverted the changes made to the sidebar.

filippoweb3 (approved):
lgtm

PR #4210, opened 2022-12-20
Unify url formatting

alfarok:
### Issue

It was recently discovered that document links formatted in the following manner:

```
[Collator subsystem][]

[collator subsystem]: https://paritytech.github.io/polkadot/book/node/collators/index.html
```

can causes conflicts with prettier.  For example. prettier could split the tag `[Collator subsystem][]` between `Collator` and `subsystem` with a newline, which breaks the link.  This exact scenario is found many times below.  Even if it is not currently the case, it can happen with future PRs when the formatter runs.

This formatting is also not audited by the link auditing tool.

### Solution

To resolve this issue, this PR unifies all link formatting to use the traditional markdown syntax, such as `[link](url)`.

### Testing

I ran the url auditing tool to verify this change didn't introduce any broken links

### DIFF
--- a/docs/general/kusama/kusama-getting-started.md
+++ b/docs/general/kusama/kusama-getting-started.md
@@ -19,8 +19,8 @@ Kusama is owned by those who hold the Kusama tokens (KSM). There's no central ki
 changes are made through the protocol's on-chain governance.
 
 The network is a permissionless and anyone can come along and start using it. Those who participated
-in the Polkadot sale can claim a proportional amount of KSM through the [Kusama Claims
-process][claims].
+in the Polkadot sale can claim a proportional amount of KSM through the
+[Kusama Claims process](kusama-claims).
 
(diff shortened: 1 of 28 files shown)


docs/general/kusama/kusama-getting-started.md
```diff
@@ -19,8 +19,8 @@ Kusama is owned by those who hold the Kusama tokens (KSM). There's no central ki
 changes are made through the protocol's on-chain governance.
 
 The network is a permissionless and anyone can come along and start using it. Those who participated
-in the Polkadot sale can claim a proportional amount of KSM through the [Kusama Claims
-process][claims].
```
alfarok, line 23: Here is one of many examples where prettier was splitting/breaking links

alfarok: was rendering as:
image

DrW3RK (approved):

Issue #4095, opened 2022-11-25
Maintenance Automation Enhancements


alfarok:
As discussed in several recent tasks/issue, the team has identified several areas where running automation scripts on a regular interval is useful and saves the group time/effort.  This issue is intended to track any relevant suggestions for future integrations.

A couple existing examples of this type of automation are present in the:
- [x] Status badge generator (weekly)
- [x] Auction tool cache updater (weekly)
- [x] Url auditing tool (monthly)

Some additional ideas that have come up include:
- [x] Image auditing tool (monthly)
- [x] Prettier formatter (weekly)
- [ ] ~~Polkadot vs Kusama reference analytics ([example audit](https://github.com/w3f/polkadot-wiki/issues/4110) - monthly)~~

Please feel free to add any additional suggestions.

alfarok:
Added automated prettier formatting over the entire repo in https://github.com/w3f/polkadot-wiki/pull/4213.

alfarok:
Added automated image auditing in #4236.

PR #4213, opened 2022-12-21
Add automated Prettier repo scan to GitHub actions and run on all files

alfarok:
Addresses part of https://github.com/w3f/polkadot-wiki/issues/4095.

This PR adds a new GitHub workflow that will run prettier on the entire repo on a regular schedule.  Although we added a pre-commit hook, it is still easy to introduce un-formatted content through the GitHub file editor or committing inline suggestions.  

Sometimes when creating a new PR, prettier format content you personally have not edited (possibly from the sources listed above) will appear in the comparison diff. The automation routine should help maintain the formatting without letting too much time pass between cleanups.  This will help reduce this occurrence. 

I set the interval to weekly to begin so I can monitor that it is working correctly but we can likely reduce this to once a month if it because tedious with small changes appearing too frequently.

### DIFF
--- a/.github/workflows/prettier-all.yml
+++ b/.github/workflows/prettier-all.yml
@@ -1,0 +1,35 @@
+name: Run Prettier on entire repo
+on:
+  schedule:
+    - cron: '0 0 * * 1' # Once a week
+  workflow_dispatch: # Or manually from GitHub UI
+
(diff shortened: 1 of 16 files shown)


DrW3RK (approved):
Sounds good to me!

PR #4246, opened 2022-12-29
Update Crowdloan documentation

SBalaguer:
After several discussions with teams, it's not yet clear that a crowdloan can bid on subsequent auctions if and only if the first Lease Period of the crowdloan is the same or bigger than the first Lease Period of the auction.


alfarok (approved):
Thanks @SBalaguer :+1:

PR #4247, opened 2022-12-30
Wrong RPC path

stakeworld:
I think the path should be "consts.staking.maxNominatorRewardedPerValidator" not query. Kusama is currently 512 but because of the wrong path the wiki shows the wrong default of 256. Would default "0" not be better so errors show up quicker?

### DIFF
--- a/docs/maintain/maintain-guides-validator-payout.md
+++ b/docs/maintain/maintain-guides-validator-payout.md
@@ -196,8 +196,8 @@ validator, then you will lose 1 DOT in each case.
 :::caution
 
 If a validator is oversubscribed in an era, staking rewards are distributed only to the the top
-{{ polkadot: <RPC network="polkadot" path="query.staking.maxNominatorRewardedPerValidator" defaultValue={256}/> :polkadot }}
-{{ kusama: <RPC network="kusama" path="query.staking.maxNominatorRewardedPerValidator" defaultValue={256}/> :kusama }}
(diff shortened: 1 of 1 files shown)


docs/maintain/maintain-guides-validator-payout.md
```diff
@@ -196,8 +196,8 @@ validator, then you will lose 1 DOT in each case.
 :::caution
 
 If a validator is oversubscribed in an era, staking rewards are distributed only to the the top
-{{ polkadot: <RPC network="polkadot" path="query.staking.maxNominatorRewardedPerValidator" defaultValue={256}/> :polkadot }}
-{{ kusama: <RPC network="kusama" path="query.staking.maxNominatorRewardedPerValidator" defaultValue={256}/> :kusama }}
+{{ polkadot: <RPC network="polkadot" path="consts.staking.maxNominatorRewardedPerValidator" defaultValue={0}/> :polkadot }}
+{{ kusama: <RPC network="kusama" path="consts.staking.maxNominatorRewardedPerValidator" defaultValue={0}/> :kusama }}
```
alfarok:
```suggestion
{{ polkadot: <RPC network="polkadot" path="consts.staking.maxNominatorRewardedPerValidator" defaultValue={256}/> :polkadot }}
{{ kusama: <RPC network="kusama" path="consts.staking.maxNominatorRewardedPerValidator" defaultValue={512}/> :kusama }}
```

alfarok (approved):
Nice catch @stakeworld :+1: 

In terms of the defaults, we normally set them to the most recent value.  A console error should surface if the RPC connection fails but I agree they are not currently easy to track.

PR #4285, opened 2023-01-10
Update smallest unit of ETH to use correct name

taylorferran:
ETH's smallest unit of measurement is Wei, not Gwei. 
1 wei == 1
1 gwei == 1e9
1 ether == 1e18

We should compare 1 Planck to 1 Wei, not 1 Gwei.


DrW3RK (approved):
Agreed! Thank you!

PR #4286, opened 2023-01-10
Update "Avoid unresponsiveness" list

piffle-rack:
It looks like there was a copy-paste error somewhere and two list items got merged into one. I have split them into separate items.

I didn't redo the text flow in the Markdown source so as not to create spurious changes in the Git history.


DrW3RK (commented):
Thank you!

DrW3RK (approved):

PR #4296, opened 2023-01-13
Stylo is not actively maintained any more

Tbaut:
While Stylo has been very useful to many during the time when Parity Signer was not in its best shape, things have changed. Since the v5 Signer has been enhanced, and while I'd like to see more releases of it, metadata upgrade over QR code is a killer feature that I will not be able to add to Stylo. After several years of service, I've decided to discontinue Stylo's support.


DrW3RK (approved):
Thanks for the update!

PR #4309, opened 2023-01-16
Update `BlockHashCount` in Transaction Mortality section

Imod7:
### Description
In the [Transaction Mortality](https://wiki.polkadot.network/docs/build-protocol-info#transaction-mortality) section of the wiki it is mentioned that `BlockHashCount` is set to `2400` blocks. 

However at [line 81](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/lib.rs#L81) of the Polkadot repo it looks like it was increased to `4096`. The change happened in this [PR](https://github.com/paritytech/polkadot/pull/6037) from `2400` to `4096`.

Also changed the hours from `four` to `seven` by doing the following calculations (based on this [Polkadot Parameters](https://wiki.polkadot.network/docs/maintain-polkadot-parameters)) :
- 4096 * 6 (seconds) = 24.576
- 24.576 / 60 / 60 = 6.82 so approx 7h

### DIFF
--- a/docs/build/build-protocol-info.md
+++ b/docs/build/build-protocol-info.md
@@ -203,8 +203,8 @@ valid. If the extrinsic is not included in a block within this validity window, 
 from the transaction queue.
 
 The chain only stores a limited number of prior block hashes as reference. You can query this
-parameter, called `BlockHashCount`, from the chain state or metadata. This parameter is set to 2400
-blocks (about four hours) at genesis. If the validity period is larger than the number of blocks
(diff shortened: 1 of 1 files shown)


DrW3RK (approved):
Thank you!

Imod7:
Hello @DrW3RK My apologies but I pushed a second commit after you approved that uses the `RPC` call. Is that ok ? Otherwise I can revert back to the hard coded value.

DrW3RK:
All good. I was planning on doing it before merging! Thanks for fetching the value from the chain!

PR #4323, opened 2023-01-17
Create new RPC tests for all active paths

alfarok:
### Description
Addresses #4322 

Initial run results:
```
✅ Retrieves and applies a 'const' RPC value (1170 ms)
✅ Retrieves and applies a 'query' RPC value (1063 ms)
✅ RPC falls back to default (15 ms)
✅ Human readable filter with integer value (9 ms)
✅ Human readable filter with float value (8 ms)
✅ RPC Path Test: consts.system.blockHashCount (2226 ms)
✅ RPC Path Test: consts.balances.existentialDeposit (1133 ms)
✅ RPC Path Test: query.staking.validatorCount (1332 ms)
✅ RPC Path Test: consts.staking.maxNominatorRewardedPerValidator (1127 ms)
✅ RPC Path Test: consts.identity.basicDeposit (1139 ms)
✅ RPC Path Test: query.staking.currentEra (1215 ms)
✅ RPC Path Test: consts.phragmenElection.votingBondBase (1125 ms)
✅ RPC Path Test: query.staking.minNominatorBond (1118 ms)
✅ RPC Path Test: consts.staking.maxNominations (1057 ms)
✅ RPC Path Test: consts.democracy.votingPeriod (1018 ms)
✅ RPC Path Test: consts.crowdloan.minContribution (1128 ms)
✅ RPC Path Test: query.nominationPools.minJoinBond (1195 ms)
✅ RPC Path Test: consts.auctions.endingPeriod (1149 ms)
✅ RPC Path Test: consts.democracy.enactmentPeriod (1126 ms)
✅ RPC Path Test: consts.electionProviderMultiPhase.signedMaxSubmissions (1173 ms)
❌ RPC Path Test: query.multisig.depositBase (1077 ms)
✅ RPC Path Test: consts.proxy.maxProxies (1131 ms)
❌ RPC Path Test: query.nomiationPools.minJoinBond (1107 ms)
✅ RPC Path Test: consts.democracy.voteLockingPeriod (1051 ms)
❌ RPC Path Test: consts.assets.assetDeposit (1022 ms)
✅ RPC Path Test: consts.treasury.spendPeriod (1099 ms)
✅ RPC Path Test: consts.electionProviderMultiPhase.maxElectingVoters (1033 ms)
```

I addressed the failing tests, so running the suite now yields:
```
Test Suites: 1 passed, 1 total
Tests:       27 passed, 27 total
Snapshots:   0 total
Time:        30.67 s, estimated 35 s
Ran all test suites.
[...]

PR #4325, opened 2023-01-18
Update Chill Other section

### LINKED ISSUE #4318: Update Chill Other section on Polkadot Wiki
https://wiki.polkadot.network/docs/maintain-guides-how-to-chill#chill-other

https://paritytech.github.io/substrate/master/pallet_staking/struct.Pallet.html#method.chill_other

A MaxNominatorCount and MaxValidatorCount must be set which is used to determine how close we are to the threshold. As MaxNominatorCount is set to `None`, chill other can only be successfully called when  `controller` belongs to a nominator who has become non-decodable.

https://github.com/paritytech/substrate/blob/28937302694787cd1e6dbd846ff2ba4b1cf82025/frame/staking/src/pallet/mod.rs#L1752

DrW3RK:


alfarok (approved):
LGTM :+1:.  I am also going to add `query.staking.chillThreshold` to my open testing PR https://github.com/w3f/polkadot-wiki/pull/4323.  Right now it is a manual step to add new RPC paths for coverage but in the future we could scan the repo for all unique paths as part of the test so we don't have to remember to add each one manually.  I used a simple python script to extract them originally.

PR #4423, opened 2023-02-02
Add Chrome as a supported browser for Polkagate

Nick-1979:


DrW3RK (approved):
Thanks for the update!

PR #4440, opened 2023-02-06
Fix future pdk url in build-parachains.md

muddlebee:
Fix broken link for future pdk redirection
Previously : https://wiki.polkadot.network/docs/build-pdk##future-pdks
Fix :             https://wiki.polkadot.network/docs/build-pdk#future-pdks

image


DrW3RK (approved):
Thank you!

PR #4519, opened 2023-02-27
Fix link to tech stack

semuelle:
Not sure the link makes sense or should be relative, as it's the previous page in the sidebar. Anyway, the old URL is archived and points back to this wiki.


filippoweb3 (approved):
@semuelle thanks for this! Cheers!

PR #4543, opened 2023-03-05
Add missing bullet point for Cancel proxy

alan-null:
minor fix
image


filippoweb3 (approved):
right, good catch! Thanks!

PR #4561, opened 2023-03-08
Add caution on KSM Treasury page

DrW3RK:
Before we revamp the treasury and OpenGov docs on Kusama Guide, add cautionary notes to not use Gov v1 for Treasury proposals


filippoweb3 (approved):
lgtm, just a rendering issue that I fixed

PR #4563, opened 2023-03-08
Updated application form

paradox-tt:
This proposed change updates the application form link to direct users to the Polkadot form instead of the Kusama application form.  If this is confusing we could note both application forms regardless of the network.


filippoweb3 (approved):
thanks! I fixed a couple of rendering issues.

Issue #4541, opened 2023-03-04
[Feature Request] XCM Documentation Elaboration & Restructure


CrackTheCode016:
This issue is meant to keep track of any discussion and highlight the changes to be made to the [XCM](https://wiki.polkadot.network/docs/learn-xcm) section of the wiki.  

Currently, the wiki _does_ cover the most relevant concepts but could use much more elaboration and context for how to use XCM and where to look for resources.   

A set of guides centering around configuring XCM and relevant parameters would also be beneficial. 

As part of this effort, [`xcm-format`](https://github.com/paritytech/xcm-format) should also be updated, as it gives a good technical reference for the wiki to refer to when discussing instructions & registers.   I've already started a [refactor on my fork.](https://github.com/CrackTheCode016/xcm-format)

### Page Reorg - for "Learn"

In terms of reorganizing the docs, new pages should be added to highlight other aspects (similar to how [Parachains in the wiki are laid out](https://wiki.polkadot.network/docs/learn-parachains)): 

 - Introduction to XCM (Cross Consensus Messaging)
 - Transporting XCM Messages - XCMP, HRMP, & VMP
 - XCM Executor, XCVM, & the xcm-simulator
 - XCM Format Instruction & Register Specification

_...more may be added as needed!_

Each of these pages will hold domain-specific resources for further action (if applicable) and comprehensive info on how to use XCM.  

In the future, it may also be beneficial to include things more builder oriented in the "Build" section of the wiki.

### Outstanding issues 

- [x] Update XCM use cases to ensure accuracy (add as needed), add common ones
- [x] Cover XCM v3 changes
- [ ] Comprehensive table of XCM instructions and registers, what they do, and links to their source.
- [x] Deep dive into the XCM executor and XCVM
- [x] Document the xcm-pallet
- [ ] Document XCM transport methods
- [ ] Document how fees are handled in XCM
- [x] Document crucial XCM config aspects, i.e., transactors, traders, barriers, reserves
- [ ] Document XCM weight calculation
- [ ] Document how XCM messages are constructed for the means of transportation
- [x] Update `xcm-format` 
- [ ] Need more coverage on Asset trapping / claiming mechanisms
- [x] Discussing trust between chains, and how `xcm-config` reflects the trust levels.
- [x] Differences between teleportation, reserves, and which to use.
- [x] Covering `xcm-builder`

_...more may be added as needed!_

DrW3RK:
[...]

PR #4578, opened 2023-03-14
XCM Doc Restructure: Covering pallet-xcm

CrackTheCode016:
As part of issue https://github.com/w3f/polkadot-wiki/issues/4541, this PR covers and documents `pallet-xcm`:

## TODO
- [x] What `pallet-xcm` does
- [x] How `pallet-xcm` handles fees
- [x] What extrinsics are important to note in `pallet-xcm` (and what they do)
- [x] How `pallet-xcm` defines interfaces for `XcmConfig`
- [x] Transfer Reserve vs Teleport difference

### DIFF
--- a/docs/learn/learn-xcm-pallet.md
+++ b/docs/learn/learn-xcm-pallet.md
@@ -10,8 +10,8 @@ slug: ../learn-xcm-pallet
 The XCM pallet ([`pallet-xcm`](https://github.com/paritytech/polkadot/tree/master/xcm/pallet-xcm))
 provides a developer-friendly interface for most of the common XCM messages.
 
-This pallet provides some default implementations for some traits required by `XcmConfig`, as well
-as an instance of the XCM Executor provided as a trait within the pallet's own configuration.
+This pallet provides some default implementations for traits required by `XcmConfig`, as well the
+`ExecuteXcm` trait over the pallet's own configuration.  
 `pallet-xcm` provides a default interface in the form of a pallet, that can manage and deal with
 XCM-related storage and higher-level dispatchable functions.
 
@@ -21,7 +21,7 @@ local or external chains. `pallet-xcm`'s functionality is separated into three c
 :::note
 
 Remember, all XCM messages are effectively XCVM programs that contain a set of instructions. It is
-the job of the XCM exector to handle these programs, and their subsequent instructions as needed.
+the job of the XCM executor is to handle and execute these programs.
 
 :::
 
--- a/docs/learn/learn-xcm-pallet.md
+++ b/docs/learn/learn-xcm-pallet.md
@@ -11,7 +11,8 @@ The XCM pallet ([`pallet-xcm`](https://github.com/paritytech/polkadot/tree/maste
 provides a developer-friendly interface for most of the common XCM messages.
 
 This pallet provides default implementations for many traits required by `XcmConfig`, which is
-highly useful for configuration.
+highly useful for configuration. `pallet-xcm` provides a default interface, in the form of a pallet,
+that can manage and deal with XCM-related storage and dispatchable functions.
 
 It defines a set of extrinsics that can be utilized to build XCVM programs, either to target the
 local or external chains. Pallet-xcm's functionality is separated into three categories:
@@ -19,3 +20,42 @@ local or external chains. Pallet-xcm's functionality is separated into three cat
 1. Primitive, dispatchable functions to locally execute an XCM message.
 2. High-level, dispatchable functions for asset transfers.
 3. Version negotiation-specific dispatchable functions.
+
+## Primitive Extrinsics
+
+There are two primary primitive extrinsics. These extrinsics handle sending and executing XCVM
+programs as dispatchable functions within the pallet.
+
+1. `execute` - This call is a direct access to the XCM executor. It checks the origin, message, and
+   ensures no barrier/filter will block the execution of the XCM message. Once it's deemed valid,
+   the message will then be _locally_ executed, therein returning the outcome as an event. This
+   operation is executed on behalf of whichever account has signed the extrinsic.
+2. `send` - This call specifies where a message should be sent externally to a particular
+   destination, i.e another parachain. It checks the origin, destination, and the message, and is
+   then sent to the `XcmRouter`.
+
+## Asset Transfer Extrinsics
+
+There are several extrinsics that handle asset transfer logic. They define a redetermined set of
+instructions in which to send and execute XCM messages. There are two variants of these functions,
+prefixed with `limited_`. These dispatchables have the same functionality, but with the option of
+specifying a weight to pay for the XCM fee.
+
+Otherwise, the fee is taken as needed from the asset being transferred.
+
+1. `reserve_transfer_assets` - Transfer some assets from the local chain to the sovereign account of
+   a destination chain and forward a notification XCM.
+2. `teleport_assets` - Teleport some assets from the local chain to some destination chain.
+
+## Version Negotiation Extrinsics
+
+These extrinisics require root, as they are only used when bypassing XCM version negotiation.
+
+1. `force_xcm_version` - Modifies the `SupportedVersion` storage to change a particular
+   destination's stated XCM version.
+2. `force_default_xcm_version` - Modifies the `SafeXcmVersion` storage, which stores the default
+   version to use when the destination's version is unknown.
+3. `force_subscribe_version_notify` - Sends an XCM message with a `SubscribeVersion` instruction to
+   a destination.
+4. `force_unsubscribe_version_notify` - Sends an XCM message with a `UnsubscribeVersion` instruction
+   to a destination.
--- a/docs/learn/learn-xcm-pallet.md
+++ b/docs/learn/learn-xcm-pallet.md
@@ -92,8 +92,8 @@ While both extrinsics deal with transferring assets, they exhibit fundamentally 
 :::info
 
 A sovereign account refers to an account within a particular consensus system. Even though accounts
-may be different in terms of factors such as address, XCM agnostic nature enables for communication
-between these sovereign accounts that are in different consensus systems.
+may be different in terms of factors such as an address format, XCM agnostic nature enables
+communication between these sovereign accounts that are in other consensus systems.
 
 :::
 
@@ -119,7 +119,7 @@ applicable, fees are withdrawn from the assets from the specified `MultiLocation
 payment to execute any subsequent instructions within the XCM.
 
 Fees are generally dependent on several factors within the `XcmConfig`. For example, the barrier may
-negate any fees be paid at all.
+negate any fees to be paid at all.
 
 Before any XCM is sent, and if the destination chain’s barrier requires it, a `BuyExecution`
 instruction is used to buy the necessary weight for the XCM. XCM fee calculation is handled by the
--- a/docs/learn/learn-xcm-pallet.md
+++ b/docs/learn/learn-xcm-pallet.md
@@ -10,11 +10,12 @@ slug: ../learn-xcm-pallet
 The XCM pallet ([`pallet-xcm`](https://github.com/paritytech/polkadot/tree/master/xcm/pallet-xcm))
 provides a set of pre-defined, commonly used XCVM programs in the form of a set of extrinsics.
 
-This pallet provides some default implementations for traits required by `XcmConfig`, as well the
-`ExecuteXcm` trait over the pallet's own configuration. Where the XCM format defines a set of
-instructions used to construct XCVM programs, `pallet-xcm` defines a set of extrinsics that can be
-utilized to build XCVM programs, either to target the local or external chains. `pallet-xcm`'s
-functionality is separated into three categories:
+This pallet provides some default implementations for traits required by `XcmConfig`. The XCM
+executor is also included as an associated type within the pallet's configuration.
+
+Where the XCM format defines a set of instructions used to construct XCVM programs, `pallet-xcm`
+defines a set of extrinsics that can be utilized to build XCVM programs, either to target the local
+or external chains. `pallet-xcm`'s functionality is separated into three categories:
 
 :::note
 
@@ -108,6 +109,18 @@ They change any relevant storage aspects that enforce anything to do with XCM ve
 
 ## Fees in the XCM Pallet
 
-Message fees are only paid if the interior location is **not** equal to the interpreting consensus
-system (known as `Here` in the context of an XCM Multilocation). Otherwise, the chain bears the
-fees. If applicable, fees are withdrawn from the assets from the specified `MultiLocation`.
+Message fees are only paid if the interior location does not equal the interpreting consensus system
+(known as Here in the context of an XCM `Multilocation`). Otherwise, the chain bears the fees. If
+applicable, fees are withdrawn from the assets from the specified `MultiLocation` and used as
+payment to execute any subsequent instructions within the XCM.
+
+Fees are generally dependent on several factors within the `XcmConfig`. For example, the barrier may
+negate any fees be paid at all.
+
+Before any XCM is sent, and if the destination chain’s barrier requires it, a `BuyExecution`
+instruction is used to buy the necessary weight for the XCM. XCM fee calculation is handled by the
+Trader, which iteratively calculates the total fee based on the number of instructions.
+
+The Trader used to calculate the weight (time for computation in consensus) to include in the
+message. Fee calculation in XCM is highly configurable and, for this reason, subjective to whichever
+configuration is in place.
--- a/docs/learn/learn-xcm-pallet.md
+++ b/docs/learn/learn-xcm-pallet.md
@@ -94,7 +94,6 @@ They change any relevant storage aspects that enforce anything to do with XCM ve
 
 ## Fees in the XCM Pallet
 
-Fees for messages are only paid if given interior location is **not** equal to the interpreting
-consensus system (known as `Here` in the context of an XCM Multilocation). Otherwise, the chain
-bears the fees. Fees are withdrawn from the assets from the specified `MultiLocation`, if
-applicable.
+Message fees are only paid if the interior location is **not** equal to the interpreting consensus
+system (known as `Here` in the context of an XCM Multilocation). Otherwise, the chain bears the
+fees. If applicable, fees are withdrawn from the assets from the specified `MultiLocation`.
--- a/docs/learn/learn-xcm-pallet.md
+++ b/docs/learn/learn-xcm-pallet.md
@@ -11,11 +11,19 @@ The XCM pallet ([`pallet-xcm`](https://github.com/paritytech/polkadot/tree/maste
 provides a developer-friendly interface for most of the common XCM messages.
 
 This pallet provides some default implementations for some traits required by `XcmConfig`, as well
-as an instance of the XCM Executor. `pallet-xcm` provides a default interface in the form of a
-pallet, that can manage and deal with XCM-related storage and dispatchable functions.
+as an instance of the XCM Executor provided as a trait within the pallet's own configuration.
+`pallet-xcm` provides a default interface in the form of a pallet, that can manage and deal with
+XCM-related storage and dispatchable functions.
 
 It defines a set of extrinsics that can be utilized to build XCVM programs, either to target the
 local or external chains. `pallet-xcm`'s functionality is separated into three categories:
+
+:::note
+
+Remember, all XCM messages are effectively XCVM programs that contain a set of instructions. It is
+the job of the XCM exector to handle these programs, and their subsequent instructions as needed.
+
+:::
 
 1. Primitive, dispatchable functions to locally execute an XCM message.
 2. High-level, dispatchable functions for asset transfers.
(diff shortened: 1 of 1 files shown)


docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
-This pallet provides default implementations for many traits required by `XcmConfig`, which is
-highly useful for configuration.
+This pallet provides some default implementations for some traits required by `XcmConfig`, as well
+as an instance of the XCM Executor provided as a trait within the pallet's own configuration.
+`pallet-xcm` provides a default interface in the form of a pallet, that can manage and deal with
+XCM-related storage and higher-level dispatchable functions.
 
 It defines a set of extrinsics that can be utilized to build XCVM programs, either to target the
-local or external chains. Pallet-xcm's functionality is separated into three categories:
+local or external chains. `pallet-xcm`'s functionality is separated into three categories:
```
joepetrowski: The XCM format defines a set of _instructions_ that can be utilized to build XCVM programs. The _extrinsics_ are pre-written, commonly used programs.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
 The XCM pallet ([`pallet-xcm`](https://github.com/paritytech/polkadot/tree/master/xcm/pallet-xcm))
 provides a developer-friendly interface for most of the common XCM messages.
 
-This pallet provides default implementations for many traits required by `XcmConfig`, which is
-highly useful for configuration.
+This pallet provides some default implementations for some traits required by `XcmConfig`, as well
+as an instance of the XCM Executor provided as a trait within the pallet's own configuration.
```
KiChjang: If it's provided as a trait, then it's not providing an instance, because by its very nature, traits are abstract concepts and not concrete implementations.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+There are two primary primitive extrinsics. These extrinsics handle sending and executing XCVM
+programs as dispatchable functions within the pallet.
+
+1. `execute` - This call is direct access to the XCM executor. It checks the origin and message and
+   ensures that no barrier/filter will block the execution of the XCM message. Once it is deemed
+   valid, the message will then be _locally_ executed, therein returning the outcome as an event.
+   This operation is executed on behalf of whichever account has signed the extrinsic. It's possible
+   for only a partial execution to occur.
+2. `send` - This call specifies where a message should be sent externally to a particular
+   destination, i.e., another parachain. It checks the origin, destination, and message and is then
```
joepetrowski: Not necessarily another parachain. It could also be a smart contract on the local chain, or another consensus system entirely.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+   sent to the `XcmRouter`.
+
+:::info
+
+The XCM pallet needs the `XcmRouter` to send messages. It is used to dictate where messages are
+allowed to be sent, and which XCM transport protocol to use. For example, Kusama, the canary
+network, uses the `ChildParachainRoute` which only allows for Downward Message Passing from the
+relay to parachains to occur.
+
+You can read more about [XCM transport methods here.](./learn-xcm-transport.md)
```
joepetrowski:
```suggestion
You can read more about [XCM transport methods here](./learn-xcm-transport.md).
```

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
 The XCM pallet ([`pallet-xcm`](https://github.com/paritytech/polkadot/tree/master/xcm/pallet-xcm))
 provides a developer-friendly interface for most of the common XCM messages.
 
-This pallet provides default implementations for many traits required by `XcmConfig`, which is
-highly useful for configuration.
+This pallet provides some default implementations for some traits required by `XcmConfig`, as well
+as an instance of the XCM Executor provided as a trait within the pallet's own configuration.
+`pallet-xcm` provides a default interface in the form of a pallet, that can manage and deal with
+XCM-related storage and higher-level dispatchable functions.
```
KiChjang: What is this talking about? Default interface between which two systems? And what kind of XCM-related storage?

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+instructions for sending and executing XCM messages. Two variants of these functions are prefixed
+with `limited_`. They have the same functionality but can specify a weight to pay for the XCM fee.
+
+Otherwise, the fee is taken as needed from the asset being transferred.
+
+1. `reserve_transfer_assets` - Transfer some assets from the local chain to the sovereign account of
+   a destination chain and forward a notification XCM.
+2. `teleport_assets` - Teleport some assets from the local chain to some destination chain.
+
+### Transfer Reserve vs Teleport
```
joepetrowski:
```suggestion
### Transfer Reserve vs. Teleport
```

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+
+While both extrinsics deal with transferring assets, they exhibit fundamentally different behavior.
+
+- **Teleporting** an asset implies a two-step process: the asset is burned/destroyed in the origin
+  chain and re-minted to whatever account is specified at the destination. Teleporting should only
+  occur if there is an inherent and bilateral trust between the two chains, as the tokens destroyed
+  _could not_ necessarily be guaranteed to have the same properties when minted at the destination.
+- **Transferring** or **reserving** an asset implies the movement of funds from one authority to
+  another, in this case, from the Origin of the sender's chain and into the sovereign account (who
+  will ultimately benefit from the assets) on the destination chain.
```
joepetrowski: This is not really clear

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+`pallet-xcm` provides a default interface in the form of a pallet, that can manage and deal with
+XCM-related storage and higher-level dispatchable functions.
 
 It defines a set of extrinsics that can be utilized to build XCVM programs, either to target the
-local or external chains. Pallet-xcm's functionality is separated into three categories:
+local or external chains. `pallet-xcm`'s functionality is separated into three categories:
+
+:::note
+
+Remember, all XCM messages are effectively XCVM programs that contain a set of instructions. It is
```
KiChjang:
```suggestion
Remember, all XCMs are effectively XCVM programs that contain a set of instructions. It is
```

I know we say "XCM messages" colloquially, but do remember that the term is an oxymoron, because the M in XCM already stands for message.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+the job of the XCM exector to handle these programs, and their subsequent instructions as needed.
```
KiChjang:
```suggestion
the job of the XCM executor to handle these programs.
```

I'd probably delete that entire sentence because it's unclear what the subsequent instructions are.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+3. `force_subscribe_version_notify` - Sends an XCM message with a `SubscribeVersion` instruction to
+   a destination.
+4. `force_unsubscribe_version_notify` - Sends an XCM message with a `UnsubscribeVersion` instruction
+   to a destination.
+
+## Fees in the XCM Pallet
+
+Message fees are only paid if the interior location is **not** equal to the interpreting consensus
+system (known as `Here` in the context of an XCM Multilocation). Otherwise, the chain bears the
+fees. If applicable, fees are withdrawn from the assets from the specified `MultiLocation`.
```
joepetrowski: Not really, there are a lot of barrier conditions for when to allow unpaid execution, and there are other ways to pull fees into the holding register, like claiming a trap.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+
+:::
 
 1. Primitive, dispatchable functions to locally execute an XCM message.
 2. High-level, dispatchable functions for asset transfers.
 3. Version negotiation-specific dispatchable functions.
+
+## Primitive Extrinsics
+
+There are two primary primitive extrinsics. These extrinsics handle sending and executing XCVM
```
KiChjang, line 35: Is "primary primitive extrinsic" a technical term? Who designated `execute` and `send` as "primary primitive extrinsics"?

docs/learn/learn-xcm-pallet.md
CrackTheCode016: It was meant to show that the pallet is used to provide a common/default interface for developers to interact with XCM, but perhaps this wording is incorrect. 

> And what kind of XCM-related storage?

Storage related to versioning, i.e `SafeXcmVersion`.

docs/learn/learn-xcm-pallet.md
CrackTheCode016: I think it would be preferable to use XCMs over XCM messages, I was debating which to use throughout.

docs/learn/learn-xcm-pallet.md
CrackTheCode016: Within the [v2 messaging audit](https://blog.quarkslab.com/resources/2022-02-27-xcmv2-audit/21-12-908-REP.pdf), they were called 'primitive' functions.

docs/learn/learn-xcm-pallet.md
CrackTheCode016: I think the notion that XCMs == XCVM programs need to be pushed more, as sometimes the nomenclature implies they are something akin to an arbitrary packet (rather than a set of instructions that need a VM).

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+programs as dispatchable functions within the pallet.
+
+1. `execute` - This call is direct access to the XCM executor. It checks the origin and message and
+   ensures that no barrier/filter will block the execution of the XCM message. Once it is deemed
```
KiChjang: I don't think the `execute` extrinsic does this -- the XCM executor does.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+   for only a partial execution to occur.
+2. `send` - This call specifies where a message should be sent externally to a particular
+   destination, i.e., another parachain. It checks the origin, destination, and message and is then
+   sent to the `XcmRouter`.
+
+:::info
+
+The XCM pallet needs the `XcmRouter` to send messages. It is used to dictate where messages are
+allowed to be sent, and which XCM transport protocol to use. For example, Kusama, the canary
+network, uses the `ChildParachainRoute` which only allows for Downward Message Passing from the
```
KiChjang:
```suggestion
network, uses the `ChildParachainRouter` which only allows for Downward Message Passing from the
```

docs/learn/learn-xcm-pallet.md
CrackTheCode016: > - **Transferring** or **reserving** an asset implies that **equivalent** assets (i.e, native currency) are withdrawn from the ownership of the current origin and deposited into the sovereign account on the destination chain.  Unlike teleporting an asset, it is not destroyed, rather a trusted entity is used (i.e., Statemint on Polkadot) to **reserve** the assets, wherein the sovereign account on the destination chain obtains ownership of them.

Is this clearer? Might be worth creating a diagram for to show the difference.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+2. `teleport_assets` - Teleport some assets from the local chain to some destination chain.
+
+### Transfer Reserve vs Teleport
+
+While both extrinsics deal with transferring assets, they exhibit fundamentally different behavior.
+
+- **Teleporting** an asset implies a two-step process: the asset is burned/destroyed in the origin
+  chain and re-minted to whatever account is specified at the destination. Teleporting should only
+  occur if there is an inherent and bilateral trust between the two chains, as the tokens destroyed
+  _could not_ necessarily be guaranteed to have the same properties when minted at the destination.
```
KiChjang: This isn't the fundamental problem -- the more fundamental problem is whether or not the two chains can trust each other to have burnt or minted the funds beforehand. A chain can simply send the `ReceiveTeleportedAsset` XCM instruction without previously having burnt any.

docs/learn/learn-xcm-pallet.md
CrackTheCode016: Would it be more accurate to assume that the pallet is a 'wrapper' for the `xcm-executor`?

docs/learn/learn-xcm-pallet.md
CrackTheCode016: yes, I meant for it to be executor - will adjust to be clearer. The `execute` extrinsic merely calls the executor (`T::XcmExecutor::execute_xcm_in_credit`).

docs/learn/learn-xcm-pallet.md
CrackTheCode016: I was planning to cover this in another part of the Wiki, however I'll add some context around those scenarios here as well.

docs/learn/learn-xcm-pallet.md
```diff
@@ -10,12 +10,90 @@ slug: ../learn-xcm-pallet
...
+## Asset Transfer Extrinsics
+
+Several extrinsics within the pallet handle asset transfer logic. They define a predetermined set of
+instructions for sending and executing XCM messages. Two variants of these functions are prefixed
+with `limited_`. They have the same functionality but can specify a weight to pay for the XCM fee.
+
+Otherwise, the fee is taken as needed from the asset being transferred.
+
+1. `reserve_transfer_assets` - Transfer some assets from the local chain to the sovereign account of
+   a destination chain and forward a notification XCM.
```
KiChjang:
```suggestion
   a destination chain and forward an XCM containing a `ReserveAssetDeposited` instruction.
```

docs/learn/learn-xcm-pallet.md
KiChjang: It should say that it is an interface between FRAME and XCM. I would not even say that it is the default, because the transport pallets all require specifying an XCM executor.

docs/learn/learn-xcm-pallet.md
KiChjang: Is the term equivalent asset defined anywhere else other than here?

docs/learn/learn-xcm-pallet.md
```diff
@@ -8,14 +8,119 @@ slug: ../learn-xcm-pallet
...
+1. Primitive, dispatchable functions to locally execute an XCM.
 2. High-level, dispatchable functions for asset transfers.
 3. Version negotiation-specific dispatchable functions.
+
+## Primitive Extrinsics
+
+There are two primary primitive extrinsics. These extrinsics handle sending and executing XCVM
+programs as dispatchable functions within the pallet.
+
+1. `execute` - This call contains direct access to the XCM executor. It is tje job of the executor
```
KiChjang:
```suggestion
1. `execute` - This call contains direct access to the XCM executor. It is the job of the executor
```

docs/learn/learn-xcm-pallet.md
```diff
@@ -8,14 +8,119 @@ slug: ../learn-xcm-pallet
...
+   to check the message and ensure that no barrier/filter will block the execution of the XCM. Once
+   it is deemed valid, the message will then be _locally_ executed, therein returning the outcome as
+   an event. This operation is executed on behalf of whichever account has signed the extrinsic.
+   It's possible for only a partial execution to occur.
+2. `send` - This call specifies where a message should be sent externally to a particular
+   destination, i.e. a parachain, smart contract, or any system which is governed by consensus. As
+   before with `execute`, the Executor checks the origin, destination, and message and is then sent
```
KiChjang: No? I don't think the local XCM executor is consulted before the XCM is sent.

KiChjang: https://github.com/paritytech/polkadot/blob/5feae5d7e5630e9e3837a2f9517c8ae0a376dea4/xcm/pallet-xcm/src/lib.rs#LL769C5-L769C5

CrackTheCode016: A mistake on my part - only for local execution is the executor called. Send merely sends it _to be_ executed on the destination.

docs/learn/learn-xcm-pallet.md
```diff
@@ -8,14 +8,119 @@ slug: ../learn-xcm-pallet
...
+1. `reserve_transfer_assets` - Transfer some assets from the local chain to the sovereign account of
+   a destination chain and forward an XCM containing a `ReserveAssetDeposited` instruction, which
+   serves as a notification.
+2. `teleport_assets` - Teleport some assets from the local chain to some destination chain.
+
+### Transfer Reserve vs. Teleport
+
+While both extrinsics deal with transferring assets, they exhibit fundamentally different behavior.
+
+- **Teleporting** an asset implies a two-step process: the asset is burned/destroyed in the origin
```
KiChjang: They're not necessarily always burnt or destroyed; what is always true though is that the assets being teleported are removed from active circulation.

docs/learn/learn-xcm-pallet.md
```diff
@@ -8,14 +8,119 @@ slug: ../learn-xcm-pallet
...
+  chain and re-minted to whatever account is specified at the destination. Teleporting should only
+  occur if there is an inherent and bilateral trust between the two chains, as the tokens destroyed
+  _could not_ necessarily be guaranteed to have the same properties when minted at the destination,
+  as well as that there has to be **trust** that the a particular chain burned, or re-minted the
+  assets.
+- **Transferring** or **reserving** an asset implies that **equivalent** assets (i.e, native
+  currency, like `DOT` or `KSM`) are withdrawn from _sovereign account_ of the origin chain and
+  deposited into the sovereign account on the destination chain. Unlike teleporting an asset, it is
+  not destroyed and re-minted, rather a trusted, 3rd entity is used (i.e., Statemint on Polkadot) to
+  **reserve** the assets, wherein the sovereign account on the destination chain obtains ownership
```
KiChjang: The sovereign account is located on the reserve chain, not the destination chain.

KiChjang: I think you meant to say:
```suggestion
  **reserve** the assets, wherein the sovereign account of the destination chain on the reserve chain obtains ownership
```

docs/learn/learn-xcm-pallet.md
```diff
@@ -8,14 +8,119 @@ slug: ../learn-xcm-pallet
...
+
+## Version Negotiation Extrinsics
+
+The following extrinsics require root, as they are only used when bypassing XCM version negotiation.
+They change any relevant storage aspects that enforce anything to do with XCM version negotiations.
+
+1. `force_xcm_version` - Modifies the `SupportedVersion` storage to change a particular
+   destination's stated XCM version.
+2. `force_default_xcm_version` - Modifies the `SafeXcmVersion` storage, which stores the default
+   version when the destination's version is unknown.
```
KiChjang:
```suggestion
   XCM version to use when the destination's version is unknown.
```

docs/learn/learn-xcm-pallet.md
CrackTheCode016: Under the `xcm-format` repo, it's used a few times: https://github.com/paritytech/xcm-format#transferreserveasset

Is it worth noting that this asset is representative of the origin chain? i.e, DOT becoming known as "wrapped" DOT on the destination chain, or is that inaccurate?

docs/learn/learn-xcm-pallet.md
CrackTheCode016: I guess this is subjective to how one implements [`check_out`](https://github.com/paritytech/polkadot/blob/master/xcm/xcm-executor/src/traits/transact_asset.rs#L66), but I will add that detail.

docs/learn/learn-xcm-pallet.md
CrackTheCode016: Reworded 👍

CrackTheCode016:
Just a sidenote - the goal of this portion is to cover some concepts that will be covered in more depth later in the wiki, (example: Trader is mentioned but not explained in depth quite yet).  Generally, I think it's good to showcase pallet-xcm initially as an example of where to start when looking into the more technical aspects of understanding XCM.

docs/learn/learn-xcm-pallet.md
```diff
@@ -8,14 +8,124 @@ slug: ../learn-xcm-pallet
...
+   chain.
+
+:::info
+
+The XCM pallet needs the `XcmRouter` to send XCMs. It is used to dictate where XCMs are allowed to
+be sent, and which XCM transport protocol to use. For example, Kusama, the canary network, uses the
+`ChildParachainRouter` which only allows for Downward Message Passing from the relay to parachains
+to occur.
+
+You can read more about [XCM transport methods here](./learn-xcm-transport.md).
```
KiChjang:
```suggestion
You can read more about [XCM transport mechanisms here](./learn-xcm-transport.md).
```
Or "protocols", but not "methods" as it conflicts with the naming of class/object methods.

CrackTheCode016: Used "protocols", as we reference them as such in other places in the wiki.

CrackTheCode016:
@KiChjang @joepetrowski Thank you both for the review!

If it looks good, I think it would be ready to polish slightly and merge.  In subsequent XCM sections, more content and detail can be provided.

DrW3RK (approved):

PR #4579, opened 2023-03-15
XCM Doc Restructure: Covering XCVM

CrackTheCode016:
As part of the issue https://github.com/w3f/polkadot-wiki/issues/4541, this PR covers and documents XCVM, executor, and xcm-simulator.

## TODO

- [x] XCM Message anatomy
- [x] XCVM Coverage, relation to xcm-executor and xcm-builder
- [x] Instruction types and explaination
- ~~[ ] Fee and weight handling~~
- [x] Key vocabulary documentation (`Instruction`, `Multilocation`, `Junction`, etc)


docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
+the program's goal, whether it is to teleport assets from one chain to another or call a smart
+contract on another chain, XCMs usually require changes to the registers before any changes to the
+consensus system can be made.
+
+## XCM Executor & Configuration
+
+The XCM Executor's implementation centers around a core piece: the XCM configuration. Each instance
+of the Executor must have a valid configuration, which specifies a multitude of options on how a
+chain may treat incoming messages via
+[Barriers](https://github.com/paritytech/polkadot/blob/master/xcm/xcm-executor/src/config.rs#L52),
```
profradha: Always include permalinks for the GitHub links included in the Wiki

DrW3RK (commented):
I think one of the XCM devs needs to review this before we merge. The content is informative and I was able to follow it as a reader.

docs/learn/learn-xcvm.md
CrackTheCode016: Done!

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
+    register used in many instructions is the `Holding` register, which places assets in a holding
+    until fulfilled on the destination chain.
+3.  The resulting instructions modify the Holding register as needed and execute extrinsics on the
+    destination chain to perform the state change. The Executor is able to convert incoming
+    instructions into FRAME-compatible origins, which enables the actual state changes to take
+    place.
+
+### Example Register: The Holding Register
+
+There are many instructions that depend on the Holding register. The Holding register is an XCM
```
franciscoaguirre:
```suggestion
There are many instructions that depend on the Holding register. The Holding register is an XCVM
```

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
+register that provides a place for any assets that are in an intermediary state to be held until
+they are withdrawn to a beneficiary. They require an instruction to place assets within and another
```
franciscoaguirre: I think using the word "withdraw" for removing assets from the holding register might be confusing since the instructions usually call it `Deposit` and `Withdraw` is the opposite operation

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
+chain may treat incoming messages via
+[Barriers](https://github.com/paritytech/polkadot/blob/26b0c4f6273190f4538b24939a56b6a0b51a344c/xcm/xcm-executor/src/config.rs#L52),
+calculate weight for a message via the
+[Weigher](https://github.com/paritytech/polkadot/blob/26b0c4f6273190f4538b24939a56b6a0b51a344c/xcm/xcm-executor/src/config.rs#L55),
+how much weight to purchase via the
+[Trader](https://github.com/paritytech/polkadot/blob/26b0c4f6273190f4538b24939a56b6a0b51a344c/xcm/xcm-executor/src/config.rs#L58),
+[configure fees](https://github.com/paritytech/polkadot/blob/26b0c4f6273190f4538b24939a56b6a0b51a344c/xcm/xcm-executor/src/config.rs#L89),
+how to
+[convert origins](https://github.com/paritytech/polkadot/blob/26b0c4f6273190f4538b24939a56b6a0b51a344c/xcm/xcm-executor/src/config.rs#L40),
+and more.
```
franciscoaguirre, line 53: We should have a section in the wiki for all the possible configurations. I don't think we have one right now. A short introduction here is the way to go though

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
-whom/where the asset is to be transferred.
+1. **Instruction** - Results in a state change in the local consensus system or some state change.
+2. **Trusted Indication** - Tells the XCVM, or the Executor, that some action has been done before
+   already - meaning, this action is now trusted and can be acted on, i.e., in a teleport scenario.
+3. **Information** - Provides additional information about a particular origin, usually the result
+   of a query, i.e., a `QueryResponse` instruction.
+4. **System Notification** - Typically used in the context of when an HRMP channel is being opened,
+   closed, or accepted.
 
+Typically, an XCM takes the following path:
```
franciscoaguirre: This doesn't look like a series of steps an XCM goes through on the XCVM. It should be something like:
1. An XCM instruction is read
2. The instruction is executed. This means that the current values of the XCVM registers, the instruction type, and the instruction operands are all used to execute some operation, which might result in some registers changing their value, or in an error being thrown, which halts execution
3. The next instruction is read and so on until the end of the message

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
+register that provides a place for any assets that are in an intermediary state to be held until
+they are withdrawn to a beneficiary. They require an instruction to place assets within and another
+to withdraw them. The simplest example of this occurring is the `DepositAsset` instruction, which in
+its Rust form looks like this:
+
+```rust
+enum Instruction {
+    DepositAsset {
+        assets: MultiAssetFilter,
+        max_assets: u32,
```
franciscoaguirre: This operand was removed in v3

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,108 @@ slug: ../learn-xcvm
...
 ```
+
+This instruction specifies how many assets are going to be withdrawn (useful for fee calculation),
+which assets to filter from the Holding register, and the beneficiary (the recipient of) the assets.
+It is very common for instructions to remove and place assets into the Holding register when
+transacting between chains.
+
+### Example: TransferAsset
+
+An example below illustrates how a chain may transfer assets locally using an XCM. In this message,
```
franciscoaguirre: It may also transfer assets remotely. I mean, "locally" on a remote chain. I think the wiki calls it remote transfers

docs/learn/learn-xcvm.md
CrackTheCode016: Ah, so to be clear: to _deposit_ means to take from the holding register, while withdraw means to withdraw assets _to_ the holding register?

docs/learn/learn-xcvm.md
CrackTheCode016: Agreed - in the works after the high level stuff is done :D

docs/learn/learn-xcvm.md
CrackTheCode016: I suppose it covers more than the XCVM - will modify as needed

CrackTheCode016:
@franciscoaguirre review addressed - LMK if anything else needs to be changed!

docs/learn/learn-xcvm.md
franciscoaguirre: Yeah, at least that's the logic with the instructions `DepositAsset` and `WithdrawAsset`. They act on accounts, where they `deposit` or `withdraw` but as a side effect end up taking from or putting in the holding register respectively.

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
-executor following the XCVM specification is provided by Parity, and it can be extended or
-customized, or even ignored altogether and users can create their own construct that follows the
-XCVM spec.
-
-A _message_ in XCM is simply just a programme that runs on the `XCVM`: in other words, one or more
-XCM instructions. To learn more about the XCVM and the XCM Format, see the latest
-[blog post](https://medium.com/polkadot-network/xcm-the-cross-consensus-message-format-3b77b1373392)
+program, referred to as an **"XCM"** or **"XCMs"** for multiple messages. The XCVM is a
+register-based state machine. The state is tracked in domain-specific registers that can hold
+information regarding the temporary state of a particular message. Most of the XCM format comprises
```
franciscoaguirre:
```suggestion
register-based state machine. The state is tracked in domain-specific registers that hold
information that is used and mutated along the execution of a particular message. Most of the XCM format comprises
```

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
+The primary implementation of the XCVM is the
+[`xcm-executor`](https://github.com/paritytech/polkadot/tree/26b0c4f6273190f4538b24939a56b6a0b51a344c/xcm/xcm-executor).
+It follows the XCVM specification provided by Parity. It's engineered to be extendable, providing
+maximum customizability when configuring XCM. Because the `xcm-executor` is just an implementation
+of XCVM, it's entirely possible to create another implementation if desired.
+
+## XCMs are XCVM Programs
+
+A cross consensus message (XCM) is just a program that runs on the `XCVM`: in other words, one or
+more XCM instructions that are executed by the `xcm-executor`. To learn more about the XCVM and the
```
franciscoaguirre:
```suggestion
more XCM instructions that are executed by an XCVM implementation, such as the `xcm-executor`. To learn more about the XCVM and the
```

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
+2. **Trusted Indication** - Tells the XCVM, or the Executor, that some action has been done before
+   already - meaning, this action is now trusted and can be acted on, i.e., in a teleport scenario.
+3. **Information** - Provides additional information about a particular origin, usually the result
+   of a query, i.e., a `QueryResponse` instruction.
+4. **System Notification** - Typically used in the context of when an HRMP channel is being opened,
+   closed, or accepted.
 
+Typically, an XCM takes the following path through the XCVM:
+
+1.  Instructions within an XCM are read one-by-one by the XCVM. An XCM may old one or more
```
franciscoaguirre:
```suggestion
1.  Instructions within an XCM are read one-by-one by the XCVM. An XCM may contain one or more
```

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
+    might result in some registers changing their value, or in an error being thrown, which would
+    halt execution.
+3.  Each subsequent instruction within the XCM is read until the end of the message has been
+    reached.
+
+### Example Register: The Holding Register
+
+There are many instructions that depend on the Holding register. The Holding register is an XCVM
+register that provides a place for any assets that are in an intermediary state to be held until
+they are deposited to the beneficiary. They require an instruction to place assets within and
```
franciscoaguirre:
```suggestion
they are deposited to some beneficiary. It requires an instruction to place assets within it and
```

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
+another to withdraw them. The simplest example of this occurring is the `DepositAsset` instruction,
```
franciscoaguirre, line 84:
```suggestion
another to take them out of it. The simplest example of this is the `DepositAsset` instruction,
```

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
+    DepositAsset {
+        assets: MultiAssetFilter,
+        beneficiary: MultiLocation,
+    },
+    /* snip */
+}
 ```
+
+This instruction specifies how many assets are going to be withdrawn (useful for fee calculation),
+which assets to filter from the Holding register, and the beneficiary (the recipient of) the assets.
```
franciscoaguirre:
```suggestion
This instruction specifies which assets (kind and amount), already present in the Holding register, are going to be taken from it and deposited to a beneficiary (recipient).
```

docs/learn/learn-xcvm.md
```diff
@@ -39,12 +118,45 @@ enum Instruction {
...
-assets, and in the case of a fungible asset, it represents some defined amount of the asset.
+- A `MultiAsset` is a general identifier for an asset. It may represent both fungible and
+  non-fungible assets, and in the case of a fungible asset, it represents some defined amount of the
+  asset.
+
+- A `MultiLocation` is a relative identifier, meaning that it can only be used to define the the
+  relative path between two locations, and cannot generally be used to refer to a location
+  universally.
+
+`TransferAsset` is one of the many instructions that can be executed within an XCM. For more
```
franciscoaguirre:
```suggestion
`TransferAsset` is one of the many instructions that can be executed within an XCVM. For more
```
or
```suggestion
`TransferAsset` is one of the many instructions that can be contained within an XCM. For more
```

docs/learn/learn-xcvm.md
```diff
@@ -39,12 +118,45 @@ enum Instruction {
...
+information, please read [XCM Instructions in the wiki](./learn-xcm-instructions.md).
+
+## Locations in XCM
+
+XCM's abstract nature allows for locations to be relatively abstract notions that point to where but
+also _to who_ a particular action may affect. The `MulitLocation` type is what XCM uses to define
+these locations. A `MultiLocation` is a relative identifier that defines a **relative** path into
+some state-bearing consensus system.
```
franciscoaguirre:
```suggestion
XCM's general nature requires representing a wide array of "locations" (think relay chains, para chains, solo chains, smart contracts, etc). It also requires specifying accounts or "bodies" inside these systems.
The `MulitLocation` type is what XCM uses to define these locations.
A `MultiLocation` is an identifier that defines a **relative** path into some state-bearing consensus system.
```

docs/learn/learn-xcvm.md
```diff
@@ -39,12 +118,45 @@ enum Instruction {
...
+- A series of paths, called `Junctions`, which define an interior portion of state to descend into
+  it (sometimes called a "sub-consensus" system, such as a smart contract or pallet). An interior
+  location may also be used to refer to a Junction, used in the context of "a parachain is an
+  interior location of the relay chain", or how a UTXO is interior to Bitcoin's consensus.
+- The number of parent junctions at the beginning of a `MultiLocation`'s formation - in other words,
+  the number of parent consensus systems above it.
+
+There are a number of various `Junction` variants that may be used to describe a particular
+location - whether it's a 32 byte account, a Substrate pallet, or a pluralistic body.
+
```
franciscoaguirre, line 159: I would also add an example of a multilocation, e. g:

If you were on Parachain X on Polkadot and wanted to reference an account on Statemint (para chain id 1000) you would write the following `MultiLocation`:

```
../Parachain(1000)/AccountId32(<some_id>)
```
which in rust would be:

```rust
MultiLocation {
  parents: 1,
  interior: X2(Parachain(1000), <some_id>.into())
}
```

As illustrated by the `../` or `parents: 1` in Rust, this `MultiLocation` will go "up" by one consensus system and then go "down" the two junctions `Parachain` and `AccountId32`.
Keep in mind this location is relative, if this `MultiLocation` was interpreted by a Polkadot para chain, it would go "up" to the relay chain and then "down" to statemint (para chain id 1000) and to an account with some id.
If it was interpreted in Kusama, the underlying location it identifies would be different.
Some consensus systems might not even be able to interpret this location (Ethereum for example).

franciscoaguirre (commented):
I added a last batch of comments regarding writing and an example in the end.
Looking good :sunglasses:. Can't wait to go to the polkadot wiki and see all this.

docs/learn/learn-xcvm.md
CrackTheCode016: This is great, I'll encapsulate this 🎉

CrackTheCode016:
@franciscoaguirre thank you **very** much for the superb feedback and review!

@DrW3RK This PR is ready for a last review and merge 🎉

franciscoaguirre (approved):

docs/learn/learn-xcvm.md
```diff
@@ -39,12 +118,77 @@ enum Instruction {
...
+  asset.
+
+- A `MultiLocation` is a relative identifier, meaning that it can only be used to define the the
+  relative path between two locations, and cannot generally be used to refer to a location
+  universally.
+
+`TransferAsset` is one of the many instructions that can be contained within an XCM. For more
+information, please read [XCM Instructions in the wiki](./learn-xcm-instructions.md).
+
+## Locations in XCM
```
gavofyork, line 132: It might be prudent to mention the Universal Location in its own subsection here.

CrackTheCode016: @gavofyork Added - I'm not entirely sure it's accurate, so let me know if anything needs to be changed! If it's all well, I'll add a graphic to demonstrate a scenario.

As a sidenote, this concept will really come in handy when explaining how a bridge between global networks could actually look like 👍

docs/learn/learn-xcvm.md
```diff
@@ -8,28 +8,107 @@ slug: ../learn-xcvm
...
+register-based state machine. The state is tracked in domain-specific registers that hold
+information that is used and mutated along the execution of a particular message. Most of the XCM
+format comprises these registers and the instructions used to compose XCVM programs.
+
+The XCVM is an ultra-high-level non-Turing-complete computer whose instructions are designed to be
+roughly at the same level as transactions in terms of definition. Messages are one or more XCM
+instructions executed in order by the XCVM. An XCM is executed until it either runs to the end or
+hits an error, at which point it finishes up and halts.
+
+The primary implementation of the XCVM is the
```
gavofyork: Let's not call it "primary". That would imply we never expect any other implementation to become more significant, which is an overtly political assertion that does not belong in a specification.

"The first implementation" would work just as well.

docs/learn/learn-xcvm.md
```diff
@@ -187,6 +187,19 @@ Keep in mind that this location is specific to this interaction. The identities
 if this location was defined on another consensus system, such as Kusama. On other consensus
 systems, such as Ethereum, it won't be able to interpret it.
 
+### UniversalLocation in XCM
+
+While a `MultiLocation` may handle relative paths in _local_ consensus, a `UniversalLocation` refers
```
gavofyork: No - a `MultiLocation` is just a relative location. It has no concept of local consensus and can trivially refer to systems outside of local consensus.

docs/learn/learn-xcvm.md
```diff
@@ -187,6 +187,19 @@ Keep in mind that this location is specific to this interaction. The identities
...
+to any global consensus system. A global consensus system is an entity like another relay chain,
+Bitcoin, or Ethereum. A body of global governance capable of externalizing its consensus can be
```
gavofyork: No idea what "a body of global governance" means here.

docs/learn/learn-xcvm.md
```diff
@@ -187,6 +187,19 @@ Keep in mind that this location is specific to this interaction. The identities
 if this location was defined on another consensus system, such as Kusama. On other consensus
 systems, such as Ethereum, it won't be able to interpret it.
 
+### UniversalLocation in XCM
+
+While a `MultiLocation` may handle relative paths in _local_ consensus, a `UniversalLocation` refers
+to any global consensus system. A global consensus system is an entity like another relay chain,
```
gavofyork: "...is an entity which provides its own top-level consensus through some non-derivative consensus algorithm able to exist without reference to any other singleton data system".

docs/learn/learn-xcvm.md
```diff
@@ -187,6 +187,19 @@ Keep in mind that this location is specific to this interaction. The identities
...
+Bitcoin, or Ethereum. A body of global governance capable of externalizing its consensus can be
+referred to as a `UniversalLocation`.
+
+The `GlobalConsensus` junction refers to a global consensus system, and takes a `NetworkId` that
+specifies a particular remote network. A `UniversalLocation` allows for overarching consensus
+systems to communicate using this junction. Sub-consensus systems (i.e., a parachain on Polkadot)
+may refer to other sub-consensus systems (a parachain on Kusama) using a combination of
```
gavofyork: ...other *remote* sub-consensus systems...

docs/learn/learn-xcvm.md
```diff
@@ -187,6 +187,19 @@ Keep in mind that this location is specific to this interaction. The identities
...
+`UniversialLocation` and `MultiLocation` declarations.
```
gavofyork: This is superfluous and misleading. Locations may refer to other locations through a `MultiLocation` value - i.e. a relative path between them.

All locations held in consensus are also referencable through an interior location (i.e. a relative location with no parents) from the Universal Location down.

CrackTheCode016: thank you, addressed all outstanding comments!

CrackTheCode016:
@gavofyork Addressed and solved, thank you for the review!

DrW3RK (approved):

docs/learn/learn-xcvm.md
gavofyork: This is still wrong. Why was the comment ignored??

docs/learn/learn-xcvm.md
```diff
@@ -39,12 +118,93 @@ enum Instruction {
...
+notions that point to _where_ but also _to who_ a particular action may affect. The `MulitLocation`
+type is what XCM uses to define these locations.
+
+A `MultiLocation` is a relative identifier that defines a **relative** path into some state-bearing
+consensus system.
+
+It is used to define the relative path between two locations, and cannot generally be used to refer
+to a location universally. It is very much akin to how a **relative** filesystem path works and is
+dependent on the context on which consensus system the message is being sent from, and what
+consensus system it is being sent to be executed.
```
gavofyork:
```suggestion
dependent on the which consensus system the location expression is being evaluated.
```

gavofyork:
Looks reasonable modulo my most recent remark.

PR #4632, opened 2023-03-28
Fast Unstake is live on Polkadot

DrW3RK:


filippoweb3 (approved):
lgtm, should we just remind the reader that fast unstaking applies only to nominators and not to pool members?

DrW3RK:
Yes, let's state it explicitly. What would be the best way? Adding a separate note in nomination pools doc or modify the existing note here? To keep things simple, we can do it in a different PR.

filippoweb3:
Agree, we can do it in another PR. Not sure of the pool page already states that fast unstake is not possible. If not, we can probably add it there and say in the "non-pool" pages that fast unstake is possible only for nominators and not pool members. What do you think?

DrW3RK:
sounds good. merging this

PR #4649, opened 2023-04-01
Fix for current auction status in auction tool

alfarok:
### Description
It was previously discovered in #4593 that if no current or pending auctions were available the auction tool would hang on "Loading Auctions...".  #4593 corrects the issue in some scenarios but introduces additional edge cases that will result in "There are currently no pending or ongoing auctions" being displayed when there are indeed pending or active auctions.  For example, the Kusama variant currently displays this message in production but there is currently an active and future auctions in the cache.

Moving the statement from the previous fix to out of the for-loop and setting the index to the last auction in the cache when no auctions are active or pending should correct failing edge cases.

**Before** fix (current state) w/ active and pending:

image

**After** fix w/ no active but pending:

image

**After** fix w/ no active and no pending:

image

**After** fix w/ active and pending:

image

### FYI
@DrW3RK @filippoweb3 @CrackTheCode016 Hey guys! 👋 I happen to notice the auction status was incorrect when checking the tool recently and had a few minutes to look into it.  This PR builds off Bader's fix to handle all the various states.  Hope all is well!

### DIFF
--- a/components/Auction-Schedule.jsx
+++ b/components/Auction-Schedule.jsx
@@ -124,7 +124,8 @@ function GetCurrentOrNextAuction(chain, auctions, currentBlock) {
 			return [index, status];
 		}
 	}
-	return [auctions.length - 1, status = `There are currently no pending or ongoing auctions.  Auction #${auctions.length} was the last active auction.`];
+	status = `There are currently no pending or ongoing auctions.  Auction #${auctions.length} was the last active auction.`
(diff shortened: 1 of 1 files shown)


CrackTheCode016:
@alfarok thank you so much for this fix!! really appreciate the time you took to do this, lgtm.

CrackTheCode016 (approved):

PR #4679, opened 2023-04-12
Add Kusama Treasury Proposal Template

DrW3RK:
The PR was originally submitted here https://github.com/w3f/kusama-guide-hosting/pull/1


filippoweb3 (approved):
lgtm, I removed the banners within the conditional rendering for Kusama as the automatic prettier is reformatting them in a way they are not correctly displayed on both versions of the Wiki. Something to keep in mind next time we have banners within conditional rendering. I tried to spend some time to find a workaround, but I did not find a solution so far.

Issue #4376, opened 2023-01-26
Wiki Reorganisation


filippoweb3:
See [this google doc](https://docs.google.com/document/d/1Fl8ReBIzcaQEOyF6s03d7MEcgk8aTVEt7xjmzAk_k5I/edit) for more information about this issue.

- [x] Reorg Polkadot Comparisons page
- [x] Reorg Polkadot Launch Phases page
- [x] Reorg Availability and Validity page
- [x] Reorg Nominator page
- [x] Wallet Page / General Section
- [x] New Parity Signer Page (Polkadot Valut)
- [ ] Unify graphics
- [ ] Reorg Video Tutorials Page
- [x] Advanced How-to Guides
- [x] Archive Section (Gov, etc.)
- [x] Staking Dashboard Page
- [x] ETH 2.0 Page
- [x] Reorg Cryptography Page (move Randomness Page content there) 
- [x] Unify Statemin* pages
- [ ] Cards for all category sections in the sidebar for better navigation.

PR #4688, opened 2023-04-12
Wiki Reorg: Transfers Category Sidebar

filippoweb3:

### DIFF
--- a/docs/learn/learn-extrinsics.md
+++ b/docs/learn/learn-extrinsics.md
@@ -1,6 +1,6 @@
 ---
 id: learn-extrinsics
-title: Extrinsics
+title: Extrinsics (a.k.a. Types of Transactions)
 sidebar_label: Extrinsics
 description: Learn about Transaction Types (i.e. Extrinsics)
 keywords: [transaction, DOT, extrinsics, KSM]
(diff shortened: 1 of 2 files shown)


DrW3RK (commented):
Just one nit - Not all extrinsics are transfers. It would be better to leave it where it is.

filippoweb3:
Yeah, I was struggling to pick a name for that section but I think is good to have it separated from accounts, also because other pages fit well in there. What do you suggest? Transactions? Something that the user can easily understand as extrinsic is something that is not understandable without reading the pages. IMO extrinsics page does not fit well in Accounts. We even mention extrinsics (aka types of transactions). My point is that technically speaking "Transactions" or "Transfer" is not 100% right but it drives the reader in the right place to learn about extrinsics.

filippoweb3:
@DrW3RK I moved the Extrinsic page to General as it contains important security information.

DrW3RK:
If you renamed it from "Transfers" to "Transactions", I think it fits very well. Let me move it in there

DrW3RK:
If you think that the security-related info on the extrinsics page is well suited to be on the scams page or dyor page, that content can be moved there, and then we can point to that section in this doc.

Having Extrinsics, Balance Transfers and Transaction fees under the "Transactions" folder is apt.

DrW3RK (approved):

PR #4690, opened 2023-04-12
Wiki Reorg: Components

filippoweb3:


DrW3RK:
> Duplicated content about forkless upgrades in WASM and Runtime Upgrades. Solution: since the content in WASM is minimal, add it to Runtime Page

I think moving the Wasm content to runtime upgrades doc makes more sense like mentioned in the Google doc. Here I see that everything moved to the Wasm page instead.

We need to have a dedicated page that is titled "Runtime Upgrades" given its popularity and relevance. We can place "Runtime Upgrades" doc under "Architecture".

filippoweb3:
@DrW3RK I recreated the runtime upgrades page. I think there are a couple of issues regarding the type of contentment in Components and Architecture that can be addressed in a separate PR.

DrW3RK (approved):

PR #4805, opened 2023-05-16
Update build-protocol-info.md

maayank:
This fixes an incorrect amount for the existential deposit of Kusama in the wiki.

For reference, taking a look at
https://github.com/paritytech/polkadot/blob/master/runtime/kusama/constants/src/lib.rs
```
    /// The existential deposit.
    pub const EXISTENTIAL_DEPOSIT: Balance = 1 * CENTS;

    pub const UNITS: Balance = 1_000_000_000_000;
    pub const QUID: Balance = UNITS / 30;
    pub const CENTS: Balance = QUID / 100;
```
and at
https://github.com/paritytech/polkadot/blob/master/node/service/chain-specs/kusama.json
```
    "tokenDecimals": 12,
```
We can see that the minimum reserve is 0.0003333333333333333 KSM:
EXISTENTIAL_DEPOSIT = 1 * CENTS = QUID / 100 = UNITS / 30 / 100 = 1_000_000_000_000 / 30 / 100
And that ^ divided by 1e12 yields the result.

### DIFF
--- a/docs/build/build-protocol-info.md
+++ b/docs/build/build-protocol-info.md
@@ -8,6 +8,8 @@ keywords: [build, protocol, extrinsics, events, transaction]
 slug: ../build-protocol-info
 ---
 
+import RPC from "./../../components/RPC-Connection";
+
 This page serves as a high-level introduction to the Polkadot protocol with terminology that may be
 specific to Polkadot, notable differences to other chains that you may have worked with, and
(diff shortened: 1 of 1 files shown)


DrW3RK (approved):
Thank you! I made an update to your PR to fetch the data on-chain. This way, we never have to manually change it when the on-chain values change.

PR #4814, opened 2023-05-19
Edits Signer Pages

filippoweb3:
- add video tutorial
- minimum hardware requirement
- more security notes

### DIFF
--- a/docs/general/polkadot-vault.md
+++ b/docs/general/polkadot-vault.md
@@ -20,6 +20,19 @@ and power users. See [Ledger](./ledger.md) devices for other cold storage soluti
 
 :::
 
+:::warning Use a smartphone that is no more than 2-3 years old
+
+Note that although the Vault app is available for old smartphones, different versions will be
+installed according to the phone's hardware. Old phones like iPhone 6 will install Parity Signer
+(the old brand name of the Vault app) with limited capabilities. There will be no log, no warning if
+the phone was connected while not using the app, no metadata updates possible, and no option to add
+new networks. This would not be as secure as the latest version of the app. We recommend using a
+smartphone that is no more than 2-3 years old.
+
+:::
+
+<iframe width="560" height="315" src="https://youtube.com/embed/IG_RGLsb2g0" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe><br/><br/>
+
 [Polkadot Vault](https://www.parity.io/technologies/signer) (formerly Parity Signer) is a cold
 storage solution that allows you to use a phone in airplane mode as an air-gapped wallet. The Vault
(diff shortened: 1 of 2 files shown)


docs/general/polkadot-vault.md
```diff
@@ -20,6 +20,19 @@ and power users. See [Ledger](./ledger.md) devices for other cold storage soluti
...
+:::warning Use a smartphone that is no more than 2-3 years old
+
+Note that although the Vault app is available for old smartphones, different versions will be
+installed according to the phone's hardware. Old phones like iPhone 6 will install Parity Signer
+(the old brand name of the Vault app) with limited capabilities. There will be no log, no warning if
+the phone was connected while not using the app, no metadata updates possible, and no option to add
+new networks. This would not be as secure as the latest version of the app. We recommend using a
+smartphone that is no more than 2-3 years old.
+
+:::
```
profradha: The point to be conveyed here is that the smartphone should be compatible with the latest version on the app. There could be smartphones that are older that do support the latest version or some latest ones that are not compatible. 

```suggestion

:::warning Smartphone compatibility with Polkadot Vault

Note that although the Vault app is available for old smartphones, different versions will be
installed according to the phone's hardware. For example, smartphones like iPhone 6 will install Parity Signer
(the old brand name of the Vault app) with limited capabilities. There will be no log or warning if
the phone was connected to the internet while not using the app. Also, no metadata updates are possible, and no option to add
new networks. This would not be as secure as the latest version of the app. We would recommend that you use smartphones compatible with the latest Polkadot Vault app.

:::
```

docs/general/polkadot-vault.md
```diff
@@ -122,6 +135,14 @@ To securely update the Polkadot Vault app follow the instructions
 
 ## Security Notes
 
+:::info Remove SIM cards and Forget Networks
+
+To avoid unintended connection of your phone to the Internet, remove SIM cards, reset eSIM settings,
+and forget any WiFi network. In this way, the only possibility to have an intended internet
+connection is either with a SIM card or a WiFi router.
```
profradha:
```suggestion
connection is either through the cellular network or through WiFi connection.
```

docs/general/polkadot-vault.md
```diff
@@ -122,6 +135,14 @@ To securely update the Polkadot Vault app follow the instructions
 
 ## Security Notes
 
+:::info Remove SIM cards and Forget Networks
+
+To avoid unintended connection of your phone to the Internet, remove SIM cards, reset eSIM settings,
+and forget any WiFi network. In this way, the only possibility to have an intended internet
```
profradha:
```suggestion
and forget any added WiFi networks. In this way, the only possibility of having an unintended internet
```

DrW3RK (changes requested):
Suggested a few changes. Feel free to rephrase further.

DrW3RK (approved):

PR #4815, opened 2023-05-19
Update build-open-source.md to update status of FIAT on-off ramp 

dsm-w3f:


DrW3RK (approved):
@dsm-w3f These updates should happen automatically every weekend and do not have to update manually! Let us know if you see any discrepancies after the job runs.

PR #4844, opened 2023-05-28
Update learn-comparisons-cosmos.md

muddlebee:
added more clarification to summarize cosmos blockchain in the `conclusion` section

> meaning that inter-chain communication is still bounded by the trust that the receiving chain has in the sending chain.

I wasn't able to understand the line above, so researched a bit and added more context for better clarification

### DIFF
--- a/docs/learn/learn-comparisons-cosmos.md
+++ b/docs/learn/learn-comparisons-cosmos.md
@@ -203,4 +203,7 @@ logic and security processes across chains so that they can interact knowing tha
 interlocutors execute within the same security context.
 
 The Cosmos network uses a bridge-hub model to connect chains with independent security guarantees,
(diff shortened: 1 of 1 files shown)


docs/learn/learn-comparisons-cosmos.md
```diff
@@ -203,5 +203,4 @@ logic and security processes across chains so that they can interact knowing tha
 interlocutors execute within the same security context.
 
 The Cosmos network uses a bridge-hub model to connect chains with independent security guarantees,
-meaning that inter-chain communication is still bounded by the trust that the receiving chain has in
-the sending chain.
+meaning that when data is sent from one chain to another (inter-chain communication), the receiving chain must trust the sending chain. Each blockchain in the Cosmos network has its own security mechanisms, they're independently secured and do not rely on the security of other blockchains or the hub. This allows each chain to function independently while still being part of the Cosmos ecosystem.
```
CrackTheCode016: "Each blockchain in the Cosmos network has its own independent security mechanisms. They're independently secured and do not rely on the security of other blockchains or the hub."

CrackTheCode016 (changes requested):
It looks good to me besides the tiny nit I noted. Thank you for this!

filippoweb3 (approved):
thanks for the PR @muddlebee  and for the review @CrackTheCode016!

PR #4860, opened 2023-05-31
Validator Guide: add instructions for installing multiple binaries

mrcnski:
# PULL REQUEST

## Overview

After extensive discussion (see [here](https://github.com/paritytech/polkadot/pull/7147) and [here](https://forum.polkadot.network/t/ux-of-distributing-multiple-binaries-take-2/2854)), it's been determined that we will be distributing multiple binaries along with `polkadot`. I'm updating the Validator Guide to reflect this. 

In the future, to make it easier for validators to install multiple binaries, this guide may be the basis for some script if that is desired.

## TODO

- [x] 1. Raise polkadot PR splitting out the binaries and adding `--workers-path` flag. 
  - https://github.com/paritytech/polkadot/pull/7337
- [x] 2. Update polkadot README with new build instructions
- [x] 3. Update release process to produce new binaries

## Related

https://github.com/paritytech/polkadot/pull/7147
https://github.com/paritytech/polkadot/issues/7284


Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/ux-of-distributing-multiple-binaries-take-2/2854/39

PierreBesson:
@mrcnski if there is a need to set the `--worker-path` flag pointing to the path of the worker binaries when starting the node as a validator this needs to be clearly explicited in this doc. For now this is missing from this PR.

mrcnski:
Thanks @PierreBesson! We tried to make it so that existing workflows work in a backwards-compatible way without further changes. The `--workers-path` option was added just-in-case, but shouldn't be necessary for any of the installation methods in the guide. I'm not sure if it's worth mentioning in the guide for that reason, or if it would just add noise. Feedback always welcome though. I've pushed an update.

docs/maintain/maintain-guides-how-to-validate-polkadot.md
```diff
@@ -181,17 +179,19 @@ If it is not enabled, please see the
 [official docs ("Kernel support")](https://docs.kernel.org/userspace-api/landlock.html#kernel-support)
 if you would like to build Linux with landlock enabled.
 
-### Installing the `polkadot` binary
+### Installing the Polkadot binaries
```
wpank, line 184:
```suggestion
:::info Multiple Validator Binaries

In addition to the `polkadot` binary, recent changes have separated out functionality into two additional needed binaries, `polkadot-prepare-worker`, and `polkadot-execute-worker`.  All three binaries are needed to properly run a validator node. More context on these changes can be found [here](https://github.com/paritytech/polkadot/pull/7337)

:::
```

wpank (commented):
Overall would say this looks good - one thing would be to maybe add a small bit of context that polkadot now requires 3 binaries

docs/maintain/maintain-guides-how-to-validate-polkadot.md
mrcnski: Good idea, done!

wpank (approved):

PR #4866, opened 2023-06-01
Wiki Reorg: Add Future Implementation Section

filippoweb3:


filippoweb3:
@CrackTheCode016 I added a new category as you suggested.

I think it would be best if: 1) we create a new .md page just for async backing, 2) we put the page into this category, 3) we link the page to the learn-parachains-protocol.md as that page is already a beast! Let me know what you think. I think after moving the new content, we can close PR https://github.com/w3f/polkadot-wiki/pull/4782- @DrW3RK do you agree?

CrackTheCode016:
@filippoweb3 I agree! I think that would be good for this PR (async backing + the new section).  Will add more content after the initial merge

CrackTheCode016 (approved):

PR #4872, opened 2023-06-02
Validator Guide: add note about landlock

mrcnski:
# PULL REQUEST

## Overview

Adds a note about making sure landlock is enabled. See https://github.com/paritytech/polkadot/pull/7303.

### DIFF
--- a/docs/maintain/maintain-guides-how-to-validate-polkadot.md
+++ b/docs/maintain/maintain-guides-how-to-validate-polkadot.md
@@ -162,6 +162,25 @@ making it on chain.
 
 :::
 
+### Make Sure Landlock is Enabled
+
+[Landlock](https://docs.kernel.org/userspace-api/landlock.html) is a Linux security feature used in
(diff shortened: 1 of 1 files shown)


DrW3RK:
Thank you! I will have someone with the subject matter expertise to get this reviewed and merged. Does this PR https://github.com/paritytech/polkadot/pull/7303 needs to be merged before this one?

mrcnski:
Thanks @DrW3RK! 

> Does this PR https://github.com/paritytech/polkadot/pull/7303 needs to be merged before this one?

Good question. I think no, the order does not really matter.

We can add this note ahead of time, and let validators know about the new requirement, in anticipation of that PR. On the other hand, we can also safely merge that PR first -- if a validator does not have landlock enabled nothing will happen besides a warning being logged. 

I don't believe either order is preferable really, so I would just start with whichever PR is ready-to-merge first.

DrW3RK (approved):

PR #4947, opened 2023-06-23
Fix redirects - Kusama Guide and Polkadot Wiki

DrW3RK:


filippoweb3 (approved):
thanks a lot for this!

PR #4959, opened 2023-06-26
Update community.md and add telegram link for substrate devs

muddlebee:
add telegram link for substrate devs

**Note** : the telegram channel is bridged to the element matrix channels, so same content is synced in both channels.


muddlebee:
Note : the telegram channel is bridged to the element matrix channels, so same content is synced in both channels.

DrW3RK (approved):
Thank you!

PR #4967, opened 2023-06-27
Add Mermaid.js Theme

CrackTheCode016:
As part of an effort to unify graphics, mermaid can be used.  This PR adds a basic theme for mermaid js diagrams.  

## todo
- [ ] test with multiple types of graphics
- [ ] Add Kusama theme


kpx001:
yes

CrackTheCode016:
Closing this in favor of themed graphics which we have utilized.  It is likely something like this would be more involved, and for now, it is not a priority.

PR #5005, opened 2023-07-05
Update with faq & resources

stakeworld:
I've added a faq and some resources/web links and moved some things around. My main idea is to give newer applicants some more relevant info (the faq and resources) so we also can refer to this for common questions on the matrix 1kv pages.

Probably a good idea to see if @wpank has remarks or corrections.

### DIFF
--- a/docs/general/thousand-validators.md
+++ b/docs/general/thousand-validators.md
@@ -34,10 +34,92 @@ descending order.
 The backend nominates as many validators as it reasonably can in such a manner to allow each nominee
 an opportunity to be elected into the active set.
 
-### Weights
+## Frequently asked questions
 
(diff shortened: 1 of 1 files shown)


filippoweb3:
@stakeworld, thanks for this contribution! Let me make some minor changes :)

filippoweb3 (approved):
I made some edits without removing the added content, but keeping the wiki unopinionated and removing subjective comments. Let me know what you think.

@DrW3RK IMO this can be merged but have a look if there is anything to add / change.

stakeworld:
Fine by me. Fair point keeping the wiki more about facts.

DrW3RK (approved):

PR #5024, opened 2023-07-07
Polkadot's Vision - Ubiquitous Computer, Core Time, Accords and Resilience

### LINKED ISSUE #4992: [Feature Request] : Update Polkadot 2.0 vision
https://wiki.polkadot.network/docs/learn-launch#polkadot-20

Context of Polkadot 2.0 : [Gavin Wood Outlines the Future of Polkadot | Polkadot Decoded 2023](https://youtu.be/GIB1WeVuJD0)

filippoweb3:


laboon:
We should also mention accords as part of this PR @filippoweb3 

https://polkadot.network/blog/polkadot-decoded-presents-powerful-vision-across-the-ecosystem-in-two-action-packed-days

docs/general/polkadot-uc.md
```diff
@@ -0,0 +1,337 @@
+---
+id: polkadot-uc
+title: The Polkadot Ubiquitous Computer
```
shawntabrizi: I personally feel this is a "gav-ism" and instead naming should be left to marketing and such before this goes live.

docs/general/polkadot-uc.md
```diff
@@ -0,0 +1,337 @@
...
+  [parathreads](../learn/learn-parathreads.md)). At the time of writing (mid 2023), on Polkadot,
+  there are around 50 cores independently operating in parallel. Prediction models show that cores
+  can increase to 500-1000 in the following few years.
+
+As a computational resource, Polkadot can be described as a traditional CPU, and at the time of
+writing it has the following benchmarking features:
+
+- Bandwidth: ~1MB/s
+- Compute: Geekbench 5 SC score ~380
+- Latency: 6s (time between two consecutive tasks)
```
shawntabrizi: i think you should provide sources and reproducible benchmarks for this before putting such things in docs

docs/general/polkadot-uc.md
```diff
@@ -0,0 +1,337 @@
...
+
+In general, bulk coretime NFTs can be:
+
+- sold to one application (similar to the lease auctions in Polkadot 1.0, with the difference that
+  coretime in Polkadot 1.0 is not sold, but applications need to lock their DOT tokens)
+- sold to several applications taking turns on a core
+- placed on the instantaneous market
+- exposed and sold by a broker system parachains via [XCM](./learn-xcm-index) NFT interface to
+  Polkadot-powered applications that want to allow for these NFTs to be exchanged and traded on
+  their market.
```
shawntabrizi: this is a lot of specific implementation details rather than simply a vision / direction.

docs/general/polkadot-uc.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Core Usage in Polkadot 1.0
+
+In Polkadot 1.0, applications produced blocks at a fixed rate of 12 seconds, whether needed or not.
+This led to inefficient energy allocation and economic incentives for producing full blocks under
+heavy traffic and empty blocks under light traffic.
+
+The figure below shows the core usage for Polkadot 1.0, where the horizontal axis is time, and each
+row represents a core. Colors show different parachains, each using one core (i.e., one parachain,
+one core formula).
```
shawntabrizi: i think you should probably start the whole narrative somewhere here.

docs/assets/core-usage-agile-rangeSplit.png
shawntabrizi, line 1: "cores" as an axis is not clear.

there is no reason you cannot label each row with `Core #0`, `Core #1`, and so on.

Also lines or boxes to separate the rows of cores would add clarity. Right now, things look as if they are colors on a two dimensional grid, rather than lines of processes on specific cores.

Also some images have 5 cores, others have 4. This should probably be consistent.

docs/assets/agile-polkadot-accords.png
shawntabrizi, line 1: from my understanding of accords, I think this is not really a great visualization of what is going on.

the more clear visualization is that two applications are communicating and interoperating with one another via some shared logic and state on the relay chain. Because the relay chain exists at all, and provides shared security to the parachains, such interactions are trustless and secure, whereas the same cannot be said if logic or data was stored on the individual parachains.

you might think of this like two untrusted people doing buisness in a police station / court house, knowing that everything is happening under the watchful eye of a mutually trusted entity.

docs/assets/agile-polkadot.png
shawntabrizi, line 1: > Polkadot should be an integrated system (big circle), a single computer on which many applications run (open circles). There are boundaries between business logic components of Polkadot (dashed lines). Those boundaries can enclose parachains or core contracts. Applications should be able to communicate with each other within and across those boundaries. Last but not least, like a laptop, knowing how Polkadot works must be optional for the users who just need it to use applications.

This really makes no sense to me. I challenge you to actually make tangible names which would represent real world examples of of this. For example:

- first name for me multiple applications which would fit this model
- then name for me the grouping of some of those applications by a "logic component"
- and then name or describe to me how multiple "logic components" would interact

shawntabrizi: "logic component" should probably be "logical component" as logic and component are both nouns

shawntabrizi:
In general, this looks like a written summary of Gav's presentation with little distilling of the key details which are actually important into describing the vision and direction of Polkadot 2.0. I question whether this content should be in the Wiki as is.

Instead, a deeper understanding of the shift from:
- the "parachain focused" paradigm
    - where each application owned an execution core at all times (which are these fixed parachain slots), which allowed a simple and secure, sharded execution environment
- to the "agile core" paradigm
    - where we remove the assumption that each application owns a core, and instead that all cores are a resource to be consumed and used as needed by all applications in our ecosystem. harder to implement, which is why it is next on the roadmap, but builds entirely on top of existing working primitives.

should be highlighted in this document. Specific details like "4 week sales via NFTs" and stuff like that is literally irrelevant for this vision, and it enshrines implementation detail which no code has been written for into docs. This leads to public confusion when these things inevitably change. Instead it should be noted that generally speaking, core time can be "tokenized", sold, and traded.

Then talking about Accords makes sense under the framing of Polkadot relay chain taking on new responsibilities to improve cross chain communication and the security guarantees of those messages. Being more tangible about how the relay chain and Polkadot protocol is specifically well suited for this, and such things could not exist in a non-shared-security environment.

It might make sense to also emphasize at some point the way that Polkadot itself is scaling by moving on-chian logic into its own system parachains, allowing it to have more bandwidth for parachains protocol and accords.

filippoweb3:
thanks @shawntabrizi for the review! I will work on this in a couple of weeks and get back to you.

Yes, the page is a summary of Gavin's Talk at Decoded (see banner at the top). I agree there is probably way too much detail that hides the most important points you mentioned above. I think some of the details can be left somewhere (maybe at the end of the page on a dedicated section) to make the reader aware of the possibilities to use coretime although nothing has been implemented yet (and when it will, we will create a detailed page). Since these implementations will take some time it would be good to have some detailed options as described by Gavin.

docs/general/polkadot-uc.md
filippoweb3: I thought these data were already coming from some tests (as they were presented at the Decoded)

docs/general/polkadot-uc.md
filippoweb3: Would Polkadot 2.0 be a better option? From the Marcomm Element chat, I understood Polkadot UC was preferred.

docs/general/polkadot-uc.md
filippoweb3: Agree, I will propose a solution for this

docs/assets/agile-polkadot-accords.png
filippoweb3: right! I will propose a solution for the new representation :)

docs/assets/agile-polkadot.png
filippoweb3: I think it would be nice to confront Polkadot with big social experiments like EU, US, etc. or maybe airports? It was attempted at Decoded with the EU, but I am not a lawyer, and it would be great to have some feedback for international rights experts.

- logical components can be European countries with their laws
- applications can be different businesses that run following the law of each country
- businesses can have offices in different countries and thus operate under different law contexts but still serve the same function, although here I really do not know how it really works

docs/general/polkadot-uc.md
laboon: MarComms has not yet put together a specific term, last I checked. We should not use the term "Polkadot 2.0".  I've been using the rather vague phrase "potential changes to Polkadot".

shawntabrizi: i personally like polkadot 2.0, but yeah bill I think is taking the most sensible approach

docs/general/polkadot-uc.md
shawntabrizi: sounds like you should check this before you write down specific benchmarks into a polkadot official wiki

shawntabrizi: we should verify that and reference it

filippoweb3: deleted part until we verify those data

docs/general/polkadot-uc.md
filippoweb3: I deleted the part about bulk and instantaneous coretime rental

filippoweb3:
@shawntabrizi I implemented your feedback, removed all the parts that contain too much detail, changed the title of the page, removed polkadot representation of the "cookie" with the accords (I am working on it separately). let me know if I satisfied your feedback. Cheers!

docs/general/polkadot-v2.md
```diff
@@ -0,0 +1,234 @@
...
+transportation method and a secure route but not a framework for binding agreements.
+
+An **Accord** is an _opt-in_ treaty across many chains, where treaty logic cannot be changed or
+undermined by one or more of those chains, and Polkadot guarantees faithful execution of this logic.
+Accords will be specific to a particular function, and any chain that enters the accord will be held
+to it and will service that particular function. To lower the entry barrier, accords can be proposed
+permissionlessly, but because they are opt-in, the accord proposal will take effect until chains
+agree and sign up.
+
+Polkadot is the only ecosystem where accords can properly exist because it has an homogenous
```
KiChjang: You have explained XCM, XCMP and Accords above, but there's nothing here that ties all of these 3 concepts together. I was looking for a summary that says something to that effect, but currently there's nothing like that here.

docs/general/polkadot-v2.md
```diff
@@ -0,0 +1,234 @@
+---
+id: polkadot-v2
+title: Potential Changes to Polkadot
```
KiChjang: I don't think this is a "potential change", it is more like the next iteration of how we're viewing Polkadot; a perspective change, or a paradigm shift if you may.

docs/general/polkadot-v2.md
```diff
@@ -0,0 +1,234 @@
...
+sidebar_label: Potential Changes to Polkadot
+description: Polkadot as a Decentralized and Secure Ubiquitous Computer.
+keywords:
+  [polkadot vison, polkadot 2.0, Decentralization, ubiquitous computer, coretime, blockspace]
+slug: ../polkadot-v2
```
KiChjang: Just to reiterate what we've discussed in the Marcomms channel: we should avoid any public mentions of "Polkadot 2.0" and instead always specifically talk about the name of the feature that we want to explain.

docs/general/polkadot-v2.md
```diff
@@ -0,0 +1,234 @@
...
+The slot auction mechanism is not agile, creates high entry barriers, and is designed for
+long-running single applications (i.e., the original Polkadot vision proposed in the whitepaper).
+
+We depart from the classic lease auctions and propose an agile marketplace for coretime, where
+essentially **coretime becomes a commodity that can be tokenized, sold and traded**. This setup
+maximizes the agility of Polkadot and lets the market figure out the best solution needed for
+applications to be successful.
+
+Revenues from coretime sales can be burnt, used to fund the Treasury, split between coretime
+providers (i.e., validators), or used for a mix of those options. The topic is currently under
```
KiChjang: IIRC, the last method (splitting between coretime providers) is no longer considered. Check the latest discussion on RFC 10 for more details.

docs/general/polkadot-v2.md
```diff
@@ -0,0 +1,234 @@
...
+Polkadot is the only ecosystem where accords can properly exist because it has an homogenous
+security layer that also provides a specific state transition function for each of its logic
+components. This allows patterns of cooperation between multiple logic components (i.e.,
+trans-applications) that would not be possible to achieve over bridges.
+
+Accords will be implemented using [SPREE technology](../learn/learn-spree.md).
+
+## Agile Coretime Allocation
+
+In Polkadot 1.0 coretime is a fixed two-year period on one specific core. Here we remove this
```
KiChjang:
```suggestion
In Polkadot 1.0, coretime is a fixed two-year period on one specific core. Here we remove this
```

docs/general/polkadot-v2.md
```diff
@@ -0,0 +1,234 @@
...
+### Strided Coretime
+
+Ranges can be strided (i.e., applications can take turns on a core) to share costs or decrease block
+production rate, for example.
+
+![core-usage-agile-rangeStrided](../assets/core-usage-agile-rangeStrided.png)
+
+### Combined Coretime
+
+An application's task can be assigned to multiple cores simultaneously. Some applications can have a
```
KiChjang: We need to explain that "tasks" are essentially what we hitherto called "parachains"/"parathreads".

KiChjang:
Overall, I feel like the article is explaining the mechanics of Agile Coretime; while I agree it's important, it's currently deficient in the aspect of "why does this matter to me?"

One of the big things that we didn't talk about is how we're moving away from the slot auction model -- **this really needs to be highlighted**, as I've talked with teams in person before, and they're absolutely hyped about the ease of use and flexibility that the coretime marketplace is going to bring. We should also then explain how one can acquire coretime easily under this new conception of Polkadot.

docs/general/polkadot-v2.md
filippoweb3: changed to Polkadot Direction

docs/general/polkadot-v2.md
filippoweb3: added a sentence to tie all these three together

docs/general/polkadot-v2.md
filippoweb3: removed "task" as application is here a broad term for parachains and parathreads

filippoweb3:
Thanks for the feedback @KiChjang! Answering your questions:
- There is a section about switching from slot auctions to coretime sales on the page. Is that not enough? I can add a paragraph about bulk and instantaneous sales, but after @shawntabrizi feedback I removed those details completely and focused on the main changes. I probably removed too much!?
- Why does switching from auction to coretime sales matter to me? I am planning to link this page to the async backing page PR https://github.com/w3f/polkadot-wiki/pull/5176 and explain how in periods of high/low demand parachains can adjust their coretime / core allocation "profile" to get parablocks ancetstors in the unincluded segments being backed and included at different speeds. I see the coretime narrative as an elegant solution that is tightly bound to async backing and the fact that in periods of:
  - **low demand** parablock generation time (12s, 18s, more? to fill up blockspace) >> backing + inclusion (6s) -> unincluded segment shrinks -> share a core to increase backing time -> pay less
  - **high demand** parablock generation time (below 6s) << backing + inclusion (6s) -> unincluded segment grows -> more cores / coretime to lower backing times -> pay more

Let me know if the above makes sense and if there is anything else you suggest to add.

KiChjang:
> * Why does switching from auction to coretime sales matter to me?

Previously, securing a parachain slot required a massive amount of upfront capital to win the auction. With agile coretime, there's no need for such an auction anymore; teams can either purchase some Instantaneous Coretime, or reserve Bulk Coretime as required, and it greatly decreases the barrier-to-entry for software tinkerers and parachain teams.

In accordance to what Shawn mentioned, we don't need to mention specific details on the mechanism of coretime purchase; the core of the message is that **it lowers the barrier-to-entry for prospective builders**. This is literally the most important aspect when it comes to building on Polkadot. Hence from this perspective, async backing is interesting, but they're again intricate details and doesn't speak directly to what teams care about the most.

filippoweb3:
@KiChjang I added a summary that highlights the main points of the future direction, mentioned again the bulk and instantaneous coretime rental, and reorganized some text.

shawntabrizi:
I haven't had a chance to re-review the changes here. The feedback from @KiChjang generally looks good, but I didnt agree with:

> Just to reiterate what we've discussed in the Marcomms channel: we should avoid any public mentions of "Polkadot 2.0" and instead always specifically talk about the name of the feature that we want to explain.

Dunno the deeper conversations here, but Polkadot 2.0 is just such a simple brand and idea to capture all of these things, I think it would be a shame if we didnt use it. Even it we clarified that is it likely versioning of the internal software. Just as we have a Polakdot version 1.0 now, these features will be included in a future Polkadot 2.0 version (or sooner).

filippoweb3:
@shawntabrizi @KiChjang, after Bill's reply and reading the Marcomm element chat, I think it would be best to leave the current title until there is clear guidance from the marketing team.

I am still keen on adding a proper diagram to explain the core / application narrative. This can be added as a separate PR if you agree. Let me know if the current version of the page looks better and can be merged.

docs/general/polkadot-direction.md
```diff
@@ -0,0 +1,265 @@
...
+**Increasingly fewer tasks will be handled by the relay-chain** that will focus efforts only on
+primary tasks: securing the network and providing secure message-passing capability.
+[System parachains](../learn/learn-system-chains.md) will be used to take over secondary relay-chain
+tasks such as staking, governance, etc.
+
+### XCM and Accords
+
+[XCM](../learn/learn-xcm-transport.md) is a format, a language of intention abstract over
+functionality common within chains. It creates an expressive language of what you intend to do or
+want to happen. There are no guarantees that chains faithfully interpret XCM messages; thus, XCM is
```
shawntabrizi: i think this needs to be slightly re-worded.

as someone reading this, if XCM isnt good for trustless environments, and chains my not faithfully interpret XCM messages, why do we have it at all?

I think you need to preface the different actors and needs those actors might have, and the goal of providing different options for developers to use for their needs and designs

shawntabrizi (approved):
Overall, I still find the page a little funky.

The narrative voice is strange, since it uses first person `we` sometimes, and then switches to third person in later parts, and the presentation of the content / topics is not to me fluid.

Ultimately, I think these are just some of the challenges of converting a talk to a document.

That being said, I dont think there is much wrong or inaccurate with the content on the page, and thus funkiness aside, someone who reads it will probably be able to learn many of the new thoughts happening in the space.

I am not against merging this, which may help move forward to building an even better version of this content and data.

Issue #5052, opened 2023-07-22
Broken link in learn-xcm.md


LStan:
In https://github.com/w3f/polkadot-wiki/blob/master/docs/learn/learn-xcm.md there is a broken link to https://storage.googleapis.com/w3f-tech-ed-contents/XCMP.mp4

PR #5054, opened 2023-07-23
Clean up XCMP Portion + Video

CrackTheCode016:
This PR does two things: 

- Addresses and closes issue #5052 by removing it for now (until a suitable replacement can be found)
- Puts XCMP details in the transport layer page, where it should've been

### DIFF
--- a/docs/learn/learn-xcm-transport.md
+++ b/docs/learn/learn-xcm-transport.md
@@ -95,6 +95,9 @@ A tutorial on how to open an HRMP channel on a parachain can be found
 
 ### XCMP (Cross Consensus Message Passing) Design Summary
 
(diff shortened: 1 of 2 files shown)


DrW3RK (approved):

LStan:
Why are there "cross-chain" in some places? Should it be "cross-consensus" everywhere?

DrW3RK:
@LStan XCM (Cross Consensus Messaging) enables cross-chain communication https://polkadot.network/features/cross-chain-communication/ In the context of Polkadot parachains where all of them are under the same consensus model, "cross-chain" loosely refers to the functionality of XCM that lets people build cross-chain apps. But XCM is not just limited to parachain communication, but also communication with external blockchains with their own consensus models.

DrW3RK:
XCMP is a Cross Chain Messaging Protocol which is very different from XCM (Cross Consensus Messaging Format). XCMP is a protocol under development for direct communication between parachains. XCMP is like HTTP, whereas XCM is like HTML.

PR #5068, opened 2023-07-31
Reposition Cautionary note - Proxy Page

DrW3RK:
Moving the cautionary note to a section where it is more relevant (and the terminology used has been already introduced)


CrackTheCode016 (approved):
lgtm thanks!

PR #5072, opened 2023-08-02
Edits Voluntary Locking

### LINKED ISSUE #5049: Conviction Voting - Polkadot Params
There is some ambiguity with the archived page section showing up in the search results that misleads readers about current params https://wiki.polkadot.network/docs/learn-governance#voluntary-locking Either remove the section or update it to point to the correct params for OpenGov

image

### LINKED ISSUE #5059: Clarify voting replaces initial voting
https://wiki.polkadot.network/docs/maintain-guides-polkadot-opengov#voting-on-referenda

Voting a second time replaces your original vote, e.g. voting with 100 DOT, then a second extrinsic to vote with 10 DOT, means that you are voting with 10 DOT, not 110 DOT. This should be clarified either in this section of the wiki or in the linked support document ( https://support.polkadot.network/support/solutions/articles/65000184120-polkadot-opengov-how-to-vote ).

filippoweb3:


filippoweb3:
@michalisFr, it seems voluntary locking is less conservative in OpenGov than Gov1, and is the same on both Polkadot and Kusama. This PR fixes the wiki.

michalisFr:
@filippoweb3 This is done by mistake. The locking periods should be the same as they were in Gov1 (although I believe reducing the times on Kusama to multiples of 7 instead of 8 is intentional).
https://github.com/paritytech/polkadot/issues/7394

michalisFr (changes requested):
Keep the changes to Kusama (to multiples of 7) but also keep the original values of Polkadot and add a comment that the current values (that match those on Kusama) are by mistake that's going to be fixed to the proper values (see my comment).

filippoweb3:
@michalisFr I addressed your feedback, let me know if the PR is good to go

PR #5110, opened 2023-08-16
Can votes stack?

gilescope:
We want people to vote. We want them to vote often and on lots of things!

But people will be scared to vote if they think their votes will stack up to lock their dot for many years. So let's allay their fears and they will then vote more.


filippoweb3 (approved):
thanks for the contribution! I reworded it a bit without changing the meaning of the paragraph. lgtm!

PR #5122, opened 2023-08-22
Remove parachain limit ambiguity

### LINKED ISSUE #5121: Remove 100 limit
https://wiki.polkadot.network/docs/learn-parachains#parachain-slot-acquisition

I've no idea who pulled this 100 parachains limit out of the blue, but there was never any estimate of 100 parachains being any sort of limit.  It obviously depends upon a slew of factors, like hardware specs, ISP specs, blocktimes, all the optimizations upon which the parachains team works, excess CPU capacity held in reserve, and now core multiplexing.

100 might be the limit if validator specs required only one physical CPU core --- 1 second * 35 checkers * 100 / 4 checks is kinda close to 1000 validators

It should be removed since it creates confusion like https://substrate.stackexchange.com/questions/9652/how-a-parachain-will-connect-to-polkadot-relaychain-when-the-hundred-parachain-l

DrW3RK:
Need to check every reference to the 100 limit on the number of parachains and remove it from the Wiki.

### DIFF
--- a/docs/learn/learn-parachains.md
+++ b/docs/learn/learn-parachains.md
@@ -182,9 +182,7 @@ maintain the stronger property of single block composability.
 
 ## Parachain Slot Acquisition
 
-{{ polkadot: Polkadot :polkadot }}{{ kusama: Kusama :kusama }} supports a limited number of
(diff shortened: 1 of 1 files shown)


filippoweb3 (approved):
lgtm

PR #5139, opened 2023-08-29
Update build-open-source.md

Noc2:
Replace Chainbridge with Sygma, which is also developed by Chainsafe and the official successor


CrackTheCode016 (approved):
ty!

PR #5163, opened 2023-09-07
Add info about Polkadot's Native Account Abstraction

### LINKED ISSUE #5161: Add page on Polkadot-native account abstraction
Account abstraction already exists on Substrate/Polkadot and various features (batching, proxies, native multisigs, payments with non-native tokens, etc.) are already in use in the ecosystem. We should have a page explaining this, probably as a sub-page of https://wiki.polkadot.network/docs/learn-accounts-index and based on Bader's article here: https://hackernoon.com/abstracting-away-account-abstraction-on-polkadot

filippoweb3:


docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,64 @@
...
+sidebar_label: Account Abstraction
+description: Polkadot's Native Account Abstraction.
+keywords: [account, polkadot account, account abstraction, proxy, multisig, batch]
+slug: ../learn-account-abstraction
+---
+
+## Your Keys, Your Responsibility
+
+Account abstraction addresses the challenges associated with managing cryptographic keys that on
+blockchains represent accounts and, more generally, your digital identity. In Web3, you digitally
```
profradha:
```suggestion
Account abstraction addresses the challenges of managing cryptographic keys representing accounts on
blockchains and, more generally, your digital identity. In Web3, you digitally
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,64 @@
...
+sign any transaction or, more generally, any message using your private key. Data are recorded on a
+public ledger (blockchain) whose multiple copies of it are stored in computers participating in a
+P2P network. Assuming the network is sufficiently decentralized (meaning that those computers do not
+belong to a few operators), those computers do not own your data, you do since you can verify you
+signed them with your key.
```
profradha: > Assuming the network is sufficiently decentralized (meaning that those computers do not
> belong to a few operators), those computers do not own your data, you do since you can verify you
> signed them with your key.

I think this needs to be refined further. Verifying that a message has been signed with your private key does not necessarily grant you ownership of the message's contents. Yes, you can prove ownership of digital content like NFTs because the creator of the NFT posts it to the chain, and the ownership is transferred to you through a cryptographically secure transaction. Whether decentralized or centralized, no one can challenge that a message signed by your private key was signed by someone else, so decentralization has nothing to do with ownership in this context. 

Although, it is possible that the network does not make the data you own "available." That is where decentralization helps (like IPFS, which is not a blockchain protocol), but that has little to do with account abstraction.

We can remove those lines.

```suggestion
sign any transaction or, more generally, any message using your private key. Data is recorded on a
public ledger (blockchain) whose multiple copies of it are stored in computers participating in a
P2P network.
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,64 @@
...
+
+While they grant users control and ownership, losing keys results in a loss of digital assets and a
+loss and fragmentation of your digital identity since you will need to create a new account with a
+new key after losing a key. This poses a hurdle for both users and developers regarding security and
+adoption.
```
profradha:
```suggestion
While the account keys grant users control and ownership, losing them results in losing access to digital assets and fragmentation of your digital identity since you will need to create a new account with a
new set of keys. This poses a hurdle for both users and developers regarding security and
adoption.
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Definition of Account Abstraction
+
+Account abstraction introduces a layer of on-chain logic that controls an account, typically in the
+form of a smart contract. This abstracts some key management complexities, reducing the likelihood
+of compromise. It allows for dynamic changes in an account’s operation without altering the core
+protocol.
```
profradha: Some questions when reading this paragraph -

How does it reduce the likelihood of a compromise? What are the dynamic changes? What core protocols need to be altered if Account abstraction is not present?

DrW3RK (commented):
Some preliminary feedback.

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,61 @@
...
+sidebar_label: Account Abstraction
+description: Polkadot's Native Account Abstraction.
+keywords: [account, polkadot account, account abstraction, proxy, multisig, batch]
+slug: ../learn-account-abstraction
+---
+
+## Your Keys, Your Responsibility
+
+Account abstraction addresses the challenges of managing cryptographic keys representing accounts on
+blockchains and, more generally, your digital identity. In Web3, you digitally
```
CrackTheCode016: Suggested rephrase, or something along these lines:

```suggestion
blockchains.  Accounts on blockchains represent entites  that are participating, from an individual's identity, to a company.
```

docs/learn/learn-account-abstraction.md
CrackTheCode016: This is partially answered in the next paragraph, but it reduces the importance of having to manage the private key as the _sole_ method of access for a user's on-chain representation.  Of course, in our context, this depends on things like the social recovery pallet.

Core protocols in the Substrate context include either: 

1. Have the ability to do this with ink! (to my knowledge possible, but no one has done it)
2. A pallet to do this (social recovery)

In short, either the **runtime** needs to be altered or the implementation/enforcement of an ink! smart contract.

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,61 @@
...
+sign any transaction or, more generally, any message using your private key. Data is recorded on a
+public ledger (blockchain) whose multiple copies of it are stored in computers participating in a
+P2P network.
+
+While the account keys grant users control and ownership, losing them results in losing access to digital assets and fragmentation of your digital identity since you will need to create a new account with a
+new set of keys. This poses a hurdle for both users and developers regarding security and
```
CrackTheCode016: It could be we emphasize that the user is _still_ responsible for their keys - that has not changed, but that they can take precautions using the functionality that Polkadot provides in order to ensure they don't end up losing their accounts.

CrackTheCode016 (commented):
Overall it looks good to me, it sets the stage very well for why this is important.  Maybe we could emphasize that this can be done at either the: 

1. Protocol level with pallets
2. Smart contract level with ink!
3. Emphasize how types in Substrate are malleable, enabling for multiple address and schemes to be abstracted to even support different protocols (like Address20 for ETH, for example)

WDYT?

filippoweb3:
@DrW3RK let me know if the edits and additions make sense, and if we need somebody to provide additional review

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+it are stored in computers participating in a P2P network.
+
+While the account’s private keys grant users control and ownership, losing them results in losing
+access to digital assets and fragmentation of your digital identity since you will need to create a
+new account with a new set of keys. This poses a hurdle for both users and developers regarding
+security and adoption.
+
+## Definition of Account Abstraction
+
+The term _abstraction_ refers to separating the user experience from the private key, enabling a
```
CrackTheCode016: I'm not sure if it's a good idea, but we could mention that the contept was derived from Ethereum first via erc4337

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+originally were not engineered to be flexible and decreased chances of key mismanagement.
+
+Users are still responsible for their keys, but through account abstraction, they can take
+precautions to ensure they do not end up losing their accounts.
+
+Account abstraction introduces a layer of on-chain logic that controls an account, typically in the
+form of a smart contract. Without a smart contract, abstracting accounts would require changes in
+the core architecture of the protocol.
+
+The highly generic codebase of Polkadot's architecture makes accounts natively flexible and abstract
```
CrackTheCode016:
```suggestion
Polkadot's generic codebase makes the concept of an account natively flexible and abstract
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+without the direct need for smart contracts.
+
+## Origin Abstraction in Polkadot
+
+Adopting a generic design is crucial in scaling Web3 technologies. Abstraction and generalization of
+protocols are essential to improving user experience and security in blockchain adoption.
+
```
CrackTheCode016, line 47: Not sure if this counts as 'opinionated' for Wiki contenxt, although I think it's nice to include

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+When users interact with a blockchain they call _dispatchable_ functions to do something. Because
+those functions are called from the outside of the blockchain interface, in Polkadot's terms any
+action that involves a dispatchable function is an [extrinsic](./learn-extrinsics.md). Extrinsics
+are calls coming from the _outside_ of the blockchain interface that (if successfully executed)
+invoke some changes in the _inside_ of a blockchain, the blockchain's state. An extrinsic is always
```
CrackTheCode016:
```suggestion
invoke some changes in the _inside_ of the blockchain's state. An extrinsic is always
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+directed to a specific function within a particular pallet (or module).
```
CrackTheCode016:
```suggestion
directed to a specific function within a particular [pallet](../general/glossary.md#pallet).
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+**Accounts happen to be just one variant (or corner case) of Substrate's FRAME possible origins, the
+`frame_system::RawOrigin::Signed`.** In Substrate, the concept of account is completely
+deprioritized. Substrate itself remains indifferent to an account's balance and nonce. While FRAME
+can support their presence, it fundamentally does not need to rely on them.
+
+On a lower level, the [Cross-Consensus Messaging (XCM)](./learn-xcm-index) format provides a much
+powerful origin abstraction that allows calling personas that are so abstract to not necessarily
+have direct representation on the local chain within its FRAME system origin.
+
+### Protocol-level Account Abstraction
```
CrackTheCode016, line 82: Lately, I've been seeing the word 'enshrined' to describe things that are natively part of the protocol.  I wonder if we should start using it here too?

```suggestion
### Enshrined Account Abstraction
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+- [multi-signature accounts](./learn-account-multisig.md) to control an account using different ones
+- [proxy accounts](./learn-proxies.md) for role-based representation, and ownership representation
+  through [pure proxies](./learn-proxies.md#anonymous-proxy-pure-proxy)
+- [derivative accounts](./learn-account-advanced.md#derivation-paths) for using the same _parent_
+  private key on multiple _children_ accounts
+- account recovery mechanisms such as social recovery to help regain access to your key using
+  trusted third-party accounts
+- [batching functionality](./learn-balance-transfers.md#batch-transfers) to submit multiple calls in
+  one single transaction
+- payments with non-native tokens
```
CrackTheCode016:
```suggestion
- [Multi-signature accounts](./learn-account-multisig.md) to control an account using different ones
- [Proxy accounts](./learn-proxies.md) for role-based representation, and ownership representation
  through [pure proxies](./learn-proxies.md#anonymous-proxy-pure-proxy)
- [Derivative accounts](./learn-account-advanced.md#derivation-paths) for using the same _parent_
  private key on multiple _children_ accounts
- Account recovery mechanisms such as social recovery to help regain access to your key using
  trusted third-party accounts
- [Batching functionality](./learn-balance-transfers.md#batch-transfers) to submit multiple calls in
  one single transaction
- Payments with non-native tokens
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+[Polkadot SDK](https://github.com/paritytech/polkadot-sdk).
+
+**Accounts happen to be just one variant (or corner case) of Substrate's FRAME possible origins, the
+`frame_system::RawOrigin::Signed`.** In Substrate, the concept of account is completely
+deprioritized. Substrate itself remains indifferent to an account's balance and nonce. While FRAME
+can support their presence, it fundamentally does not need to rely on them.
+
+On a lower level, the [Cross-Consensus Messaging (XCM)](./learn-xcm-index) format provides a much
+powerful origin abstraction that allows calling personas that are so abstract to not necessarily
+have direct representation on the local chain within its FRAME system origin.
```
CrackTheCode016:
```suggestion
On a lower level, the [Cross-Consensus Messaging (XCM)](./learn-xcm-index) format can take advantage of origin abstraction for cross-consensus communications by specifying the context for a particular message.  Origins in this case imply the authority under which a message is being sent (and thereby, executed).
```

docs/learn/learn-account-abstraction.md
```diff
@@ -0,0 +1,114 @@
...
+funds between two accounts, changing the balances of those accounts and thus the chain state (as
+accounts hold some state within the blockchain).
+
+In [Substrate](../general/glossary.md#substrate)'s FRAME, functions are not necessarily called by
+accounts. Functions can be called by any origin, where origins are caller-personas associated with
+privilege levels. For example, the Polkadot [OpenGov](./learn-polkadot-opengov.md) has different
+origins with different privileges, such as allocating treasury funds, cancelling a referendum, etc.
+Neither of those origins is subservient to the concept of an account or assume anything about state
+or associated data. Custom origins can be created while designing your chain using the
+[Polkadot SDK](https://github.com/paritytech/polkadot-sdk).
```
CrackTheCode016: We mention Polkadot SDK here, but substrate somewhere else.  It's not a huge deal but it should probably be consistent

CrackTheCode016 (commented):
A few minor things, but I think after they are addressed it LGTM!

docs/learn/learn-account-abstraction.md
filippoweb3: I would opt for keeping "protocol-level" as more accessible term to a wider audience

docs/learn/learn-account-abstraction.md
filippoweb3: added together with the previous paragraph about XCM

CrackTheCode016 (approved):
reads very well, amazing job!

PR #5176, opened 2023-09-11
Reorg Async Backing

filippoweb3:
- reorg content
- figure addition


filippoweb3:
@BradleyOlson64 @CrackTheCode016 I reorganized the text after reading the academy docs. Can you please check if everything makes sense? I think it would be beneficial to add some figures that explain how asynch baking can be configured in case of high demand and low demand. As far as I understand, the length of the included segments plays an important role.

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
 slug: ../learn-async-backing
 ---
 
+import RPC from "./../../components/RPC-Connection";
+
+:::info Learn about the Parachain Consensus
+
+To fully follow the material on this page, it is recommended to be familiar with the main stages of
+the [Parachain Protocol](./learn-parachains-protocol.md).
+
```
CrackTheCode016:
```suggestion
:::info Learn about Parachain Consensus

To fully follow the material on this page, it is recommended to be familiar with the primary stages of
the [Parachain Protocol](./learn-parachains-protocol.md).

```

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
...
-segment of block ancestors upon which the newest parablock is built.
-
-These blocks can be prepared in anticipation of being included later rather than keeping in sync
-with the relay chain's progress 1-1. The block can be in a different stage from another block being
-built, as long as it abides by the parameters set forth by the asynchronous backing configuration.
+Asynchronous backing brings the following changes to the parachain protocol:
+
+- parablock generation is decoupled from the backing process, this avoids rushing through the
+  backing process in six seconds and gives more time to collators to properly use blockspace,
+  decreasing the probability of generating empty parablocks
```
CrackTheCode016: I'm not sure this is implemented _yet_. (the last sentence) If I recall, this is more in the scope of elastic scaling.

```suggestion
- Parablock generation is now decoupled from the backing process, avoiding the previous 6-second deadline through the backing process. This gives more time to collators to properly use blockspace,
  decreasing the probability of generating empty parablocks.
```

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
...
+- parachains can generate blocks and have them cashed (or stacked) into **unincluded segments** of
+  parablock ancestors (i.e. parablocks that are seconded but not yet backed by paravalidators and
+  included in the relay chain) rather than ancestors included in the relay chain state.
```
CrackTheCode016: I suggested the removal of the stacked / caching wording, but I think in the graphic it would be better. 

```suggestion
- Parachains can generate blocks and have them placed into **unincluded segments** of
  parablock ancestors (i.e. parablocks that are seconded but not yet backed by paravalidators and
  included in the relay chain) rather than ancestors included in the relay chain state.
```

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
...
+
+These changes allow to generate parablock `P + 1` soon after parablock `P` has been generated and
+cashed into the unicluded segment within the first six-second time window. At the same time,
+parablock ancestor `P - 1`, pushed to the unincluded segment before `P`, is backed and included in
+the relay chain.
+
```
CrackTheCode016: We probably need to talk more about the prospective parachains subsystem here, along with the concept of parachain fragments.  Prospective parachains keep track of the various paths that a parachain's state can be in, of which each prospective parablock may be in a different state.

```suggestion
These changes allow for a parablock `P + 1`, to begin the generation process soon after parablock `P` has been generated and appended into the unincluded segment within the first six-second time window (the span of one relay chain block). At the same time, parablock ancestor `P - 1`, could be pushed to the unincluded segment before `P`, is backed and included in the relay chain.

```

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
...
+The **contextual execution** (which is the context built by block ancestors used by the parachains
+to generate new parablocks) shifts from being the latest included parent block in the relay chain,
+to being the latest ancestor parablock pushed into the unincluded segment. This allows to start
+building parablocks earlier, giving plenty of time to parachain collators to fit more transactions
+and prepare block candidates for backing and inclusion.
+
+The execution context for the latest parablock is thus derived from the unincluded segment of block
+ancestors upon which the newest parablock is built. These blocks can be prepared in anticipation of
+being included later rather than keeping in sync with the relay chain's progress 1-1. A parablock
```
CrackTheCode016:
```suggestion
being included later rather than keeping in sync with the relay chain's progress one-to-one. A parablock
```

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
...
+can be in a different stage from another one being built if it abides by the parameters set forth by
+the asynchronous backing configuration.
```
CrackTheCode016: It's possible we could emphasize some aspects of the configuration to help understand the different factors, i.e., max_depth for unincluded segments

docs/learn/learn-async-backing.md
```diff
@@ -7,100 +7,111 @@ keywords: [parachains, slots, backing, parablock]
...
+
+Asynchronous backing also introduces a parameter to define the maximum number of ancestor blocks
```
CrackTheCode016: As part of the reorg, should we maybe categorize this as "Further Benefits" or similar?

CrackTheCode016 (commented):
Added a preliminary review, thank you for rearranging the content.  I have a lot of notes on prospective parachains, and I think that will give the detail we need to fully drive the point home. 

For those who aren't aware, @filippoweb3 and I wanted to also eventually answer the question: 

- With **less** time needed for backing, how can **more** blockspace be utilized in less time?  

I think it is partially answered here already, but thought I'd mention :)

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,179 @@ keywords: [parachains, slots, backing, parablock]
...
+
+Before diving into asynchronous backing it is important to understand what synchronous backing is
+and what its main limitations are.
+
+:::info What is backing?
+
+**Backing** refers to the process in which a parablock is verified by a subset of validators who
+determine the blocks validity. It is an important step in the validation process for parablocks, as
+it is the first line of defense in ensuring valid state transitions are taking place. Validators who
+back the parablock put their stake at risk, and if the block is found to be malicious, those
```
profradha:
```suggestion
back the parablock put their stake at risk, and if the block is later found to be invalid, those
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,179 @@ keywords: [parachains, slots, backing, parablock]
...
+(six-second window), and (if successfully backed) it will be included in a relay-chain block (often
+referred to as the **parent** block, as the parablock anchors itself to it) after an additional six
+seconds. Thus, a parablock can be produced every 12 seconds because a new parablock can be produced
+after the inclusion of the previous one.
+
+The **contextual execution** of new parablocks are parablock ancestors included in the relay chain
+after 12 seconds, where contextual execution refers to the context built by included parablock
+ancestors used by the parachains to generate new parablocks.
+
+:::info It is important to note the clarification below
```
profradha:
```suggestion
:::info Candidate Recepts
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,179 @@ keywords: [parachains, slots, backing, parablock]
...
+
+- Not the entire parablock, but the **paraheader** is placed on the parent block on the relay chain.
+- The relay chain does not access the entire state of a parachain but only the values that changed
+  during that block and the merkelized hashes of the unchanged values.
```
profradha: Not just the headers of the parablock, but a bit more info goes into the relaychain and the formal term for it is candidate receipt https://medium.com/polkadot-network/the-path-of-a-parachain-block-47d05765d7a

```suggestion
- Not the entire parablock, but the **candidate receipt** consisting of the hash of the parablock, state roots and ID info is placed on the parent block on the relay chain.
- The relay chain does not access the entire state of a parachain but only the values that changed
  during that block and the merkelized hashes of the unchanged values.
  
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,179 @@ keywords: [parachains, slots, backing, parablock]
...
+
+:::
+
+### Synchronous Backing Diagram
```
profradha:
```suggestion
### Synchronous Backing
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,179 @@ keywords: [parachains, slots, backing, parablock]
...
+
+:::
+
+### Pipelining
+
 Asynchronous backing is a feature that introduces
 [pipelining](https://www.techtarget.com/whatis/definition/pipelining) to the parachain block
-generation and validation process. It is analogous to the logical pipelining of processor
+[generation](./learn-parachains-protocol.md), [backing](./learn-parachains-protocol.md) and
+[inclusion](./learn-parachains-protocol.md). It is analogous to the logical pipelining of processor
```
profradha:
```suggestion
[generation, backing and inclusion](./learn-parachains-protocol.md). It is analogous to the logical pipelining of processor
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+it is the first line of defense in ensuring valid state transitions. Validators who back the
+parablock put their stake at risk, and if the block is later found invalid, those validators are
+slashed.
+
+:::
+
+In synchronous backing, parablock generation is tightly coupled to the relay chain's progression on
+a one-to-one basis. Every parablock must be generated and backed within a relay-chain block
+(six-second window), and (if successfully backed) it will be included in a relay-chain block (often
+referred to as the **parent** block, as the parablock anchors itself to it) after an additional six
```
CrackTheCode016: I'm the one who probably uses parent block, but maybe we should keep it consistent and just use relay parent? Switching between them could be confusing for a new reader. 

```suggestion
referred to as the **relay parent**, or latest block on the relay chain, as the parablock anchors itself to it) after an additional six
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+:::
+
+### Synchronous Backing Mechanics
 
-:::info
+In synchronous backing we can imagine parablocks on a conveyor belt with the following properties:
 
-In order to realize which exact processes this upgrade improves, it is recommended to be familiar
-with the stages of parablock validation, which you may [read here](./learn-parachains-protocol.md).
+- the belt is synched with the relay chain: every parablock is generated and backed every 6 seconds,
```
CrackTheCode016:
```suggestion
- The belt is in sync with the relay chain - every parablock is generated and backed every 6 seconds,
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+  and included every 12 seconds
+- the belt can carry only one parablock at a time (no [pipelining](#pipelining))
```
CrackTheCode016:
```suggestion
- The belt can carry only one parablock at a time (no [pipelining](#pipelining)) and it has to wait for it to be fully processed before another can be put on the belt.
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
 
-Asynchronous backing enables logical pipelining over the parablock
-[**generation**](https://wiki.polkadot.network/docs/learn-parachains-protocol#collators),
-[**backing**](https://wiki.polkadot.network/docs/learn-parachains-protocol#parachain-phase), and
-[**inclusion**](https://wiki.polkadot.network/docs/learn-parachains-protocol#inclusion-pipeline)
-processes. A parablock may be at different stages, but multiple parablocks can be processed
-simultaneously (in parallel) if needed. Most notably, parablock N + 1 can be generated and backed
-while its predecessor, parablock N, is undergoing inclusion on the relay chain. Processes can occur
-while their ancestors are included on the relay chain.
+### Pipelining
```
CrackTheCode016: Perhaps we can take out pipelining here, as we want to introduce what async backing is.  Pipelining is a good way to describe it, but it isn't _fully_ what async backing represents IMHO.

```suggestion
```

filippoweb3:
@BradleyOlson64 we added some graphics to the page and reorganized the content. Can you provide a final review of the page content and figures? Cheers!

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+for the latest included parent block of the relay chain. The core functionality that asynchronous
+backing brings is the ability to build on these unincluded segments of block ancestors rather than
+ancestors included in the relay chain state.
 
-1. The parablock is not placed on the relay chain in its entirety, but rather the **paraheader**
-2. The relay chain does not access the entire state of a parachain but only the values that changed
-   during that block and the merkelized hashes of the unchanged values.
+:::info Unincluded segments happen on the collator portion
+
+**Unincluded** segments are blocks that are seconded but still need to be backed. The backing
```
CrackTheCode016:
```suggestion
**Unincluded** segments are generated blocks that still need to be backed. The backing
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
-process during the next block.
+- Parachains can generate blocks and have them placed into **unincluded segments** of parablock
+  ancestors (i.e. parablocks that are seconded but not yet backed by paravalidators and included in
+  the relay chain) rather than ancestors included in the relay chain state.
 
-:::info
+### Unincluded Segments
 
-It is important to note the following clarifications:
+Unincluded segments are chains of candidate parablocks that are seconded but still need to be backed
```
CrackTheCode016:
```suggestion
Unincluded segments are chains of candidate parablocks that still need to be backed
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+  giving more time for collators to properly use blockspace or increasing block velocity based on
+  demand.
 
-Currently, parablocks rely on the most recent relay chain block (often referred to as the **parent**
-block, as the parablock anchors itself to it). Each parablock must be generated and go through the
-entire backing process in a single relay block. It then proceeds to the availability + inclusion
-process during the next block.
+- Parachains can generate blocks and have them placed into **unincluded segments** of parablock
+  ancestors (i.e. parablocks that are seconded but not yet backed by paravalidators and included in
+  the relay chain) rather than ancestors included in the relay chain state.
```
CrackTheCode016:
```suggestion
  ancestors rather than the latest, included relay chain parent.
```

CrackTheCode016 (commented):
Overall the graphics are good, maybe a little busy (but understandable given the content). 

A small concern would be emphasizing the usage of blockspace more than the actual async backing mechanism.  I think it's good to mention, but perhaps is slightly outside of the scope of this specific document, but lets see what the others say too :)

docs/learn/learn-async-backing.md
filippoweb3: I would leave it as it helped me to get the picture of what is happening and there are some parts linked to it

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+throughput of the entire network.
+
+### Synchronous vs. Asynchronous Backing
 
 Asynchronous Backing has three overarching goals:
 
 1. Decrease parachain block (parablock) validation time to **6 seconds** from **12 seconds**
-2. Increase the amount of usable blockspace, allowing more state changes per relay chain block.
-3. Allow for parablocks to be re-proposed to the network if they are not included successfully on
+2. **Increase the usable blockspace**, allowing more state changes per relay chain block.
```
BradleyOlson64: Flows a tiny bit better
```suggestion
2. **Increase usable blockspace**, allowing more state changes per relay chain block.
```

docs/learn/learn-async-backing.md
BradleyOlson64: There is indeed more to it than just pipelining, though that doesn't make it wrong to spend time hammering home the pipelining comparison. If the reader takes away any one thing, I'd have it be the notion of completing the backing and inclusion steps for different blocks at the same time.

Here's a useful extension of the pipelining comparison. Async backing doesn't just allow for pipelining within a single pipe. It laid the foundation for a large number of pipes (~10 cores) to run for the same parachain at the same time. In that way we have two distinct new forms of parallel computation.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+:::
+
+In synchronous backing, parablock generation is tightly coupled to the relay chain's progression on
+a one-to-one basis. Every parablock must be generated and backed within a relay-chain block
+(six-second window), and (if successfully backed) it will be included in a relay-chain block (often
+referred to as the **parent** block, as the parablock anchors itself to it) after an additional six
+seconds. Thus, a parablock can be produced every 12 seconds because a new parablock can be produced
+after including the previous one.
+
+The **contextual execution** of new parablocks are parablock ancestors included in the relay chain
```
BradleyOlson64: I don't see why we're discussing contextual execution at this point. I'd suggest either removing this paragraph entirely or rewording it and placing it later on. 

Put simply, here's the difference in execution context between synchronous and asynchronous backing:

Synchronous - The entire execution context comes from the latest relay block and most recently included parablock from the same parachain.

Asynchronous - A baseline execution context is first set up using a parablock's relay parent and the most recently included parablock as of that relay parent. The relay parent need not be the most recent relay block.

From here, we modify the context for each unincluded parablock predecessor of the block under construction. Said predecessors live in the unincluded segment.

In this way, synchronous backing uses execution context entirely pulled from the relay chain. While asynchronous backing augments this with off-chain context from the unincluded segment.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+after 12 seconds, where contextual execution refers to the context built by included parablock
+ancestors used by the parachains to generate new parablocks.
+
+:::info Candidate Recepts
```
BradleyOlson64: This also seems like it could be moved later on or removed

docs/learn/learn-async-backing.md
BradleyOlson64:
```suggestion
- the belt is synched with the relay chain: It takes six seconds (one relay block) to generate and back a parablock. It then takes another six seconds to make that parablock available and mark it as included on the relay chain. The full process, from one end of the belt to the other, takes twelve seconds, and can't be started for parablock N + 1 until it is fully complete for parablock N (no pipelining).
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
-:::info
+In synchronous backing we can imagine parablocks on a conveyor belt with the following properties:
 
-In order to realize which exact processes this upgrade improves, it is recommended to be familiar
-with the stages of parablock validation, which you may [read here](./learn-parachains-protocol.md).
+- the belt is synched with the relay chain: every parablock is generated and backed every 6 seconds,
+  and included every 12 seconds
+- the belt can carry only one parablock at a time (no [pipelining](#pipelining))
+
+The diagram below shows parablocks on their way from being generated to being backed and included
```
BradleyOlson64: I'd suggest stripping out the visualization of how full of transactions parablocks are to begin with. Also you could separate out the basic pipelining visual on the right into its own graphic to be shown first. The end result would be three different visuals with increasing complexity:
1. Simple pipelining table
2. This graphic without parablock transaction fill accounted for
3. This graphic in full

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+3. **Allow parablocks to be re-proposed** to the network if they are not included successfully on
    the first attempt.
 
-Asynchronous backing provides a form of **contextual execution**, which allows for more time for
-parachain collators to fit more transactions and prepare block candidates for backing and inclusion.
-**Contextual execution** refers to how a parablock can be built earlier using the context provided
-by an _unincluded segment_ of recent block ancestors.
+In synchronous backing, parablock generation relies on the most recent relay-chain block. Each
+parablock must be generated and go through the entire backing process in a single relay block. It
+then proceeds to the inclusion process while generating the next relay-chain block.
```
BradleyOlson64:
```suggestion
then proceeds be made available and marked as included in the next relay block.
```

docs/learn/learn-async-backing.md
BradleyOlson64: Unincluded segments can contain blocks that have been backed on the relay chain. They can contain parablocks at any stage pre-inclusion.
```suggestion
Unincluded segments are chains of candidate parablocks have yet to be included in the relay chain.
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+backing brings is the ability to build on these unincluded segments of block ancestors rather than
+ancestors included in the relay chain state.
 
-1. The parablock is not placed on the relay chain in its entirety, but rather the **paraheader**
-2. The relay chain does not access the entire state of a parachain but only the values that changed
-   during that block and the merkelized hashes of the unchanged values.
+:::info Unincluded segments happen on the collator portion
+
+**Unincluded** segments are blocks that are seconded but still need to be backed. The backing
+process occurs on the relay chain, whereas unincluded segments are moreso held in the collator.
```
BradleyOlson64:
```suggestion
process occurs on the relay chain, whereas unincluded segments live in the runtimes of parachain collators.
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
 
 :::
 
-For more information on the validity and availability process, be sure to visit the
-[parachain protocol](../learn/learn-parachains-protocol.md) page.
+Compared to synchronous backing, contextual execution shifts from being the parablock ancestors
```
BradleyOlson64: See my earlier comment for a more complete description of what's going on here. Part of the execution context still comes from the last included ancestor.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+- The conveyor belt can carry multiple parablocks ([pipelining](#pipelining)), and a parablock can
+  be in a different stage from another one being built if it abides by the parameters set forth by
+  the asynchronous backing configuration.
+- Backing and inclusion events can happen within the same relay chain block (i.e. in 6 seconds)
+- The last parablock in the unincluded segment is used as execution context to generate a new
+  parablock and, because the unincluded segment can carry multiple parablocks, parachains can take
+  more time to fill new blocks while the segment is emptied.
+
+The diagram below shows parablocks on their way from being generated to being backed and included
+into the relay chain in the context of asynchronous backing.
```
BradleyOlson64: This diagram just isn't correct. Every parablock candidate recorded in the unincluded segment is immediately advertised to validators to begin the backing process. And the unincluded segment keeps record of all candidates in the backing process right up until they are included.

Think of it this way. The unincluded segment isn't a step in the pipeline. It's just a record of all the blocks in the pipeline currently. We need that record to determine which parablocks we can build and when.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+
+![async-backing](../assets/async-backing.png)
+![async-backing-legend](../assets/async-backing-legend.png)
+
+Parablock 1 (P1) is included, and P2 is backed within R1. In the meantime, the unincluded segment is
+full and contains P3-5. Because P6 is not rushing to be backed, it can be filled in 6s while P2 is
```
BradleyOlson64: Even with async backing, we're only recommending 2 seconds of block building/filling time. This limit may change slightly in the future as we optimize block validation in the backing and approval steps.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+included and P3 is backed in R2. This is possible because the context execution shifted from being
+the latest included parablock to being the latest ancestor in the unincluded segment (i.e. P5).
+
+In this scenario, blockspace demand decreases and as a consequence P6 is filled only 50% using 6
+seconds. Next block P7 can be filled in 12 seconds while the unincluded segment is emptied. While
```
BradleyOlson64: Validators are the limiting factor here. Even if a collator has 12 seconds to produce a block, the validator will still have less than 6 seconds (really ~2 seconds) to check it.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+filling up P7, P3 and P4 are included in R3 and R4, respectively. In 24 seconds, four parablocks, P1
+to P4, have been included in the relay chain, P5 has been backed, and two more blocks have been
+filled and pushed to the unincluded segment, ready to be backed. The unincluded segment is now 2/3
+filled.
+
+Collators were able to make better use of blockspace, including 2x more data, by doubling block
+generation time from 6 to 12 seconds in a period of lower demand. This was possible due to the
```
BradleyOlson64: We previously only allowed 0.5 seconds of block generation time, since we needed the other 5.5 seconds to get the block backed on chain.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,232 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
-within six seconds. This time limit reduces the amount of data a collator has time to add to each
-block.
+Prospective parachains manage and represent all unincluded segments from all parachains.
 
-A particular parablock, `P1`, would **only** be valid for backing at relay chain parent `R1 + 1` and
-subsequently, be included at `R1 + 2` should it be backed successfully. Essentially, a parablock is
-rushing to be backed and included within this two-block window due to the inherent requirement for
-synchrony between the parachain and relay chain.
+The
+[Prospective Parachains subsystem](https://paritytech.github.io/polkadot/book/node/backing/prospective-parachains.html),
```
BradleyOlson64: A better way to describe prospective parachains:
Prospective parachains is the relay chain's record of all parablock candidates undergoing the backing and inclusion process. It simultaniously keeps track of candidates from all forks of all parachains.

It is as if you folded the unincluded segments from every fork of every parachain into one giant data structure. When you fold unincluded segments representing different chain forks together, they create a tree structure. Hence the term fragment tree.

A single unincluded segment tells a collator whether it can build on top of one fork of one parachain.

Prospective parachains tells a validator whether it should accept blocks built on top of any fork from any parachain.

BradleyOlson64 (commented):
Didn't quite nail the concepts this time. But async backing isn't exactly easy to understand. Keep at it!

filippoweb3:
@BradleyOlson64 thanks for the review and the suggestions! How does it read now? Are the figures better?

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,317 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+
+### Sync Backing Mechanics
+
+The diagram below shows the pipelining table for synchronous backing.
+
+![sync-backing-pipeline](../assets/sync-backing-pipeline.png)
+
+Backing (B) and inclusion (I) of each parablock happen on different relay chain blocks, and new
+parablocks are generated once the previous parablock has been included.
+[Contextual execution](#contextual-execution) is driven by the latest included parablock in the most
```
ordian: contextual execution is an old name for async backing, so it should not be used

filippoweb3: Hi! :) The term has already passed 3 rounds of review, and it is mentioned in the PBA material: https://polkadot-blockchain-academy.github.io/pba-book/polkadot/async-backing-shallow/page.html#async-backing-another-look (shallow dive async backing). Are you 200% sure?

docs/learn/learn-async-backing.md
```diff
@@ -34,43 +34,51 @@ slashed.
...
+recent relay parent. After 24 seconds, two parablocks, P1 and P2, have been included in the relay
+chain, and P3 has been backed.
 
-### Synchronous Backing Mechanics
+In synchronous backing we can imagine parablocks on a conveyor belt. The belt is synched with the
+relay chain: It takes six seconds (one relay block) to generate and back a parablock. It then takes
+another six seconds to make that parablock available and mark it as included on the relay parent.
```
BradleyOlson64: The relay parent of a parablock and the relay block including that parablock are always different.
```suggestion
another six seconds to make that parablock available and mark it as included in a relay block.
```

docs/learn/learn-async-backing.md
```diff
@@ -122,74 +117,75 @@ Asynchronous backing brings the following changes to the parachain protocol:
...
+Compared to [synchronous backing](#synchronous-backing-mechanics), in asynchronous backing:
 
-In asynchronous backing, parablock generation is decoupled from backing and inclusion. Compared to
-[synchronous backing](#synchronous-backing-mechanics):
-
-- Parablocks are not rushing to being backed, they are not placed on the conveyor belt when they are
-  generated
-- Filled blocks are placed in the unincluded segment, and from there, they are injected onto the
-  conveyor belt
+- Filled blocks are placed in the unincluded segment, a record of yet not included parablocks stored
```
BradleyOlson64:
```suggestion
- Filled blocks are placed in the unincluded segment, a record of not yet included parablocks stored
```

docs/learn/learn-async-backing.md
```diff
@@ -122,74 +117,75 @@ Asynchronous backing brings the following changes to the parachain protocol:
...
+  locally on the parachain side.
+- Every parablock candidate recorded in the unincluded segment is immediately advertised to
+  validators to begin the backing process. And the unincluded segment keeps record of all candidates
+  in the backing process right up until they are included (in the unincluded segments there
```
BradleyOlson64:
```suggestion
  in the backing process right up until they are included. An unincluded segment may include candidates that are seconded, backable, or backed.
```

docs/learn/learn-async-backing.md
```diff
@@ -122,74 +117,75 @@ Asynchronous backing brings the following changes to the parachain protocol:
...
+  parablocks that are seconded, backable, and backed).
+- A relay parent (not necessarily the latest block) and the parablock ancestors in the unincluded
+  segment are used as execution context to generate a new parablock and, because the unincluded
```
BradleyOlson64:
```suggestion
  segment are used as execution context to generate a new parablock. Because the unincluded
```

docs/learn/learn-async-backing.md
```diff
@@ -200,38 +196,127 @@ blocks in advance to fit more data, Polkadot with asynchronous backing can deliv
...
+backing is as follows:
+
+In synchronous backing, the entire execution context comes from the latest relay block and most
+recently included parablock from the same parachain.
+
+In asynchronous backing, a baseline execution context is first set up using a parablock's relay
+parent and the most recently included parablock as of that relay parent. The relay parent need not
+be the most recent relay block. Predecessors of the block under construction live in the
+[unincluded segment](#unincluded-segments) placed directly in the runtime of parachain collators
+(i.e. off-chain).
```
BradleyOlson64: I was wrong to use the term off-chain here. Ancestor blocks in the unincluded segment aren't stored on the relay chain. But they are stored on-chain in the parachain runtime.
```suggestion
```

docs/learn/learn-async-backing.md
```diff
@@ -200,38 +196,127 @@ blocks in advance to fit more data, Polkadot with asynchronous backing can deliv
...
+
+Synchronous backing uses execution context entirely pulled from the relay chain. While asynchronous
+backing augments this with off-chain context from the unincluded segment.
```
BradleyOlson64:
```suggestion
backing augments this with additional context from the unincluded segment.
```

docs/learn/learn-async-backing.md
```diff
@@ -200,38 +196,127 @@ blocks in advance to fit more data, Polkadot with asynchronous backing can deliv
...
+### Unincluded Segments
+
+The purpose of each unincluded segment is two fold:
+
+- Make each parachain aware of when and at what depth it can build blocks that won't be rejected by
+  the relay chain
+- Provide critical context necessary to build parablocks with parent blocks that have yet to be
+  included. The unincluded segment is all about building parablocks.
+
+Unincluded segments are chains of candidate parablocks that have yet to be included in the relay
```
BradleyOlson64: I'd recommend the following:
1. Switch the order of paragraphs one and two so that two comes first
2. Get rid of paragraph three here. If not too repetitive, you could move it to the contextual execution section.

docs/learn/learn-async-backing.md
```diff
@@ -200,38 +196,127 @@ blocks in advance to fit more data, Polkadot with asynchronous backing can deliv
...
+included in the latest relay parent to being the latest ancestor parablock pushed into the
+unincluded segment. This allows collators to build parablocks earlier, giving them plenty of time to
+fit more transactions and prepare block candidates for backing and inclusion.
+
 ### Prospective Parachains
 
-Prospective parachains manage and represent all unincluded segments from all parachains.
+The purpose of
+[prospective parachains](https://paritytech.github.io/polkadot/book/node/backing/prospective-parachains.html)
+is also two fold:
```
BradleyOlson64:
```suggestion
is two fold:
```

BradleyOlson64 (approved):
Much better. Just a few more suggestions.

docs/learn/learn-async-backing.md
filippoweb3: will change the term to something else, like "parablock building context". Is that ok @ordian ?

ordian: I don't see any mention of "Contextual execution" on the PBA page you linked, there's only a mention of context.

Quoting from https://github.com/paritytech/polkadot/issues/3779:
> I previously referred to this as "Contextual Execution", but I'd prefer to rename it to "Asynchronous Backing", which is the goal, rather than "contextual execution" which doesn't describe the purpose very well and is fact an implementation detail.

ordian: Deferring to @rphmeier on this

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+Before diving into asynchronous backing, it is important to understand what synchronous backing is
+and what its main limitations are.
 
-:::info
+:::info What is backing?
 
-In order to realize which exact processes this upgrade improves, it is recommended to be familiar
-with the stages of parablock validation, which you may [read here](./learn-parachains-protocol.md).
+**Backing** refers to the process in which a parablock is verified by a subset of validators who
+determine the block’s validity. It is an important step in the validation process for parablocks, as
```
Sophia-Gold: The first sentence here makes it easy to confuse backing and approval checking. Backing groups are for censorship resistance. You only need to be backed by one validator so it's not ensuring validity in any meaningful sense.

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+chain, and P3 has been backed.
+
+In synchronous backing we can imagine parablocks on a conveyor belt. The belt is synched with the
+relay chain: It takes six seconds (one relay block) to generate and back a parablock. It then takes
+another six seconds to make that parablock available and mark it as included in a relay block. The
+full process, from one end of the belt to the other, takes twelve seconds, and cannot be started for
+parablock N + 1 until it is fully complete for parablock N (no [pipelining](#pipelining)).
+
+The diagram below shows parablocks on the conveyor belt, on their way from being generated to being
+backed and included into the relay parent.
```
Sophia-Gold: This is a technical spec. You don't need this conveyor belt metaphor. I don't think it makes this easier to understand either.

docs/assets/sync-backing-nofill.png
Sophia-Gold, line 1: This is very confusing. Block numbers increase with time. What does each conveyor belt represent? The outlines for parablocks are too similar so I can't really make them out.

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
-blocks, which allows for a parablock to be backed later in the future, enabling more computational
-and storage time per block. The context for the latest parablock is derived from the unincluded
-segment of block ancestors upon which the newest parablock is built.
+![async-backing-fill-scenario2](../assets/async-backing-fill-scenario2.png)
 
-These blocks can be prepared in anticipation of being included later rather than keeping in sync
-with the relay chain's progress 1-1. The block can be in a different stage from another block being
-built, as long as it abides by the parameters set forth by the asynchronous backing configuration.
+In general, asynchronous backing will allow [collators](./learn-parachains-protocol.md#collators) to
+include an estimated ~3-5x more data into parablocks while speeding up parachain block times from 12
```
Sophia-Gold: Where did you get 3-5x?

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+to 6 seconds. Due to the 2x decrease in block time to inclusion and the possibility of building
+blocks in advance to fit more data, Polkadot with asynchronous backing can deliver an estimated
+~6-10x more blockspace to its parachains.
 
 This combination of lower latency, higher storage per block, and a logical pipeline spanning
 Polkadot's networking, runtime, and collation aspects will allow for higher, more robust throughput.
```
Sophia-Gold: This doesn't increase storage per block, it increase compute time for collators (what we call "authoring duration" in Cumulus).

"Logical pipeline spanning Polkadot's networking, runtime, and collation aspects" doesn't make sense. Pipelining means you build blocks in advance before they're included. It has little to do with networking or the Polkadot runtime. 

"More rubust throughput" doesn't make sense.

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
 
+### Sync Backing as corner case of Async Backing
```
Sophia-Gold: Just say "special case". Usually "corner case" implies a bug.

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
-[**inclusion**](https://wiki.polkadot.network/docs/learn-parachains-protocol#inclusion-pipeline)
-processes. A parablock may be at different stages, but multiple parablocks can be processed
-simultaneously (in parallel) if needed. Most notably, parablock N + 1 can be generated and backed
-while its predecessor, parablock N, is undergoing inclusion on the relay chain. Processes can occur
-while their ancestors are included on the relay chain.
+In synchronous backing, parablock generation is tightly coupled to the relay chain's progression on
+a one-to-one basis. Every parablock must be generated and backed within a relay-chain block
+(six-second window), and (if successfully backed) it will be included in a relay-chain block (often
+referred to as the **relay parent** as the parablock anchors itself to it) after an additional six
+seconds. Thus, a parablock can be produced every 12 seconds because a new parablock can be produced
```
Sophia-Gold: I would mention that inclusion means other validators have attested to having received erasure coded chunks of the parablock data. You don't need to go in depth about this, just link to another part of the wiki that covers availability. The point is that this is why inclusion takes two relay chain blocks. Then from hereon out you can discuss backing and inclusion as one step to simplify your explanatoin.

docs/assets/sync-backing-nofill.png
Sophia-Gold: Maybe try something like this instead (from Rob's tracking issue):
```
Without async backing:
R<--- R1<-------- R2<----- R3
 \--- P1(b)<----- P1(i)
       \--------- P2(b)<-- P2(i)

With async backing
R<--- R1<-------- R2<-------- R3<-------- R4<------- R5
 \--x--|--------- P1(b)<----- P1(i)
        \--x------|---------- P2(b)<----- P2(i)
                   \---x----------------- P3(b)<---- P3(i)
```

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+
+![sync-backing-fill](../assets/sync-backing-fill.png)
+![sync-backing-fill-legend](../assets/sync-backing-fill-legend.png)
+
+Because P2 is rushing to be backed into R2 in 6 seconds, there are less than 6 seconds (~ 0.5
+seconds) to fill it, since the other 5.5 seconds are needed to get the block backed on chain. In
+this scenario, P2 is filled up to 70%. After the inclusion of P2 into R3, Parablock 3 (P3) will be
+generated, filled to 50%, and backed into R4 in 6 seconds. This shows that, if blockspace demand
+decreases, collators will always have 0.5 seconds to fill up blocks, leading to blocks that contain
+less and less data.
```
Sophia-Gold: Don't include examples of how full the blocks are. It's too complicated and inaccurate. Just use time.

The point here is that collating on top of unincluded parents means the only requirement for collator authoring duration is how long it takes the backer to execute it (2s). The 6s relay chain block delay includes this backing execution timeout (2s) and some time for network latency (time it takes to gossip messages across the entire network).

docs/learn/learn-async-backing.md
```diff
@@ -7,104 +7,312 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+
+The parablock generation and backing are bound together within a six-second window that limits the
+amount of data a collator can add to each parablock. Essentially, a parablock is limited to the
+requirement of being backed in six seconds, leaving little time for its generation and its
+blockspace to be properly filled.
```
Sophia-Gold: It's really that collators are waiting to see that the previous block was included before starting a new collation. That's both why they're limited to 12s and why the time they have to generate the block is so short.

docs/learn/learn-async-backing.md
filippoweb3: deleted

docs/learn/learn-async-backing.md
filippoweb3: PBA documentation, deleted and made it more general

docs/learn/learn-async-backing.md
filippoweb3: conveyor belt metaphor deleted

docs/learn/learn-async-backing.md
filippoweb3: all figures with filled blocks deleted, text as well. Added explanation

shawntabrizi:
@Sophia-Gold had a nice picture which concisely shows async backing and the performance increase from it:

https://youtu.be/7AWRCt_SLJc?si=N66meIh0hnh6YgRz&t=503

image

I think this image needs to be every article which talks about async backing.

EDIT: I see you have an image **like** this, but it is certainly not as clear as this one. The colors, fonts, sizing, everything just makes it harder to see simple details.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,253 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+:::info Every collator also runs an attached relay chain full node
 
-**Unincluded segments** are chains of candidate blocks that are not yet included in the relay chain.
-Parablocks can be added to this unincluded segment without waiting for the latest included parent
-block of the relay chain. The core functionality that asynchronous backing brings is the ability to
-build on these unincluded segments of block ancestors rather than ancestors included in the relay
-chain state.
+The attached relay node receives relay blocks via gossip. Then the relay node talks to the parachain
+node through the `CollationGeneration` subsystem. R2 is gossiped to the relay full node attached to
+the collator producing P2. Then CollationGeneration passes information about R2 to the collator
```
shawntabrizi:
```suggestion
the collator producing P2. Then `CollationGeneration` passes information about R2 to the collator
```

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,253 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+Due to asynchronous backing not being fully implemented in a running production network, each
+performance metric is not thoroughly tested nor guaranteed until proper benchmarking has occurred.
 
 :::
 
-For more information on the validity and availability process, be sure to visit the
-[parachain protocol](../learn/learn-parachains-protocol.md) page.
+### Synchronous vs. Asynchronous Backing
+
+Below, a table showing the main differences between synchronous and asynchronous backing.
```
shawntabrizi, line 103: i might softly suggest you start with the new graphic rather than this table to be the first thing users see to explain the difference between async and sync backing.

perhaps exactly the image i linked to from Sophia.

That at least anchors them to the existing reading, and then a table showing the benefits can naturally follow.

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,253 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+multiple unincluded parablocks because on their end they have the
+[unincluded segment](#unincluded-segments), a local storage of not included parablock ancestors that
+they can use to fetch information to build new parablocks. On the relay chain side
+[perspective parachains](#prospective-parachains) repeats the work each unincluded segment does in
+tracking candidates (as validators cannot trust the record kept on parachains).
+
+Collating on top of unincluded parents means the only time limit to generate a parablock is how long
+it takes to back it (2 seconds). The 6-second relay chain block delay includes a backing execution
+timeout (2 seconds) and some time for network latency (time it takes to gossip messages across the
+entire network). Thus, even if a collator has >2 seconds to produce a block, the validators will
```
shawntabrizi: im not sure what this sentence means, but I think you should use the more active voice.

"Collators are still limited to 2 seconds of block execution per parablock."

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,253 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+[generation, backing and inclusion](./learn-parachains-protocol.md). It is analogous to the logical
+pipelining of processor instruction in "traditional" architectures, where some instructions may be
+executed before others are complete. Instructions may also be executed in parallel, enabling
+multiple processor parts to work on potentially different instructions simultaneously.
+
+Bundles of state transitions represented as blocks may be processed similarly. In the context of
+{{ polkadot: Polkadot :polkadot }}{{ kusama: Kusama :kusama }}, pipelining aims to increase the
+throughput of the entire network by completing the backing and inclusion steps for different blocks
+at the same time. Asynchronous backing does not just allow for pipelining within a single pipe (or
+core). It lays the foundation for a large number of pipes (~10 cores) to run for the same parachain
```
shawntabrizi: where did 10 come from?

docs/learn/learn-async-backing.md
```diff
@@ -7,103 +7,253 @@ keywords: [parachains, backing, parablock, perspective parachains, unincluded se
...
+at the same time. In that way we have two distinct new forms of parallel computation: having one
+block backed and one included simultaneously using one core, and having having `n` blocks backed and
+`n` blocks included simultaneously using `n` cores.
```
shawntabrizi: nit pick, but are these actually distinct? how is n = 1 fundamentally different than n != 1?

filippoweb3: I will deleted this part

docs/learn/learn-async-backing.md
filippoweb3: I will deleted this part

filippoweb3:
Thanks for the feedback @shawntabrizi, I implemented it. I will wait for final approval by @Sophia-Gold and @BradleyOlson64 so that I can finally merge this beast!

BradleyOlson64 (approved):

shawntabrizi (approved):

DrW3RK (approved):

PR #5208, opened 2023-09-20
Edits Parachains' Protocol

filippoweb3:
- update figures
- adjust text


docs/assets/parachain-inclusion-pipeline.png
CrackTheCode016, line 1: A few notes just by glancing at the graph - if someone does not immediately read the text, it is not very obvious:

- The key on the bottom doesn't seem to differentiate between pending availability and backed blocks.
- The seconded keys don't match V3, V2, V1, unless I am misunderstanding it

CrackTheCode016 (approved):
overall, looks good! maybe we could just make the keys a little more clear to ensure understanding.

PR #5221, opened 2023-09-23
bad link error

NicolasBiondini:
## Title and link bug

I was looking at the Polkadot wiki documentation and found this bug, where it doesn't work if you click on the "Developing a dapp" section. It´s a minor bug, but I already fix it. 

### Before

Captura de pantalla 2023-09-23 a la(s) 16 53 22

Captura de pantalla 2023-09-23 a la(s) 16 54 16

Captura de pantalla 2023-09-23 a la(s) 16 54 49

### Now

Captura de pantalla 2023-09-23 a la(s) 16 59 49
It works.


CrackTheCode016 (approved):
thank you very much for this, it does indeed fix it!

PR #5250, opened 2023-10-04
Edits Auction Schedule

### LINKED ISSUE #5248: [Feature Request] Clarify Auction Section
This is rather a request to clarify the https://wiki.polkadot.network/docs/learn-auction section. I'd do it myself in a PR but it also touches the `AuctionSchedule` integration.

1) In the AuctionSchedule it says "Bidding Starts", which is a bit imprecise because the auction accepts biddings from the start of the auction (at least from the perspective of the bidder). It might be better to refer to this as "Ending Period Starts". "Bidding Ends" could easily be named "Auction Ends".
2) Further below it says "Auction 9 starts at block 9362014" which is also a bit imprecise. Actually, the auction started at block 
9335014, but the "ending phase" starts at the block referred to here. It would be better to say "The ending period of Auction 9 starts..."

filippoweb3:


docs/learn/learn-auction.md
```diff
@@ -96,7 +96,7 @@ auction on Kusama. Keep in mind that the candle phase has a uniform termination
 equal probability of ending at any given block, and the termination block cannot be predicted before
 or during the auction.
 
-- Auction 9 starts at
+- The ending period of auction 9 starts at
   [`block 9362014`](https://kusama.subscan.io/extrinsic/0x7b67d653c9522b623a97e20a967b83a8517fe3821370475ddb6611cd37c29a03?event=9335014-26).
```
jonasW3F: The link actually points to the auctionStarted event. So, indeed, the original intention was wrong. It does not make sense to point to it now. We could either point to the whole [overview of the auction](https://kusama.subscan.io/auction/9) or to the winningOffset of that auction.

filippoweb3: done, thanks!

jonasW3F (approved):

DrW3RK (approved):

PR #5258, opened 2023-10-07
Update learn-account-multisig.md and add conditional rendering for Polkadot/Kusama

muddlebee:
https://wiki.polkadot.network/docs/learn-account-multisig#introduction-to-multisig-accounts

image


filippoweb3 (approved):
lgtm, thanks!

PR #5298, opened 2023-10-24
Fix 404 wallet link

wirednkod:
Fix the `wallet` link from [Tools Index](https://wiki.polkadot.network/docs/build-tools-index#wallets) page that was leading to 404


DrW3RK (approved):
Thanks you!

PR #5305, opened 2023-10-25
Add Sub.ID to staking-apps.md

Yung-Beef:


DrW3RK:
Thank you very much for adding sub.id staking interface. Very intuitive!  Happy to merge. 

Here is some feedback on the app, if you are open to it

- sub.id can deprecate the controller account option. (It is already throwing an error as expected)
- Your rewards so far - this text is a a bit misleading.
- Please add info about the minimum active staking amount (you can fetch this value on-chain). 250 DOT can be a bit misleading.
-  Has sub.id considered running a nomination pool?

In the future, if you ever make a video tutorial of your app features, please share it on this Wiki page as well!

DrW3RK (approved):

PR #5309, opened 2023-10-26
Add Nomination Apps Page

filippoweb3:
Adding a page for apps dedicated in helping out nominators to select validators.


DrW3RK:
Hey Filippo! This can stay on the Staking apps page. I just did a dry run of their interface and it recommends you validators to stake after the questionnaire. If you think the sub-categorization of staking apps makes sense, you can do it within the staking apps page. Creating a new page for this will be doing injustice to these staking apps placed under a separate nomination page due to SEO.

filippoweb3:
@DrW3RK I put everything under "staking apps" ;)

DrW3RK (approved):

PR #5313, opened 2023-10-27
Add Sidebar item - Explore

DrW3RK:
Move all the relevant docs to the "Explore" section and Improve Wiki reader UX

Potential docs that can be moved to Explore section - Community and Contributors, Programmes and everything that makes sense to be in that section.

image


filippoweb3 (approved):
lgtm, I modified the sidebar a bit, and moved the Explore section at the top. I also modified the front page, added Explore card and changed the layout of the cards into 2x2 matrix. I added Explore the the header as well. Let me know what you think!

PR #5320, opened 2023-10-30
Move Multix

filippoweb3:


polkadot-wiki/sidebars.js
```diff
@@ -42,6 +41,7 @@ module.exports = {
           items: [
             "general/staking-apps",
             "general/governance-apps",
+            "general/multisig-apps",
```
profradha, line 44: I thought the consensus in our TechEd chat was that multisigs are grouped under wallets, right?

DrW3RK (approved):

PR #5329, opened 2023-10-31
Getting Started Page - Explore section

DrW3RK:
Having "Getting Started" as starting doc on the Wiki navigational sidebar makes sense. As this doc is exploratory, it is better positioned in "Explore" than "Learn".

image


CrackTheCode016 (approved):
lgtm, i like the way this looks, thank you!

PR #5336, opened 2023-11-02
Add dart lib to available libraries

leonardocustodio:


CrackTheCode016 (approved):
lgtm, ty!

docs/build/build-tools-index.md
```diff
@@ -184,6 +185,7 @@ WebAssembly related tools and projects.
   Substrate, abstracting some complex RPC calls into simple REST calls.
 - [Subxt](https://github.com/paritytech/substrate-subxt) - A Rust library to submit extrinsics to a
   Substrate node via RPC.
+- [Dart/Flutter API](https://github.com/leonardocustodio/polkadart) - Dart API for Polkadot.
```
CrackTheCode016: Perhaps this doesn't need to be in the CLI tools actually, and just in the library, unless it offers some sort of CLI functionality?  

```suggestion
```

CrackTheCode016 (changes requested):

docs/build/build-tools-index.md
leonardocustodio: It is not under CLI though it looks like by looking the difference on GH (the CLI is appearing because of the change just above it). I've put it under Alternative Libs AND rpc libs

CrackTheCode016: OH! my bad, didn't see.. I think just once above is still okay though in Alternative Libraries, if you don't mind.  😁 Sorry for the trouble!

Thank you for your contribution!

leonardocustodio: Sounds good, no problem at all :)

CrackTheCode016 (approved):

PR #5343, opened 2023-11-03
Remove documentation related to Wococo following its shut down

PierreBesson:
As part of the deployment of the Westend <> Rococo bridge, the Wococo testnet (which was one side of the old Wococo <> Rococo test bridge) was shut down. I am removing any Wococo related documentation except for a small explanation that we can keep as reference if anyone is looking for information about Wococo.


filippoweb3 (approved):
lgtm! Thanks for adding this information! :)

PR #5346, opened 2023-11-08
Document secure-validator mode

### LINKED ISSUE #4881: Secure Mode Announcement
## Announcement

Attention validators! In Polkadot v1.5.0 we will be releasing our new **secure-mode** feature. This mode takes advantage of recent security work that has been done to protect validators from malicious PVFs, and **will be enabled by default**. Make sure you meet the **requirements** listed below!

### Why?

While PVF execution is already sandboxed, in reality there can be bugs and exploits and we want validator machines to be as secure as possible. Also, on-demand parachains are coming soon, making it easier than before to get a validator to run untrusted code. Hence we have implemented additional layers of protection, to protect the very sensitive validator keys.

### How does this protect me?

We have implemented very strict filesystem, networking, and process sandboxing, on top of the existing wasmtime sandbox. This will make it very unlikely for a validator machine to be compromised by running untrusted code.

### What are the requirements?

1. **Linux on x86-64**. This is within our existing [recommendations](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-polkadot#requirements), though it was never a hard requirement. According to our [telemetry](https://telemetry.polkadot.io/), most validators should already meet this requirement.
2. **seccomp enabled**. You can check that this is the case by running the following command:

```
cat /boot/config-`uname -r` | grep CONFIG_SECCOMP=
```

The expected output, if enabled, is:

```
CONFIG_SECCOMP=y
```

3. OPTIONAL: **Linux 5.13**. Provides access to even more strict filesystem protections. Many validators are not yet on this version of the kernel, so we will not require it to run secure-mode. However, while not necessary to be reasonably secure, it is highly recommended to upgrade to at least this kernel version.

### What if I don't meet the requirements?

There will be a flag to bypass the secure-mode requirements, but we will highly discourage production validators from using it due to the security risks involved. Most validators should already meet the mandatory requirements (1) and (2), though you may need to upgrade to meet the optional requirement (3).
Note: As mentioned, PVFs are already executed inside wasmtime's sandbox, but by itself this may not be enough protection against a determined attacker.

---

## Related

Implementation issue for secure-mode: https://github.com/paritytech/polkadot-sdk/issues/1444

mrcnski:
This page will be linked to in an error message if a validator fails to meet the requirements.

Let me know if there is too much or too little detail.

cc @eskimor @wpank 

## Related

Closes https://github.com/w3f/polkadot-wiki/issues/4881
Polkadot issue: https://github.com/paritytech/polkadot-sdk/issues/1444

### DIFF
--- a/docs/maintain/maintain-guides-secure-validator.md
+++ b/docs/maintain/maintain-guides-secure-validator.md
@@ -71,13 +71,13 @@ behavior.
 
 ### Secure-Validator Mode
 
-Parity Polkadot has a Secure-Validator Mode which enables several protections for keeping keys
-secure. The protections include highly strict filesystem, networking, and process sandboxing, on top
-of the existing wasmtime sandbox.
+Parity Polkadot has a Secure-Validator Mode, enabling several protections for keeping keys secure.
+The protections include highly strict filesystem, networking, and process sandboxing on top of the
+existing wasmtime sandbox.
 
-This mode is **on by default** if the machine meets the following requirements. If not, there is an
-error message with instructions on disabling Secure-Validator Mode, though this is not recommended
-due to the security risks involved.
+This mode is **activated by default** if the machine meets the following requirements. If not, there
+is an error message with instructions on disabling Secure-Validator Mode, though this is not
+recommended due to the security risks involved.
 
 #### Requirements
 
--- a/docs/maintain/maintain-guides-secure-validator.md
+++ b/docs/maintain/maintain-guides-secure-validator.md
@@ -76,7 +76,8 @@ secure. The protections include highly strict filesystem, networking, and proces
 of the existing wasmtime sandbox.
 
 This mode is **on by default** if the machine meets the following requirements. If not, there is an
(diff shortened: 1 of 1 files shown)


docs/maintain/maintain-guides-secure-validator.md
```diff
@@ -69,6 +69,32 @@ behavior.
...
 :::
 
+### Secure-Validator Mode
+
+Parity Polkadot has a Secure-Validator Mode which enables several protections for keeping keys
+secure. The protections include highly strict filesystem, networking, and process sandboxing, on top
+of the existing wasmtime sandbox.
+
+This mode is **on by default** if the machine meets the following requirements. If not, there is an
+error message with instructions on disabling Secure-Validator Mode.
```
eskimor: Which is not recommended for obvious reasons ... would be a good addition here.

filippoweb3 (approved):
lgtm, thank you! I added a grammar check and a tiny edit

PR #5417, opened 2023-12-09
Update maintain-guides-async-backing.md

SBalaguer:
This comes from an update from the core team working on the Async Backing protocol.


DrW3RK (approved):
Thank you

PR #5441, opened 2023-12-23
Update build-protocol-info.md

muddlebee:
updated info about ED about account deactivation, since we were getting some queries in discord what will happen to the account? whether it will be deleted  or not

### DIFF
--- a/docs/build/build-protocol-info.md
+++ b/docs/build/build-protocol-info.md
@@ -91,11 +91,15 @@ You can always verify the existential deposit by checking the
 [chain state](https://polkadot.js.org/apps/#/chainstate) for the constant
 `balances.existentialDeposit`.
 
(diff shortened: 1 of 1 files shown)


docs/build/build-protocol-info.md
```diff
@@ -91,11 +91,15 @@ You can always verify the existential deposit by checking the
 [chain state](https://polkadot.js.org/apps/#/chainstate) for the constant
 `balances.existentialDeposit`.
 
+If an account drops below the _existential deposit_ ED, the account is reaped ("deactivated"), and any remaining funds are destroyed. The address can be reactivated with a new deposit larger than the existential deposit at any time. This will not restore the destroyed funds.
```
profradha:
```suggestion
If an account drops below the _existential deposit_ ED, the account is reaped ("deactivated"), and any remaining funds are burned. The address can be reactivated anytime by transferring funds greater than the existential deposit. This will not restore the burned funds when the account was reaped. 
```

DrW3RK (approved):
Thank you!

PR #5494, opened 2024-01-10
Edit HRMP opening info

### LINKED ISSUE #5485: Update HRMP documentation
Issue reference https://polkadot.polkassembly.io/referenda/403

filippoweb3:
Removed the info about using the `force_transfer` within the batchAll call, as now the track is General Admin (not Root), which does not have permissions.

Added info about transferring funds deposit (20 DOT + existential deposit + fees) to the system parachain's sovereign account before opening the channel and doing the proposal.

### DIFF
--- a/docs/build/build-hrmp-channels.md
+++ b/docs/build/build-hrmp-channels.md
@@ -21,9 +21,10 @@ the second chain needs to accept and do the same.
 To do this, the calls to be dispatched from both chains are:
 
 1. `hrmp > hrmpInitOpenChannel(recipient, proposedMaxCapacity, proposedMaxMessageSize)`: this call
-   initiates opening a channel from a parachain to a given recipient with a given channel
+   initiates opening a channel from a parachain to a given recipient.
 
-2. `hrmp > hrmpAcceptOpenChannel(sender)`: accepting the channel open request from the given sender.
+2. `hrmp > hrmpAcceptOpenChannel(sender)`: this call is for accepting the channel open request from
+   the given sender.
 
 Each parachain is to use its governance process to execute this. The call will be executed on the
 Relay Chain.
--- a/docs/build/build-hrmp-channels.md
+++ b/docs/build/build-hrmp-channels.md
@@ -7,9 +7,9 @@ keywords: [HRMP, parachain, statemint, proposal, asset hub, statemine]
 slug: ../build-hrmp-channels
 ---
 
-[HRMP](../learn/learn-xcm.md#xcmp-lite-hrmp) has the same interface and functionality as
-[XCMP](../learn/learn-xcm.md#xcmp-cross-chain-message-passing) but is much more demanding on
-resources since it passes all messages via the Relay Chain. When XCMP is implemented on Polkadot,
+[HRMP](../learn/learn-xcm-transport.md#hrmp-xcmp-lite) has the same interface and functionality as
+[XCMP](../learn/learn-xcm-transport.md#xcmp-cross-chain-message-passing) but is much more demanding
+on resources since it passes all messages via the Relay Chain. When XCMP is implemented on Polkadot,
 HRMP is planned to be deprecated and phased out in favor of it.
 
 ## Opening HRMP channel: parachain to parachain
@@ -40,9 +40,7 @@ then submitting the proposal on-chain.
 Before submitting the proposal, both the system parachain's and the parachain's sovereign accounts
 (both [system accounts](../learn/learn-account-advanced.md#system-accounts)) must have the channel
 deposit of 20 DOT plus the
-[existential deposit](../learn/learn-accounts.md#existential-deposit-and-reaping). At the time of
-execution, each parachain's sovereign account must have a free balance greater than or equal to the
-sum of the sender deposit, recipient deposit, and existential deposit.
+[existential deposit](../learn/learn-accounts.md#existential-deposit-and-reaping).
 
 :::info
 
--- a/docs/build/build-hrmp-channels.md
+++ b/docs/build/build-hrmp-channels.md
@@ -14,19 +14,19 @@ HRMP is planned to be deprecated and phased out in favor of it.
 
 ## Opening HRMP channel: parachain to parachain
 
-To open a channel from one parachain to another that are not system chains on Polkadot, the
-governance of each parachain needs to declare its intent to open a channel on the Relay Chain, and
+To open a channel from one parachain to another, that is not system chains on Polkadot, the
+governance of each parachain needs to declare its intent to open a channel on the Relay Chain and
 the second chain needs to accept and do the same.
 
-In order to do this, the calls to be dispatched from both chains are:
+To do this, the calls to be dispatched from both chains are:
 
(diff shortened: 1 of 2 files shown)


filippoweb3:
Hey @lrazovic I updated the info about opening HRMP channels. Let me know if there is additional info to add.

docs/build/build-hrmp-channels.md
```diff
@@ -30,17 +30,26 @@ the Relay Chain.
...
+[General Admin Track](../learn/learn-polkadot-opengov-origins.md#general-admin). Like all other
+governance proposals, proposers should follow best practices like opening a discussion on
 [Polkassembly](https://polkadot.polkassembly.io/) or [Subsquare](https://polkadot.subsquare.io/) and
 then submitting the proposal on-chain.
 
+Before submitting the proposal, both the system parachain's and the parachain's sovereign accounts
+(both [system accounts](../learn/learn-account-advanced.md#system-accounts)) must have the channel
+deposit of 20 DOT plus the
+[existential deposit](../learn/learn-accounts.md#existential-deposit-and-reaping) and some extra
+funds to cover for fees. Anyone can send those funds via `balances.transferKeepAlive` call to the
```
joepetrowski: This is not true. And what fees? There are no fees because there is no transaction.

docs/build/build-hrmp-channels.md
```diff
@@ -30,17 +30,26 @@ the Relay Chain.
...
+system parachain's sovereign account (if funds are not already available). System parachains have ID
```
joepetrowski: Why the complexity? This can simply say, "at the time of execution, each parachain's sovereign account must have a free balance greater than or equal to the sum of the sender deposit, recipient deposit, and existential deposit."

docs/build/build-hrmp-channels.md
```diff
@@ -67,7 +76,6 @@ the community to review and ultimately vote on:
 3.  Technical details of the proposal, including proposal parameters and technical details of this
     call (On Kusama, most proposals were designed as a batchAll calls):
 
-    - A `forceTransfer` of 20 DOT from the Treasury to AssetHub sovereign account.
```
joepetrowski, line 70: Well now the first sentence below is not true...

joepetrowski (changes requested):
Also note that after runtime 1,001,000, chains will not need any deposit at all for channels with system chains (and should be fine without any balance in their accounts) and there will be a `poke_deposit` function that will refund any existing deposits.

docs/build/build-hrmp-channels.md
joepetrowski: Actually not just the first sentence, basically everything is wrong.

docs/build/build-hrmp-channels.md
filippoweb3: implemented

docs/build/build-hrmp-channels.md
filippoweb3: removed

filippoweb3:
@DrW3RK this PR is good to go, I fixed some old links and updated the info using the documentation provided by Polimec. @lrazovic provided some feedback I implemented.

docs/build/build-hrmp-channels.md
```diff
@@ -7,40 +7,56 @@ keywords: [HRMP, parachain, statemint, proposal, asset hub, statemine]
...
+[HRMP](../learn/learn-xcm-transport.md#hrmp-xcmp-lite) has the same interface and functionality as
+[XCMP](../learn/learn-xcm-transport.md#xcmp-cross-chain-message-passing) but is much more demanding
+on resources since it passes all messages via the Relay Chain. When XCMP is implemented on Polkadot,
 HRMP is planned to be deprecated and phased out in favor of it.
 
 ## Opening HRMP channel: parachain to parachain
 
-To open a channel from one parachain to another that are not system chains on Polkadot, the
-governance of each parachain needs to declare its intent to open a channel on the Relay Chain, and
+To open a channel from one parachain to another, that is not system chains on Polkadot, the
```
JuaniRios:
```suggestion
To open a channel from one parachain to another, that is not a system chain on Polkadot, the
```

docs/build/build-hrmp-channels.md
```diff
@@ -7,40 +7,56 @@ keywords: [HRMP, parachain, statemint, proposal, asset hub, statemine]
...
+governance of each parachain needs to declare its intent to open a channel on the Relay Chain and
```
JuaniRios:
```suggestion
governance of each parachain needs to declare its intent to open a channel on the Relay Chain through a `Transact` XCM message with parachain origin, and
```

docs/build/build-hrmp-channels.md
```diff
@@ -7,40 +7,56 @@ keywords: [HRMP, parachain, statemint, proposal, asset hub, statemine]
...
 the second chain needs to accept and do the same.
 
-In order to do this, the calls to be dispatched from both chains are:
+To do this, the calls to be dispatched from both chains are:
```
JuaniRios:
```suggestion
To do this, the calls to be included in the `Transact` message from both chains are:
```

docs/build/build-hrmp-channels.md
JuaniRios: From my understanding, this is done through a transact XCM, so it should be clearly communicated

docs/build/build-hrmp-channels.md
JuaniRios: Then we could go into more detail about the transact

docs/build/build-hrmp-channels.md
```diff
@@ -7,40 +7,56 @@ keywords: [HRMP, parachain, statemint, proposal, asset hub, statemine]
...
+   initiates opening a channel from a parachain to a given recipient.
 
-2. `hrmp > hrmpAcceptOpenChannel(sender)`: accepting the channel open request from the given sender.
+2. `hrmp > hrmpAcceptOpenChannel(sender)`: this call is for accepting the channel open request from
+   the given sender.
 
-Each parachain is to use its own governance process to execute this. The call will be executed on
-the Relay Chain.
+Each parachain is to use its governance process to execute this. The call will be executed on the
+Relay Chain.
```
JuaniRios, line 30: If it is to be made only through governance, then why do we have the xcm instruction [HrmpNewChannelOpenRequest](https://paritytech.github.io/xcm-docs/journey/channels-and-bridges.html#hrmpnewchannelopenrequest) sent by the relay as a notification?
At Polimec we are using this to do automatic channel opening, without going through governance.

JuaniRios (commented):
If this would have been in the docs, it would have saved me a lot of time delving into the code. I believe these changes will also benefit future devs.

docs/build/build-hrmp-channels.md
JuaniRios: Although we had to modify the xcm-executor because it had that instruction left empty, so maybe that change can happen once the executor adds an implementation for those instructions 🤔

joepetrowski:
https://github.com/w3f/polkadot-wiki/pull/5498

PR #5530, opened 2024-01-23
Clarify Multi-sig call deposits

DrW3RK:
Remove ambiguity about deposit being held for signatory accounts and explain the process in detail.


docs/learn/learn-guides-accounts-multisig.md
```diff
@@ -57,12 +57,14 @@ Parity Technologies. There is a detailed
 that you can try out and change to see how it works.
 
 However, in anything but the simple one approval case, you will likely need more than one of the
-signatories to approve the call before finally executing it. When you create a new call or approve a
-call as a multisig, you will need to place a small deposit. The deposit stays locked in the pallet
```
profradha, line 61: From what I know, approving a multi-sig call does not require a deposit. Modifying the text to reflect that.

filippoweb3: yeah that seems right to me, the one who initiates places the deposit, which is held until:
- the multisig tx is dispatched
- the initial caller cancels it

filippoweb3 (approved):
lgtm

PR #5531, opened 2024-01-23
Add info transactions

filippoweb3:


docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
-is refunded. There is no need for the attacker to know your private key, valid signatures for those
-past transactions and nonces already exist and are stored on-chain (meaning the private key was
-already used to generate those signatures).
-
-Making a transaction mortal will almost certainly ensure that replay attacks are not possible, with
-the only exception being if the account is reaped and then re-funded shortly after submitting a
-mortal transaction, and then an attacker replays that transaction within the mortality window (i.e.,
-the specified block interval).
+Transactions are usually irreversible once confirmed and added to the blockchain, an immutable
+ledger of all transactions. This feature ensures network integrity by preventing
```
laboon: Double-spending is a very specific case; I think we can just remove this sentence

docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
-nonce before the reaping event were immortal, all past transactions can be replayed once the account
-is refunded. There is no need for the attacker to know your private key, valid signatures for those
-past transactions and nonces already exist and are stored on-chain (meaning the private key was
-already used to generate those signatures).
-
-Making a transaction mortal will almost certainly ensure that replay attacks are not possible, with
-the only exception being if the account is reaped and then re-funded shortly after submitting a
-mortal transaction, and then an attacker replays that transaction within the mortality window (i.e.,
-the specified block interval).
+Transactions are usually irreversible once confirmed and added to the blockchain, an immutable
```
laboon: I would say "generally" instead of "usually". In almost all cases, they will be irreversible on Polkadot, and almost certainly once finalized (barring some very unusual stuff which we don't need to go into detail on here).

docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
+ledger of all transactions. This feature ensures network integrity by preventing
+[double-spending](https://en.wikipedia.org/wiki/Double-spending). However, it also means **users
+must exercise caution, as sending funds to the wrong address or wrong amount cannot be rectified
```
laboon: I would change this sentence to "... as mistakes such as sending DOT to the wrong address cannot be reverted."

docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
+unless the recipient cooperates.** The permanence of transactions highlights the importance of
+careful verification before sending funds on a blockchain network. It is usually a
```
laboon:
```suggestion
careful verification before signing any transaction on a blockchain network. It is usually a
```

docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
+[good practice not to blind sign transactions](../general/transaction-attacks.md) to avoid being
+victim of an attack.
+
+In blockchain terms, irreversible transactions are called **mortal** extrinsics (i.e. valid within a
```
laboon: irreversible is not the same as mortal - mortal extrinsics just mean that the extrinsic itself expires after a certain number of blocks

docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
+defined block interval, usually short). However, extrinsics can also be **immortal** (i.e. always
+valid). By default and [for security reasons](../general/transaction-attacks.md#replay-attack), all
+extrinsics on {{ polkadot: Polkadot :polkadot }}{{ kusama: Kusama :kusama }} will be mortal, but
```
laboon: It is possible to make immortal transactions on both Polkadot and Kusama, unless something has changed recently and I missed it. However, it is HIGHLY recommended not to do so and most wallet software will not allow you to make an immortal extrinsic.

docs/learn/learn-transactions.md
```diff
@@ -77,23 +77,20 @@ Here are some key differences between the different types of extrinsics:
...
+always checking before signing is a good practice.
```
laboon: I don't think we should add this unless we tell users _how_ to check if their transaction is mortal or immortal. This will differ with every piece of wallet software, so I'd just remove it.

docs/learn/learn-transactions.md
laboon: So we should remove the "By default" (since the network itself doesn't do anything by default, it will be the software (i.e. wallets) that interact with it) and explain the above

docs/general/transaction-attacks.md
```diff
@@ -60,6 +60,22 @@ This will be showed in the Signer app and a careful user will notice it. If the
 Signer is already incorrect (or the Signer is corrupted) there is the risk of signing a non-intended
 transaction without the possibility of verifying it.
 
+### Replay Attack
+
+A replay attack is where past transactions can be replayed (same [balance](#balance-transfers),
+receiver account, etc.) without knowing private keys. This could happen in the context of
```
laboon: I would say that that it's an attack where publicly known data can be "replayed" as a new extrinsic.

docs/general/transaction-attacks.md
```diff
@@ -60,6 +60,22 @@ This will be showed in the Signer app and a careful user will notice it. If the
...
+[reaping accounts](./learn-accounts.md#existential-deposit-and-reaping) because the reaping process
+resets the nonce value. If all signed transactions until the nonce before the reaping event were
+immortal, all past transactions can be replayed once the account is refunded. There is no need for
+the attacker to know your private key, valid signatures for those past transactions and nonces
+already exist and are stored on-chain (meaning the private key was already used to generate those
+signatures).
+
+Making a [transaction mortal](../learn/learn-transactions.md#mortal-and-immortal-extrinsics) will
```
laboon: Add "... with a relatively short mortality window"

docs/general/transaction-attacks.md
```diff
@@ -60,6 +60,22 @@ This will be showed in the Signer app and a careful user will notice it. If the
...
+almost certainly ensure that replay attacks are not possible, with the only exception being if the
+account is reaped and then re-funded shortly after submitting a mortal transaction, and then an
+attacker replays that transaction within the mortality window (i.e., the specified block interval).
```
laboon: ... AND that transaction which is valid (e.g. if you only have 50 DOT, and you try to issue an extrinsic sending 100 DOT, the transfer will fail)

laboon (changes requested):

PR #5533, opened 2024-01-23
Update learn-identity.md

alineherzman:
It's the website in the on-chain identity. The one that is currently linked on the page doesn't work.


DrW3RK:
Works fine on my end. https://registrar.d11d.net/ 

What is the issue at your end?

alineherzman:
It works for me now, too. I'm not sure what was wrong half an hour ago, but thanks for checking :)

PR #5537, opened 2024-01-25
update docs on cosmos section

adiraviraj:
this PR updates the [Learn section](https://wiki.polkadot.network/docs/learn-index) on Cosmos to reflect up to date information. some of the text in the current documentation on Cosmos and IBC is not factually correct. this PR corrects that.

### DIFF
--- a/docs/learn/learn-comparisons-cosmos.md
+++ b/docs/learn/learn-comparisons-cosmos.md
@@ -55,18 +55,9 @@ Polkadot has [bridge parachains](learn-bridges.md) that offer two-way compatibil
 
 ### Cosmos
 
-Cosmos has a main chain called a "Hub" that connects other blockchains called "zones". Cosmos can
-have multiple hubs, but this overview will consider a single hub. Each zone must maintain its own
-state and therefore have its own validator community. When a zone wants to communicate with another
-zone, it sends packets over IBC. The Hub maintains a multi-token ledger of token balances
-(non-transfer messages are relayed but their state not stored in the Hub).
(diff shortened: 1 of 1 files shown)


DrW3RK:
Thanks for making the update. Could you be more specific on what you deem to be factually incorrect? 

We certainly do not want to host incorrect info and happy to revamp this document to reflect the latest technical implementation of Cosmos.

adiraviraj:
hi @DrW3RK, thank you for the message. in the current docs it says:

> "Zones monitor the state of the Hub with a light client, but the Hub does not track zone states."

 this would be incorrect as all zones/chains that have an IBC Connection between them are tracking each other's states.

> "Zones must use a deterministic finality algorithm (currently, all use Tendermint)"

Implementing IBC does not require deterministic finality. A relayer can wait for finality on source before relaying a packet to the counterpary.

In general, the docs imply a hub-and-spoke model with Cosmos Hub at the centre, and other chains connecting to it and routing messages vie the Hub. this is not accurate as there is no "main chain" within the Cosmos ecosystem. currently there is a 1:1 connection between IBC chains. for a visualization of this topography, see [mapofzones.com](https://mapofzones.com/home?columnKey=ibcVolume&period=24h)

Finally, the concept of "peg zones" were proposed back in 2017-18 and is not relevant anymore. Today we see projects leverage ZK solutions to connect Cosmos with other ecosystems like Ethereum and its L2s (teams such as [Polymer](https://www.polymerlabs.org/), [Union](https://union.build/), and [Electron](https://www.zkibc.com/) are working on this).

adiraviraj:
i just now noticed that there are other sections on IBC within the docs that are incorrect as well. can open another PR later to correct those.

DrW3RK (approved):

PR #5499, opened 2024-01-11
Update Fellowship Membership instructions

DrW3RK:
- [x] Create a new wiki page for the Polkadot Technical Fellowship
- [x] Refine the content to reflect the latest changes made (changes to adding and promoting members)

Information about evidence submission, demotion period etc. will be added through a new PR as this one is already significant reading/review work.

PR #5540, opened 2024-01-26
Add further details - Fellowship Management

DrW3RK:
As a followup to https://github.com/w3f/polkadot-wiki/pull/5499 

- [ ] Add instructions for the significant core-fellowship pallet calls
- [ ] Elaborate on fellowship membership management


DrW3RK:
Found https://github.com/polkadot-fellows/dashboard repo, which would host Fellowship info. When the info is live, we can link to it to avoid duplication of efforts on the Wiki.

PR #5556, opened 2024-01-30
Add info account balances

filippoweb3:


michalisFr (changes requested):
The transferable is calculated in the new way in balances pallet:
`transferable = free - max(frozen - reserved, ED)`

Other than that it looks good

PR #5563, opened 2024-01-31
Fellowship Induction does not require a proposal

DrW3RK:
Examples https://collectives.statescan.io/#/extrinsics/3081647-2

https://collectives.statescan.io/#/extrinsics/3087341-2


filippoweb3 (approved):
Maybe explicitly add that the induction does not require a fellowship referendum, what you think?

PR #5585, opened 2024-02-07
Add RFC Approval Flow: Write-up -> Final Implementation

filippoweb3:


DrW3RK (commented):
As the page is titled "Building on Polkadot" which is different from "Building Polkadot", I have moved the Polkadot SDK section to the front.

I think I missed this during my earlier review of this page - RFC stands for Request for Comment, not Commit. Made the necessary changes through out the doc. Moved this section after Polkadot SDK, to show that the protocol is flexible to change itself if required.

Also refined the instructions originally added by @filippoweb3 in this PR.

Please review the newly added content and let me know if anything needs to be updated.

DrW3RK (approved):

DrW3RK:
Also added a note that although the Fellowship maintains the runtime for Polkadot, they do not gate it completely. Anyone can set the runtime of Polkadot through an OpenGov referendum on root track.

filippoweb3:
@DrW3RK, thanks for the review and corrections. I added a couple of links to pages and some minor edits.

PR #5591, opened 2024-02-08
Elastic Scaling

filippoweb3:
Add information about elastic scaling: how parachains will be able to reserve time on multiple cores and thus include multiple parablocks in a relay chain block at the same time, while producing parablocks < 6s 

Some info:
- https://github.com/paritytech/polkadot-sdk/issues/1829
- https://polkadot.network/blog/elevating-polkadots-performance-and-scale-with-asynchronous-backing

### DIFF
--- a/docs/learn/learn-elastic-scaling.md
+++ b/docs/learn/learn-elastic-scaling.md
@@ -22,6 +22,10 @@ This upgrade allows parachains to utilize up to 2 seconds execution time per par
 With elastic scaling, parachains can use multiple cores to include multiple parablocks
 within the same relay chain block.
 
+The relay chain receives a sequence of parachain blocks on multiple cores, which are validated and checked if all their state roots line up during their inclusion, but assume they’re unrelated parachain blocks during backing, availability, and approvals. With elastic scaling implemented, a parachain's throughput depends upon its collator infrastructure.
+
+The [elastic scaling implementation](https://github.com/paritytech/polkadot-sdk/issues/1829) will be rolled out in multiple phases. In the first phase, elastic scaling is set to work on parachains with a trusted/permissioned collator set. With this restriction, it is possible to launch elastic scaling without changing the candidate receipt. After successfully implementing the first phase, changes can be made to the candidate receipt so the collator set can be untrusted/permissionless again. The final phase will feature full integration with the Cumulus framework, enabling parachains to be configured to access multiple cores continuously.
+
 
 Take, for example, a parachain that wants to submit four parablocks to the relay chain. Without
 elastic scaling, it will take 24 seconds to include all of them through one core. Remember that a
--- a/docs/learn/learn-elastic-scaling.md
+++ b/docs/learn/learn-elastic-scaling.md
@@ -8,10 +8,12 @@ slug: ../learn-elastic-scaling
 ---
 
 The path of parablocks from their creation to their inclusion into the relay chain (discussed in the
-[Parachain Protocol Page](./learn-parachains-protocol.md)) span two domains: the parachain's and
-relay chain's roles. Scaling the Polkadot protocol involves consideration of how parablocks are
-validated, processed, secured, made available for additional checks, and finally included in the
-relay chain as part of its final state
+[Parachain Protocol Page](./learn-parachains-protocol.md)) spans two domains: the parachain's and
+relay chain's. Scaling the Polkadot protocol involves consideration of how parablocks are produced by the
+parachain and then
+validated, processed, secured, made available for additional checks, and finally included on the
+relay chain.
+
 
 [Asynchronous backing](./learn-async-backing.md) is the first implementation of the parachain
 protocol that alters how parachains produce blocks and how the relay chain processes their data.
--- a/docs/learn/learn-elastic-scaling.md
+++ b/docs/learn/learn-elastic-scaling.md
@@ -9,23 +9,32 @@ slug: ../learn-elastic-scaling
 
 The path of parablocks from their creation to their inclusion into the relay chain (discussed in the
 [Parachain Protocol Page](./learn-parachains-protocol.md)) spans two domains: the parachain's and
-relay chain's. Scaling the Polkadot protocol involves consideration of how parablocks are produced by the
-parachain and then
-validated, processed, secured, made available for additional checks, and finally included on the
-relay chain.
-
+relay chain's. Scaling the Polkadot protocol involves consideration of how parablocks are produced
+by the parachain and then validated, processed, secured, made available for additional checks, and
+finally included on the relay chain.
 
 [Asynchronous backing](./learn-async-backing.md) is the optimization implemented on the relay chain
-that allows parachains to produce blocks faster and allows relay chain to process them seamlessly. Async backing also improves the parachain side with unincluded segments and augmented info that allows collators to produce multiple parablocks even if the previous blocks are not yet included.
-This upgrade allows parachains to utilize up to 2 seconds execution time per parablock, and the relay chain will be able to include a parablock every 6 seconds.
+that allows parachains to produce blocks faster and allows relay chain to process them seamlessly.
+Asynchronous backing also improves the parachain side with unincluded segments and augmented info
+that allows collators to produce multiple parablocks even if the previous blocks are not yet
+included. This upgrade allows parachains to utilize up to 2 seconds execution time per parablock,
+and the relay chain will be able to include a parablock every 6 seconds.
 
-With elastic scaling, parachains can use multiple cores to include multiple parablocks
-within the same relay chain block.
+With elastic scaling, parachains can use multiple cores to include multiple parablocks within the
+same relay chain block.
 
-The relay chain receives a sequence of parachain blocks on multiple cores, which are validated and checked if all their state roots line up during their inclusion, but assume they’re unrelated parachain blocks during backing, availability, and approvals. With elastic scaling implemented, a parachain's throughput depends upon its collator infrastructure.
+The relay chain receives a sequence of parachain blocks on multiple cores, which are validated and
+checked if all their state roots line up during their inclusion, but assume they’re unrelated
+parachain blocks during backing, availability, and approvals. With elastic scaling implemented, a
+parachain's throughput depends upon its collator infrastructure.
 
-The [elastic scaling implementation](https://github.com/paritytech/polkadot-sdk/issues/1829) will be rolled out in multiple phases. In the first phase, elastic scaling is set to work on parachains with a trusted/permissioned collator set. With this restriction, it is possible to launch elastic scaling without changing the candidate receipt. After successfully implementing the first phase, changes can be made to the candidate receipt so the collator set can be untrusted/permissionless again. The final phase will feature full integration with the Cumulus framework, enabling parachains to be configured to access multiple cores continuously.
-
+The [elastic scaling implementation](https://github.com/paritytech/polkadot-sdk/issues/1829) will be
+rolled out in multiple phases. In the first phase, elastic scaling is set to work on parachains with
+a trusted/permissioned collator set. With this restriction, it is possible to launch elastic scaling
+without changing the candidate receipt. After successfully implementing the first phase, changes can
+be made to the candidate receipt so the collator set can be untrusted/permissionless again. The
+final phase will feature full integration with the Cumulus framework, enabling parachains to be
+configured to access multiple cores continuously.
 
 Take, for example, a parachain that wants to submit four parablocks to the relay chain. Without
 elastic scaling, it will take 24 seconds to include all of them through one core. Remember that a
--- a/docs/learn/learn-elastic-scaling.md
+++ b/docs/learn/learn-elastic-scaling.md
@@ -19,12 +19,9 @@ relay chain.
 that allows parachains to produce blocks faster and allows relay chain to process them seamlessly. Async backing also improves the parachain side with unincluded segments and augmented info that allows collators to produce multiple parablocks even if the previous blocks are not yet included.
 This upgrade allows parachains to utilize up to 2 seconds execution time per parablock, and the relay chain will be able to include a parablock every 6 seconds.
 
-With asynchronous backing enabled, parachains can theoretically produce parablocks in the sub-second
-time range. However, to accomplish this, asynchronous backing must temporarily **hold** blocks in
-[unincluded segments](./learn-async-backing.md#unincluded-segments), while parablocks can be
-included into the relay chain every 6 seconds per core. This is the bottleneck that elastic scaling
-addresses. With elastic scaling, parachains can use multiple cores to include multiple parablocks
+With elastic scaling, parachains can use multiple cores to include multiple parablocks
 within the same relay chain block.
+
 
 Take, for example, a parachain that wants to submit four parablocks to the relay chain. Without
 elastic scaling, it will take 24 seconds to include all of them through one core. Remember that a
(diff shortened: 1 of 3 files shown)


filippoweb3:
@CrackTheCode016, asking for a preliminary review of the concept and broad mechanics ;)

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,71 @@
...
+description: Pushing Parachains' Boundaries Below 6 seconds Execution Time.
+keywords: [elastic scaling, parachains, coretime, blockspace]
+slug: ../learn-elastic-scaling
+---
+
+The path of parablocks from their creation to their inclusion into the relay chain (discussed in the
+[Parachain Protocol Page](./learn-parachains-protocol.md)) span two domains: the parachain side and
+the relay chain side. Scaling the Polkadot technology implies scaling how parablocks are produced
+and how their data is processed, secured, made available for additional checks, and finally included
+in the relay chain.
```
CrackTheCode016:
```suggestion
[Parachain Protocol Page](./learn-parachains-protocol.md)) span two domains: the parachain's role and
the relay chain's role. Scaling the Polkadot protocol involves consideration on how parablocks are validated, processed, secured, made available for additional checks, and finally included
in the relay chain as part of its final state
```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,71 @@
...
+
+[Asynchronous backing](./learn-async-backing.md) is the first implementation to the parachain
+protocol that alters how parachains produce blocks and how the relay chain processes their data.
+Briefly, parachains will take up to 2 seconds execution time (produce parablocks), while the relay
```
CrackTheCode016:
```suggestion
This upgrade will allow parachains to take up to 2 seconds execution time (for producing parablocks), while the relay
```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,71 @@
...
+chain will be able to include a parablock every 6 seconds.
+
+The limiting factor now is the relay chain and the 6 seconds it takes backing groups to back and
+include parablocks. After asynchronous backing is implemented, parachains can theoretically produce
+parablocks in the sub-second time range, but this would just create large
+[unincluded segments](./learn-async-backing.md#unincluded-segments) on their end and a bottleneck on
+the relay chain. Here is where elastic scaling comes into play. With elastic scaling, parachains can
+use multiple cores to include multiple parablocks within the same relay chain block. Take for
+example a parachain that wants to submit 4 parablocks to the relay chain:
```
CrackTheCode016:
```suggestion
With asynchronous backing enabled, parachains can theoretically produce
parablocks in the sub-second time range. However, to accomplish this, asynchronous backing must temporarily **hold** blocks in [unincluded segments](./learn-async-backing.md#unincluded-segments).  This is the bottleneck that elastic scaling addresses.  With elastic scaling, parachains can
use multiple cores to include multiple parablocks within the same relay chain block. Take, for
example, a parachain that wants to submit 4 parablocks to the relay chain:
```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,71 @@
...
+C2    |P3     B          I
+      |P4                B         I
+
+```
+
+If the pace per core on the relay chain will not change (backing and inclusion every 6 seconds per
+core), on the parachain side collators will need to increase the parablock production rate to push
+P1 and P2 to the two relay chain cores.
+
+## Challenges
```
CrackTheCode016: I'm not sure if we should include a "challenges" section in the Wiki - although for ourselves it would be nice to have for discussion purposes, or even presentations.

I know there are a few more challenges which I gleaned from skimming over the issues for elastic scaling, for example, of tracking the order of parablock validation over different cores (and their various states)

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,71 @@
...
+
+The limiting factor now is the relay chain and the 6 seconds it takes backing groups to back and
+include parablocks. After asynchronous backing is implemented, parachains can theoretically produce
+parablocks in the sub-second time range, but this would just create large
+[unincluded segments](./learn-async-backing.md#unincluded-segments) on their end and a bottleneck on
+the relay chain. Here is where elastic scaling comes into play. With elastic scaling, parachains can
+use multiple cores to include multiple parablocks within the same relay chain block. Take for
+example a parachain that wants to submit 4 parablocks to the relay chain:
+
+Without elastic scaling, it will take 35 s to include all of them through one core. Remember that a
```
CrackTheCode016: I could be missing something here, but wouldn't it be 24 seconds, if each block from backing to inclusion is 6 seconds?

CrackTheCode016 (changes requested):
Left a few comments and suggestions, overall looking good. Some thoughts: 

- Since this isn't implemented yet, I would suggest we move this to **Future Implementations**

- I am not sure we should have the challenges section, although I am open to keeping it if it facilitates understanding of the topic.

docs/learn/learn-elastic-scaling.md
filippoweb3: removed

docs/learn/learn-elastic-scaling.md
filippoweb3: yeah its 30 s becuase before p1 there was not other parablock. So there is an additional 6 s. Specified in the text

filippoweb3:
@CrackTheCode016 I addressed your feedback. Let me know if it is ok with you ;)

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+
+With asynchronous backing enabled, parachains can theoretically produce parablocks in the sub-second
+time range. However, to accomplish this, asynchronous backing must temporarily **hold** blocks in
+[unincluded segments](./learn-async-backing.md#unincluded-segments), while parablocks can be
+included into the relay chain every 6 seconds per core. This is the bottleneck that elastic scaling
+addresses. With elastic scaling, parachains can use multiple cores to include multiple parablocks
+within the same relay chain block.
+
+Take, for example, a parachain that wants to submit four parablocks to the relay chain. Without
+elastic scaling, it will take 24 seconds to include all of them through one core. Remember that a
```
CrackTheCode016, line 40: > it will take 24 seconds

24 seconds or 30?

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+C1    |P1     B          I
+      |P2                B         I
+      |P3                          B         I
+      |P4                                    B         I
+
+```
+
+The diagram above shows how the backing and inclusion of parablocks (P) happen within the same relay
+chain block (R). With one core (C1), a parablock is included every 6 seconds. Note how P4 is
+included after 30 seconds (not 24 seconds) because when P1 was pushed to the relay chain for being
```
CrackTheCode016, line 56: I still don't understand why it's 30s and not 24s 🙈 unless you are saying that this is _starting_ at 6s already for P1?

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+backed, there was no previous parablock.
+
+With elastic scaling, it will take just 12 seconds (3-second block time) to include all four
+parablocks using two cores.
+
+```
+              R1 <----- R2 <----- R3
```
CrackTheCode016, line 63: This is due to a lack of knowledge on my part - but why would P3 be backed and included before P1?  Although, I know we may not have all the details of how this works yet

We would need to elaborate on ordering and how that works on a basic level.

CrackTheCode016 (commented):
I would need to read more on core regions and how those will play a part, as I understand, that is the basis for how elastic scaling will take place ([see here](https://forum.polkadot.network/t/unifying-bulk-and-mid-term-blockspace-with-strided-regions/2228#elastic-scaling-4)),

I think that detail should be added before it is merged.   I'd like to dive a little deeper also into the technical implementations, and really ensure this is accurate before approving.
 
As a sidenote, having a dedicated page for coretime will accompany this page would be beneficial, as the two concepts go hand-in-hand.  It would even be good to include a section on how the two relate, to clear up any confusion!

docs/learn/learn-elastic-scaling.md
filippoweb3: the chain starts at R0, at R1 is 6s, and R4 is 30 s

docs/learn/learn-elastic-scaling.md
filippoweb3: yeah P1 is backed after 6s in R1

filippoweb3:
@DrW3RK @CrackTheCode016 let me know if this PR is good to go. On a high level is informative and I think it covers what elastic scaling will bring to Polkadot. In a second PR I could make better figures.

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+
+[Asynchronous backing](./learn-async-backing.md) is the first implementation of the parachain
+protocol that alters how parachains produce blocks and how the relay chain processes their data.
+This upgrade will allow parachains to take up to 2 seconds execution time (for producing
+parablocks), while the relay chain will be able to include a parablock every 6 seconds.
+
+With asynchronous backing enabled, parachains can theoretically produce parablocks in the sub-second
+time range. However, to accomplish this, asynchronous backing must temporarily **hold** blocks in
+[unincluded segments](./learn-async-backing.md#unincluded-segments), while parablocks can be
+included into the relay chain every 6 seconds per core. This is the bottleneck that elastic scaling
```
CrackTheCode016: Just a small note (and imho not really a deal breaker for this PR), but we don't really introduce the notion of core before this. Not a big deal though.

Might be worth mentioning something like, "the slot the parachain occuipies currently represents **one** core".

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+
+```
+
+The diagram above shows how the backing and inclusion of parablocks (P) happen within the same relay
+chain block (R). With one core (C1), a parablock is included every 6 seconds. Note how P4 is
+included after 30 seconds (not 24 seconds) because when P1 was pushed to the relay chain for being
+backed, there was no previous parablock.
+
+With elastic scaling, it will take just 12 seconds (3-second block time) to include all four
+parablocks using two cores.
```
CrackTheCode016, line 60: Of course, if elastic scaling improves time in this manner, than naturally they would also have to pay more for this throughput ;)

CrackTheCode016 (approved):
Although I myself am not terrible familiar with the specifics of elastic scaling, it appears sensible to me and we can update it in the future.

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
+---
+id: learn-elastic-scaling
+title: Polkadot's Elastic Scaling
+sidebar_label: Elastic Scaling
+description: Pushing Parachains' Boundaries Below 6 seconds Execution Time.
```
profradha:
```suggestion
description: Enabling parachains to scale on-demand through instantaneous coretime.
```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+keywords: [elastic scaling, parachains, coretime, blockspace]
+slug: ../learn-elastic-scaling
+---
+
+The path of parablocks from their creation to their inclusion into the relay chain (discussed in the
+[Parachain Protocol Page](./learn-parachains-protocol.md)) span two domains: the parachain's and
+relay chain's roles. Scaling the Polkadot protocol involves consideration of how parablocks are
+validated, processed, secured, made available for additional checks, and finally included in the
+relay chain as part of its final state
```
profradha: The parablock info is not part of the final state of the relay chain but, rather, part of the blockchain. This lecture slide deck can help understand what info is stored on the blockchain and what info is stored on state https://polkadot-blockchain-academy.github.io/pba-book/blockchain-contracts/structure/page.html#where-do-the-states-live

```suggestion
The path of parablocks from their creation to their inclusion into the relay chain (discussed in the
[Parachain Protocol Page](./learn-parachains-protocol.md)) spans two domains: the parachain's and
relay chain's. Scaling the Polkadot protocol involves consideration of how parablocks are produced by the
parachain and then
validated, processed, secured, made available for additional checks, and finally included on the
relay chain.

```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+
+[Asynchronous backing](./learn-async-backing.md) is the first implementation of the parachain
+protocol that alters how parachains produce blocks and how the relay chain processes their data.
+This upgrade will allow parachains to take up to 2 seconds execution time (for producing
+parablocks), while the relay chain will be able to include a parablock every 6 seconds.
```
profradha: As we mentioned the two domains: parachain and relaychain in the opening paragraph, it is apt to mention that async backing is an optimization on the relaychain side. 

```suggestion
[Asynchronous backing](./learn-async-backing.md) is the optimization implemented on the relay chain
that allows parachains to produce blocks faster and allows relay chain to process them seamlessly. Async backing also improves the parachain side with unincluded segments and augmented info that allows collators to produce multiple parablocks even if the previous blocks are not yet included.
This upgrade allows parachains to utilize up to 2 seconds execution time per parablock, and the relay chain will be able to include a parablock every 6 seconds.
```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+
+With asynchronous backing enabled, parachains can theoretically produce parablocks in the sub-second
+time range. However, to accomplish this, asynchronous backing must temporarily **hold** blocks in
+[unincluded segments](./learn-async-backing.md#unincluded-segments), while parablocks can be
+included into the relay chain every 6 seconds per core. This is the bottleneck that elastic scaling
+addresses. With elastic scaling, parachains can use multiple cores to include multiple parablocks
+within the same relay chain block.
```
profradha: "parachains can theoretically produce parablocks in the sub-second
time range" - this is not true.

Parablock creation process will still need 6 seconds. What the parachain collators can do is to produce multiple parablocks in 6 seconds.

```suggestion

With elastic scaling, parachains can use multiple cores to include multiple parablocks
within the same relay chain block.

```

docs/learn/learn-elastic-scaling.md
```diff
@@ -0,0 +1,77 @@
...
+
```
profradha, line 38: Provide more context

```suggestion

The relay chain receives a sequence of parachain blocks on multiple cores, which are validated and checked if all their state roots line up during their inclusion, but assume they’re unrelated parachain blocks during backing, availability, and approvals. With elastic scaling implemented, a parachain's throughput depends upon its collator infrastructure.

The [elastic scaling implementation](https://github.com/paritytech/polkadot-sdk/issues/1829) will be rolled out in multiple phases. In the first phase, elastic scaling is set to work on parachains with a trusted/permissioned collator set. With this restriction, it is possible to launch elastic scaling without changing the candidate receipt. After successfully implementing the first phase, changes can be made to the candidate receipt so the collator set can be untrusted/permissionless again. The final phase will feature full integration with the Cumulus framework, enabling parachains to be configured to access multiple cores continuously.

```

DrW3RK (commented):
Added more context on Elastic scaling and provided info on the phased rollout of its implementation.

The last section with figures and times will benefit from a review from the elastic scaling implementors.

filippoweb3:
@DrW3RK who would you suggest to give a final review on this PR?

DrW3RK:
@eskimor We are working on adding a high level Wiki doc on Elastic scaling. When you get time, could you please go over the content on this PR and provide feedback? Thank you.

eskimor (commented):
Looks good.

PR #5610, opened 2024-02-15
Update example.md

rogerjbos:
I am new to XCM, but it seems like Parent of Parachain C would be Relay B.  If the answer really is Relay A, more explanation should probably be provided.


DrW3RK:
You're right!

cc: @KiChjang

KiChjang (approved):

DrW3RK (approved):

PR #5613, opened 2024-02-16
Async backing upgrade guide: Replace images with code blocks

skunert:
I propose to replace the images from vscode with code blocks that are easy to copy from. I recently followed the steps so the guide itself should still be fine.

I used the code highlight features of docusaurus. However we did not have style defined for the highlighted lines, so I picked a color from the same color family.

Example: 
image


DrW3RK:
Sounds good! Thank you for replacing the images with code.

DrW3RK (approved):

PR #5627, opened 2024-02-23
Remove Ambiguity - OpenGov Decision and Confirmation Periods

### LINKED ISSUE #5625: OpenGov: Distinction between Voting Period and Decision Period is unclear
It seems like these two are being used synonymously.

### LINKED ISSUE #5626: OpenGov: Description of failure condition seems wrong
The condition for failure of a referendum is described here:

https://github.com/w3f/polkadot-wiki/blob/0eb4c4c8afc60575429565f6bd8177523138c89a/docs/learn/learn-polkadot-opengov.md?plain=1#L480-L482

Seems to me like this is wrong. In fact, this is what the in-code docs say:

https://github.com/paritytech/polkadot-sdk/blob/9bf1a5e23884921498b381728bfddaae93f83744/substrate/frame/referenda/src/lib.rs#L41-L50

The difference is this: According to the wiki, if a proposal reaches the end of its decision period and has not yet been confirmed, it is rejected, _even if it is confirming_. According to the in-code docs (which, admittedly, aren't exactly unambiguous), however, the proposal is only rejected when it has reached the end of its decision period not confirming (_passing_ and _confirming_ appear to be used synonymously).

In other words, according to the in-code docs, if a proposal starts confirming one block before the end of the decision period and stays in confirming state for the duration of the confirmation period, it is approved. If the proposal ever leaves confirming state after the close of the decision period, it fails (I'm kinda guessing here).

By my cursory reading of the code and some quick experiments, I'm fairly convinced the in-code docs are correct and the wiki is wrong.

DrW3RK:
Context available here https://github.com/w3f/polkadot-wiki/issues/5626

### DIFF
--- a/docs/learn/learn-polkadot-opengov.md
+++ b/docs/learn/learn-polkadot-opengov.md
@@ -309,7 +309,7 @@ include:
 - **Decision Period**: the time interval during which a proposal's outcome can be decided. 
 - **Confirmation Period**: the minimum amount of time the approval and
   support criteria must hold before the proposal is approved and moved to the enactment period. The
-  start of the confirmation period should be within the decision period.
+  confirmation period should start before the end of decision period.
 - **Minimum Enactment Period**: the minimum amount of waiting time before the proposed changes are
   applied
 - **Approval Curve**: the curve describing the minimum % of _aye_ votes as a function of time within
--- a/docs/learn/learn-polkadot-opengov.md
+++ b/docs/learn/learn-polkadot-opengov.md
@@ -269,7 +269,7 @@ Until they are in the lead-in period, proposals remain undecided. Once the crite
 the referendum moves to the _deciding_ state. The votes of the referendum are now counted towards
 the outcome.
 
-In (2), the proposal enters the **Decision Period** where votes can be cast. For a proposal to be
+In (2), the proposal enters the **Decision Period** where voting can continue. For a proposal to be
 approved, votes must satisfy the approval and support criteria for at least the **Confirmation
 Period**; otherwise, the proposal is automatically rejected. A rejected proposal can be resubmitted
 anytime and as many times as needed.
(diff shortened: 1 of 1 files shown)


docs/learn/learn-polkadot-opengov.md
```diff
@@ -306,10 +306,10 @@ include:
...
 - **Preparation Period**: the minimum amount of voting time needed before entering the Decision
   Period (given capacity and deposit are met).
-- **Decision Period**: the maximum time to approve a proposal. The proposal will be accepted if
-  approved by the end of the period.
-- **Confirmation Period**: the minimum amount of time (within the Decision Period) the approval and
-  support criteria must hold before the proposal is approved and moved to the enactment period.
+- **Decision Period**: the time interval during which a proposal's outcome can be decided. 
+- **Confirmation Period**: the minimum amount of time the approval and
+  support criteria must hold before the proposal is approved and moved to the enactment period. The
+  confirmation period should start before the end of decision period.
```
filippoweb3, line 312:
```suggestion
:::info Edge Case

A referendum can enter the confirmation period one block before the decision period ends. In this scenario, the referendum will pass if it satisfies approval and support thresholds for the minimum confirmation period (track-dependent).

:::
```

filippoweb3 (approved):
lgtm. Added the example of the edge case as it helps understand a bit better. We might need to create an additional figure later on.

docs/learn/learn-polkadot-opengov.md
profradha: Sounds good. I'll make sure to add it. Although, I don't think it was your intention to remove line 312 right?

DrW3RK:
I added your info note and a brief description about voting period, which takes care of the other issue.

filippoweb3:
all good to me

PR #5635, opened 2024-02-27
Async Backing Guide: Ignore warn block in prettier

skunert:
Bug:
image

A recent run of prettier seems to have messed up the warning block pretty bad, it included the entire rest of the page. It should be ignored now.


CrackTheCode016 (approved):
Good catch, thanks for this!!

PR #5647, opened 2024-03-02
NFT Pallets - Add option to use Apillon Storage

dimitrov-d:


CrackTheCode016:
@dimitrov-d thanks for the contribution! I agree with having Apillion here - but could we please keep Pinata as an option as well? Thanks1

dimitrov-d:
> @dimitrov-d thanks for the contribution! I agree with having Apillion here - but could we please keep Pinata as an option as well? Thanks1

Thank you for your review! I have made the changes you requested and left Pinata as an option, let me know if it's okay

CrackTheCode016 (approved):
Looks great, thank you so much!!

PR #5648, opened 2024-03-02
Update cosmos comparison page

### LINKED ISSUE #5645: Update Cosmos comparison page
There are still references to Democracy/Gov1 under https://wiki.polkadot.network/docs/learn-comparisons-cosmos#governance

The first two paragraphs in this section need to be essentially entirely re-done.

DrW3RK:
- Remove factually incorrect and outdated info
- Add a patch for OpenGov

A followup PR is needed to improve the content overall


CrackTheCode016 (approved):
lgtm!

A note for future comparison PRs.  It might be good to have a set explanation for consensus for these comparisons specifically, that way we don't repeat ourselves in different places

PR #5654, opened 2024-03-04
Update Polkadot RPC port in maintain-guides-how-to-validate-polkadot.md

DanielMagro97:
From 9944 to 9933


CrackTheCode016:
Thanks for the contribution! 

@DanielMagro97 have you tried this command, usually sending to the WS socket (9944) is enough, no?

DanielMagro97:
Hi @CrackTheCode016 , thanks for your reply.
Yes I tried the command, however 9944 is no longer the default WS port, as it is now the same as the RPC port, i.e. 9933.

CrackTheCode016 (approved):
@DanielMagro97 thank you so much!

PR #5785, opened 2024-04-14
Use curl instead of websocat in validator section

lobis:
I think using curl instead of websocat is more straightforward to new users as it's more readily available.


DrW3RK (approved):
Thank you!

PR #5789, opened 2024-04-16
add bridge asset transfer warning

acatangiu:
Add warning to encourage users to "prepare" their recipient account before doing bridge asset transfers to guard against possible asset-limit exhaustion attacks targeting their destination accounts.

### DIFF
--- a/docs/learn/learn-guides-DOT-KSM-bridge.md
+++ b/docs/learn/learn-guides-DOT-KSM-bridge.md
@@ -135,7 +135,7 @@ functionality is enabled, assets which are sufficient or non-sufficient on the A
 :::caution Avoid Asset Traps
 
 To avoid issues on the receiving side for non-sufficient assets, make sure to call
-[pallet_assets::touch()](https://github.com/paritytech/polkadot-sdk/blob/002d9260f9a0f844f87eefd0abce8bd95aae351b/substrate/frame/support/src/traits/misc.rs#L1209)
+[pallet_assets::touch()](https://github.com/paritytech/polkadot-sdk/blob/0ef37c75401b78b61ed35ce27af8b964da27bb3c/substrate/frame/assets/src/lib.rs#L1531) or [pallet_assets::touch_other()](https://github.com/paritytech/polkadot-sdk/blob/0ef37c75401b78b61ed35ce27af8b964da27bb3c/substrate/frame/assets/src/lib.rs#L1616)
 effectively guaranteeing the ability to successfully receive and accept the bridged assets in your
 account on the destination chain.
 This eliminates issues like your account on destination not existing or not having enough ED or having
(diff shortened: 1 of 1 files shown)


DrW3RK (approved):
Made minor edits to match Wiki style. Ready to be merged.

acatangiu (commented):
👍

docs/learn/learn-guides-DOT-KSM-bridge.md
```diff
@@ -126,3 +126,23 @@ balances of any account on Kusama Asset Hub can be queried on-chain through the
...
 ![Wrapped KSM Balance](../assets/bridge-hub/PAH-KSM-Balance.png)
+
+## Transfer of arbitrary assets between Polkadot Asset Hub and Kusama Asset Hub
+
+Transfer functionality of assets other than DOT and KSM between Asset Hubs is yet to be enabled. Once this
+functionality is enabled, assets which are sufficient or non-sufficient on the Asset Hubs can be bridged. 
+
+:::caution Avoid Asset Traps
+
+To avoid issues on the receiving side for non-sufficient assets, make sure to call
```
profradha, line 137: Can we add more clarity here? Let's say I'm transferring an asset from account A on Polkadot Asset Hub to the same account on Kusama Asset Hub which does not have the ED. On which chain should I make this call?

What if I'm sending it to account B on Kusama Asset Hub that I do not own, can I make this call for it?

acatangiu: > On which chain should I make this call?

Always on the recipient account of the destination chain. That is the one that will not accept the incoming assets if either:
- account does not exist yet and incoming assets do not satisfy creation ED,
- account already satisfies ED but has reached its "maximum number of different assets" limit.

A successful `pallet-assets::touch` will eliminate both preconditions above, and guarantees that a subsequent incoming transfer _of the same asset_ that has been "touched" will also be successful.

> What if I'm sending it to account B on Kusama Asset Hub that I do not own, can I make this call for it?

Yes, there is [pallet-assets::touch-other()](https://github.com/paritytech/polkadot-sdk/blob/0ef37c75401b78b61ed35ce27af8b964da27bb3c/substrate/frame/assets/src/lib.rs#L1616) for this case.

docs/learn/learn-guides-DOT-KSM-bridge.md
```diff
@@ -126,3 +126,23 @@ balances of any account on Kusama Asset Hub can be queried on-chain through the
...
+[pallet_assets::touch()](https://github.com/paritytech/polkadot-sdk/blob/002d9260f9a0f844f87eefd0abce8bd95aae351b/substrate/frame/support/src/traits/misc.rs#L1209)
```
acatangiu:
```suggestion
[pallet_assets::touch()](https://github.com/paritytech/polkadot-sdk/blob/0ef37c75401b78b61ed35ce27af8b964da27bb3c/substrate/frame/assets/src/lib.rs#L1531) or [pallet_assets::touch_other()](https://github.com/paritytech/polkadot-sdk/blob/0ef37c75401b78b61ed35ce27af8b964da27bb3c/substrate/frame/assets/src/lib.rs#L1616)
```

docs/learn/learn-guides-DOT-KSM-bridge.md
profradha: Thank you!

PR #5810, opened 2024-04-22
Mention the Asset Hub in the XCM monitoring sections

IkerAlus:
This section of the wiki page is about assets in general and Asset Hub in particular, however the section "XCM Transfer Monitoring" only mentions the relay chain. I just added the Asset Hub to the section since the same monitoring logic applies for DOT transfers in both chains.


CrackTheCode016 (approved):
Good catch, thank you!

PR #5826, opened 2024-04-25
Add Archival info on each section - Governance page

DrW3RK:
Due to algolia indexing and the section titles on the page that are very relevant to Polkadot OpenGov, it is necessary to have the content archival note at the beginning of every section.

This is ugly, but is needed till we find a solution for a floating banner on bottom right part of the page.


filippoweb3:
See https://github.com/w3f/polkadot-wiki/pull/5831

Ideally we can close this PR and merge the 5831.

We can then remove the old sticky banner code and add the msg box in all archive pages and PJS guides

PR #5839, opened 2024-04-29
add question, correct links

the-right-joyce:
- corrected the public notion page link (FAQ part deleted, only technical guidance now in there)
- corrected link to swapping header
- added question about different prices

### DIFF
--- a/docs/learn/learn-agile-coretime.md
+++ b/docs/learn/learn-agile-coretime.md
@@ -288,6 +288,7 @@ previous sales and can increase by 5x depending on the number of cores sold and 
 These numbers represent extreme cases though.
 
 #### Can the renewal price be different per taskID (Parachain)? Or will each sale cycle have a fix renewal price?
+
 It is in general different per task if they hit their caps.
 
 With the current configuration on Kusama, if the price is going up, every renewal pays a max of 3% more than they paid the sale before. That keeps going back to the sale cycle they first bought their coretime (or the sale cycle they renewed their lease). And because people buy in different sale cycles, or even at different times in each sale cycle, their prices can be wildly different from one another. There’s a special case for leaseholders at the end of their lease. They pay the base price in the sale in which they expire.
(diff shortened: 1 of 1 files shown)


docs/learn/learn-agile-coretime.md
```diff
@@ -287,6 +287,13 @@ The bulk price in Kusama might vary between a 50% decrease in case no cores were
 previous sales and can increase by 5x depending on the number of cores sold and when they sell.
 These numbers represent extreme cases though.
 
+#### Can the renewal price be different per taskID (Parachain)? Or will each sale cycle have a fix renewal price?
+It is in general different per task if they hit their caps.
```
profradha, line 292:
```suggestion

It is in general different per task if they hit their caps.
```

docs/learn/learn-agile-coretime.md
```diff
@@ -287,6 +287,13 @@ The bulk price in Kusama might vary between a 50% decrease in case no cores were
...
+
+With the current configuration on Kusama, if the price is going up, every renewal pays a max of 3% more than they paid the sale before. That keeps going back to the sale cycle they first bought their coretime (or the sale cycle they renewed their lease). And because people buy in different sale cycles, or even at different times in each sale cycle, their prices can be wildly different from one another. There’s a special case for leaseholders at the end of their lease. They pay the base price in the sale in which they expire.
```
profradha:
```suggestion
With the current configuration on Kusama, if the price is going up, every renewal pays a maximum of 3% more than they paid for the sale before. That keeps going back to the sale cycle they first bought their coretime (or the sale cycle they renewed their lease). And because people buy in different sale cycles, or even at different times in each sale cycle, their prices can be wildly different from one another. There’s a special case for leaseholders at the end of their lease. They pay the base price in the sale in which they expire.
```

DrW3RK (approved):

PR #5848, opened 2024-05-01
Update learn-agile-coretime.md

ltfschoen:
If you ask these teams whether this statement about what they are focusing on is true or not, they may reply that it isn't true, so propose to remove saying what they focus on.

cc @poppyseedDev


DrW3RK (approved):
Agreed.

PR #5915, opened 2024-05-24
Update build-tools-index.md

DrW3RK:
Feel free to add more tools. There are some tools that have been archived or unmaintained. We can clean them up through a follow-up PR.


CrackTheCode016 (approved):
Looks good to me; can't think of anything else to add.  _Maybe_ we could mention the different tools that the Polkadot SDK exposes, and how to install them, but that would be another PR anyway (`cargo install --git https://github.com/paritytech/polkadot-sdk --tag polkadot-v1.10.0 --force <the_tool>`)

PR #6068, opened 2024-07-11
Update min rank for 1KV on Polkadot

leonardocustodio:
The rank necessary to become part of the polkadot 1KV program has increased and can be checked in the form itself at https://forms.gle/LTCze7fJvqQinPqx6


CrackTheCode016 (approved):
LGTM, good catch, thanks!

PR #6100, opened 2024-07-24
Rm Kusama Index Page

filippoweb3:
- not used
- rm image


DrW3RK:
This was a legacy link for Kusama Guide that was listed in a document or website that was important. May be Polkadot Lite Paper? I do not remember exactly. But yeah, I agree it can be removed.

filippoweb3:
added a redirect to getting started kusama

CrackTheCode016 (approved):

PR #6122, opened 2024-08-03
Update build-data.md

LVweb3:
- The analytics panel offered by Dolpha has been discontinued more than 1 year ago after its proposal to the treasury was rejected.

- Web3Go has undergone a complete rebranding and is now known as DIN. You can confirm this by seeing that from the original domain of the Web3Go portal a redirection is made to the DIN landing page.


filippoweb3 (approved):
Thank you!

PR #6123, opened 2024-08-03
chore: clean up package manager config & artifacts

tien:
I'm trying to make some contributions, and noticed right away that:

- `package.json` will always be modified after each `yarn install`
- There's a warning for mixing lock files

Hence, creating this PR to clean those up. The deletion is quite a big change, so please feel free to have someone within the team do that in a separate PR 🙏.


filippoweb3 (approved):
lgtm, thank you for cleaning up! :)

PR #6143, opened 2024-08-09
Add metadata for systems parachains for Polkadot and Kusama

muddlebee:
have not tested this locally, but copied the RPC urls from here https://polkadot.js.org/apps/


CrackTheCode016 (commented):
This overall looks fine to me, but it would be a good idea to test locally @muddlebee 🙏

muddlebee:
> This overall looks fine to me, but it would be a good idea to test locally @muddlebee 🙏 

unfortunately I don't have a local copy as of now :( and it takes a lot of time to setup this repo.

Is there anyway we can do a quick deploy anywhere using CI? I have seen some UI repos has this to test quickly.

filippoweb3 (approved):
thank you! :)

PR #6178, opened 2024-08-26
Fix formatting on learn-identity

leonardocustodio:
I'm not sure what happened; the syntax seems correct.
Do we need to jump one line if we have the same tag twice in a row?

image


DrW3RK (approved):
Thanks for reporting the formatting issue. It should be fixed now.

PR #6200, opened 2024-09-02
Wiki reorg: Move Kusama pages to the Wiki

filippoweb3:
Create a Kusama section, adapt content, rm duplicated content.


CrackTheCode016 (approved):
Much more concise, this plus replacing conditional rendering will make for much more maintainable docs.

PR #6232, opened 2024-09-16
Adds Snowbridge library dependency

claravanstaden:
Corrects the fact that Snowbridge relies on a library for BLS signature verification.


claravanstaden:
Hey @filippoweb3! I made a small change to the bridges comparison table.

Additionally, I would like to know if this is truly accurate?

> Low latency, 5-7 minutes on reference hardware.

 As far as I know, Hyperbridge is bounded by the same constraints as Snowbridge - latency is dictated by the source chain finality mechanism, so if Ethereum takes 2 epochs to finalize, it would be the same latency for all trustless bridges.

filippoweb3:
hey @claravanstaden I remember Seun telling me that Hyperbridge has that latency because it does not use random sampling beefy but ZK circuits

claravanstaden:
> hey @claravanstaden I remember Seun telling me that Hyperbridge has that latency because it does not use random sampling beefy but ZK circuits

That would be Polkadot to Ethereum direction. Ethereum to Polkadot latency would be longer iiuc.

seunlanlege:
> Additionally, I would like to know if this is truly accurate?

This is in the Polkadot -> Ethereum direction, both snowbridge & Hyperbridge should have the same latency in the Ethereum to Polkadot direction.

filippoweb3:
@seunlanlege got it I will create a PR to correct that fact

filippoweb3 (approved):
thanks for this contribution! :D

PR #6234, opened 2024-09-17
Info bridge comparison

filippoweb3:
Continues https://github.com/w3f/polkadot-wiki/pull/6232


filippoweb3:
@claravanstaden I added info here about directions and latency

claravanstaden:
> @claravanstaden I added info here about directions and latency

Wonderful, thank you @filippoweb3! May I suggest moving the Ethereum > Polkadot direction first in the list? Imo it is the more important direction since it brings liquidity into Polkadot from Ethereum. :D

filippoweb3:
> Wonderful, thank you @filippoweb3! May I suggest moving the Ethereum > Polkadot direction first in the list? Imo it is the more important direction since it brings liquidity into Polkadot from Ethereum. :D

Added ;)

PR #6235, opened 2024-09-17
Wiki reorg: Remove conditional rendering BUILD section

filippoweb3:
Used "Relay Chain" mostly and "Polkadot" where needed.


CrackTheCode016 (approved):
LGTM except capitalization of "Relay Chain", which I am not sure if we are doing in the Wiki

PR #6238, opened 2024-09-18
Adds bridge integration protocol

claravanstaden:
Adds the relevant integration protocol for each bridge.


filippoweb3 (approved):
Thank you!

PR #6296, opened 2024-10-07
Edit Coretime Guides

filippoweb3:
Some minor changes:

- removed "advanced" as Coretime is already in the Advanced section
- removed "expect chaos" as this, I feel, creates anxiety when reading the guide.
- added page with mermaid diagram as intro with `doc` instead of `generated index`
- grammar checks

To consider:

We have now two pages explaining similar things
- Coretime for parachains
- Parachain template guide (that does not include any coretime ref in the name)
-> I suggest to merge the first into the second @CrackTheCode016

Those pages in the coretime guides section have the `build` slug, and if they are planned to stay in the learn section, it would be more appropriate to change it to `learn`.

### DIFF
--- a/docs/build/build-guides-coretime-start.md
+++ b/docs/build/build-guides-coretime-start.md
@@ -7,7 +7,7 @@ keywords: [coretime, blockspace, parathread, parachain, cores]
 slug: ../build-guides-coretime-start
 ---
 
-:::warning This section is under construction and moving! (Expect Chaos)
+:::warning This section is under construction.
 
 :::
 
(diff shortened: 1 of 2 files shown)


filippoweb3:
@CrackTheCode016 The section can be better structured with the Mermaid diagram visible at the top of the page and clickable links within the diagram that brings you to relevant Wiki pages. The Polkadot SDK intro is a bit generic and does not belong to the Coretime section IMO. I would move it to the relevant SDK page (if not already covered) and link to it.

CrackTheCode016:
@filippoweb3 They should be in build ideally - but as these will probably be merged with / migrated to the Polkadot Docs within the next month, ~~we can just change to learn initially.~~

Edit: Actually, let's put it in build, as it makes more sense overall and would be a bit more visible overall as right now it's pretty buried. 

> The section can be better structured with the Mermaid diagram visible at the top of the page and clickable links within the diagram that brings you to relevant Wiki pages. The Polkadot SDK intro is a bit generic and does not belong to the Coretime section IMO. I would move it to the relevant SDK page (if not already covered) and link to it.

This sounds like it can be made into its own section in Build maybe?

CrackTheCode016 (approved):
Reads much better! The changes here are fine to merge, perhaps the other items can be done in another PR?

Issue #6310, opened 2024-10-17
Update to Paseo the build guides and related docs


filippoweb3:
Follows PR https://github.com/w3f/polkadot-wiki/pull/6305

To do:
- [ ] update `build-parachains`
- [ ] update relevant links to that page, see `learn-parachains`
- [ ] update build guides

<img width="297" alt="Screenshot 2024-10-17 at 14 58 44" src="https://github.com/user-attachments/assets/7ea0e000-9da0-46ef-9b3a-a41928aea594">

filippoweb3:
solved with https://github.com/w3f/polkadot-wiki/pull/6313

PR #6313, opened 2024-10-17
Move + Remove Rococo from Coretime Guides

CrackTheCode016:
Addresses issue #6310

Might need some more updating to reflect the state of Paseo, but pages are moved and references are updated.

Screenshot 2024-10-17 at 3 05 59 PM


filippoweb3 (approved):
Thank you :)

PR #6349, opened 2024-10-31
Update build-guides-template-basic

juangirini:
Indicate that the region must begin before the collator can produce blocks under the bulk coretime scheme


DrW3RK (approved):
Thank you!

PR #6369, opened 2024-11-07
Wiki reorg: Introduce High-level Tech Terminology in Architecture

filippoweb3:
Within the context of Polkadot being a computer, we can use high-level terminology that more people are familiar with and then dive deeper and explain how those features are made possible.

Asynchronous backing is used to achieve **pipelining**
Agile coretime is used to achieve **multi-threading**
Elastic scaling is used to achieve **parallel computation** for one task


filippoweb3:
@Overkillus @shawntabrizi can you have a look at this PR and let me know what you think? Thank you in advance!

docs/learn/learn-agile-coretime.md
```diff
@@ -1,15 +1,30 @@
...
-Agile Coretime enables efficient utilization of Polkadot network resources and provides economic
-flexibility for builders, generalizing Polkadot beyond what was initially proposed and envisioned in
-its [whitepaper](https://polkadot.network/whitepaper/).
+import DocCardList from '@theme/DocCardList';
+
+[Multi-threading](<https://en.wikipedia.org/wiki/Multithreading_(computer_architecture)>) is a
+programming model where multiple threads (smaller sequences of programmed instructions) are created
+within a single process to perform multiple tasks at once. Multi-threading is commonly used to
+improve the performance of applications by executing different parts of a program concurrently.
+[Concurrency](<https://en.wikipedia.org/wiki/Concurrency_(computer_science)>) does not imply
```
profradha, line 27: Agile Coretime is a framework that gives you flexible options to access blockspace, with access to bulk coretime or on-demand coretime. The closest computer science term that corresponds to Agile Coretime is https://en.wikipedia.org/wiki/Scheduling_(computing), which improves efficiency. 

What could be referred to as multi-threading would be the interlacing and splitting of core regions, which is one of the features available in Agile Coretime, that lets the core be used by multiple parachain tasks (which can be treated as multiple threads). But does that provide performance improvement for parachains? No. It improves efficiency with scheduling.

shawntabrizi: > flexible options to access blockspace, with access to bulk coretime or on-demand coretime

this sounds less like a technical feature, and more about sales.

when you compare to cloud platforms, your ability to purchase resources on-demand, or in bulk, or whatever seems like the comparison.

filippoweb3: hey @DrW3RK and @shawntabrizi yes I was thinking about interlacing when doing this PR. I agree scheduling is more appropriate for agile coretime and then we can explain that scheduling enables multi-threading. I will add this feedback to the PR.

profradha: Thanks. Here are some chains that use multi-threading to improve performance, which are not available on the Polkadot tech stack yet. It would be great to see these features added on the collator side.

https://aptos.dev/en/network/blockchain/execution
https://blog.sui.io/parallelization-explained/

filippoweb3:
Thank you all for the comments! I implemented the feedback, and I am merging this PR. Any change can be addressed in a future PR, as I think these terminology changes will take some iterations before reaching the final state.

PR #6428, opened 2024-12-05
Fix grammar in text

karihirvi:
Dropped extra words that should not be there


filippoweb3 (approved):
Thanks for correcting this!

PR #6457, opened 2024-12-17
fix: update links to FAQ on how-to-validate-polkadot & how-to-validate-kusama pages

jarobar435:
Dear All,

upon going through the wiki, I found out the links to be failing:
Screenshot 2024-12-17 at 01 33 25
and it seems to be a minor issue with the unnecessary '.md' addition. Without it, the links should be working in both cases—when used from GitHub and on the production/live server.


filippoweb3 (approved):
Thanks for spotting this. The actual error was the `/` before specifying the page section with `#`

PR #6544, opened 2025-02-11
Add general case for XCM event monitoring

IkerAlus:
In a near future, certain parachains will enable more complex cross chain interactions which include the transfer of DOT from proxied accounts, using batches and other user workflows. For these cases the XCM instruction `transact` is usually used in destination which emits a different event than for usual transfers. This PR adds this consideration so that exchanges and similar entities dont miss deposits made via these special cases.


filippoweb3 (approved):
lgtm, thanks!

PR #6607, opened 2025-03-25
fix number of partial unbond chunks - Nomination pools

DrW3RK:
Context

Screenshot 2025-03-25 at 12 09 02


DrW3RK:
The value is 32 on both Polkadot and Kusama

Screenshot 2025-03-25 at 12 10 06

CrackTheCode016 (approved):

PR #6613, opened 2025-03-31
Update mkdocs.yml - Disable Social Cards

DrW3RK:
The social cards customizability is minimal for the free tier. A subscription is required to have access to full features. Temporarily disabling it till a decision is reached within the team.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -40,9 +40,6 @@ theme:
 plugins:
   - search
   - open-in-new-tab
-  - social:
-      cards_layout_options:
-        logo: assets/wiki-logo.svg
   - macros:
       module_name: macros/rpc
 


mkdocs.yml
```diff
@@ -40,9 +40,6 @@ theme:
 plugins:
   - search
   - open-in-new-tab
-  - social:
-      cards_layout_options:
-        logo: assets/wiki-logo.svg
```
filippoweb3, line 45:
```suggestion
- social:
      cards: false
```

filippoweb3 (approved):

mkdocs.yml
profradha: Are you sure this works?

It says that `cards: false` needs to be used in the document for which the card needs to be disabled. Also, it is a feature available with "insiders" tag which requires paid subscription https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/#disabling-social-cards

PR #6616, opened 2025-04-01
Remove inactive wallets

kategrowthland:
In my opinion, we should remove all the wallets that are no longer actively releasing new developments and align the wallet with ecosystem updates. Or at least we set them at lower positions. I know that it is ordered by alphabet, but from a user perspective, stuff listed at top positions means they are highlighted and recommended.


filippoweb3:
I generally agree with this. We can put them at the bottom and add a message telling the user that those wallets are not frequently updated (this would be the nicest option)

If we want to be safe from the user perspective, I would probably remove them all, as Nova and Subwallet are the go-to options for mobile platforms here.

@DrW3RK what do you think?

CrackTheCode016 (commented):
I would say this is fine, and in a follow up PR, we look to just linking to the wallets page on polkadot.com. @filippoweb3 WDYT?

filippoweb3 (approved):
lgtm. As said by @CrackTheCode016 we can think about pointing users to the relevant section on the Polkadot website and remove the page IF the content is redundant.

PR #6637, opened 2025-04-15
Fix MessageBox behavior

### LINKED ISSUE #6634: Can't get rid of Polkadot-JS warning
There is a little pop-up on Polkadot-JS related page. I click on the X and expect it to go away, but does not. It should disappear after clicking the X.

Image

CrackTheCode016:
Fix MessageBox behavior to listen for a click, rather than the DOM loading.

### DIFF
--- a/docs/javascript/MessageBox.js
+++ b/docs/javascript/MessageBox.js
@@ -2,8 +2,9 @@ document.addEventListener("DOMContentLoaded", () => {
   const messageBox = document.getElementById("messageBox");
   const closeButton = document.querySelector(".close-messagebox");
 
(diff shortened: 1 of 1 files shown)


CrackTheCode016:
@filippoweb3 Please see if it works properly by checking out the preview URL when it builds 😄

filippoweb3:
@CrackTheCode016 It seems the msg box disappears but you need to double click

filippoweb3:
@CrackTheCode016 I pushed a fix now it should work, lmk

filippoweb3 (approved):
I pushed a fix, now works on the first click

Issue #6655, opened 2025-04-29
[IA] Pre-reorg cleanup


CrackTheCode016:
- [x] Fork wiki as a backup - polkadot-wiki-archive (snapshot), as a way to preserve any pages we may need in the future
- [x] Identify which pages we need to keep from the build and maintain section -> https://github.com/w3f/polkadot-wiki/pull/6658
- [x] Place those into the IA -> https://github.com/w3f/polkadot-wiki/pull/6658
- [x] Remove Build and Maintain
- [x] Add Guides Section (move all guides from Learn)
- [x] Clean up sidebar external links / obvious pages -> https://github.com/w3f/polkadot-wiki/pull/6656
- [x] Clean up / remove all index pages -> https://github.com/w3f/polkadot-wiki/pull/6656

PR #6656, opened 2025-04-30
Remove links and index pages

filippoweb3:
Major cleanup of sidebar, removing links that are already redirected on the main website or that point to pages that are no longer maintained, such as Dotappstore.

For more context, see https://github.com/w3f/polkadot-wiki/issues/6655


CrackTheCode016 (approved):
lgtm, one thing we should try and do is allow the top-level sections (i.e., Funding) to lead to a top level sort of page as discussed before, but this can be done later after the IA restructure probably.

PR #6662, opened 2025-05-02
Edit Comparisons

filippoweb3:


filippoweb3:
hey @josepot can you provide feedback for this PR? I really appreciate any help you can provide.

filippoweb3:
@CrackTheCode016 let me know if these changes are OK with you, for more context about this see https://x.com/Joseptec/status/1917910708823490990

josepot (commented):
Apologies for the delay in reviewing this. I should also note that I'm not an expert in this domain, and my understanding of ZK rollups is quite limited.

That said, one thing I do know is that the state of ZK rollups is maintained via smart contracts on the L1 network. Because of this, it seems reasonable to infer that their finality is bound by the finality guarantees of the L1 itself.

If that’s correct (though I recommend confirming this with someone more knowledgeable), then any mention of “nearly instant finality” might be misleading and IMO should be removed.

docs/learn/learn-comparisons-rollups.md
```diff
@@ -47,7 +47,7 @@ example, optimistic rollups often have challenge periods of several hours or eve
 A **non-interactive method** (associated with validity proofs) relies on cryptographic mechanisms,
 such as SNARKs or STARKs, to validate transactions upfront. These proofs are generated off-chain and
 submitted to L1 for verification without requiring dispute resolution. While these methods offer
-near-instant finality, in theory, in practice delays are sometimes intentionally introduced as
+near-instant finality (limited by L1's finality), in theory, in practice delays are sometimes intentionally introduced as
```
CrackTheCode016: In light of what was said, along with what I know - it seems that it depends on the protocol. Ultimately, a rollup is only as good as where it is deriving its security/finality from (perhaps with the exception of a sovereign rollup). Therefore, I'll suggest the following:

```suggestion
potential near-instant finality on the rollup itself, it is ultimately limited by L1's finality (unless it is a sovereign rollup, in which case it would vary). In practice, delays are sometimes intentionally introduced as
```

docs/learn/learn-comparisons-rollups.md
```diff
@@ -145,7 +145,7 @@ be found on [l2beat](https://l2beat.com/scaling/summary).
...
+| **Finality**                 | Because the proof is immediately available, finality is only constrained by L1's finality (~13 minutes for Ethereum).                                                                                                                                                                 | Delayed finality (a week) due to fraud-proof mechanisms.                                                                                                                                 | Fast finality (under 1 minute) via relay chain consensus.                                                                                                                                                                                                                                                                                           |
```
CrackTheCode016:
```suggestion
| **Finality**                 | Because the proof is immediately available, finality is only constrained by L1's finality (for example, ~13 minutes for Ethereum).                                                                                                                                                                 | Delayed finality (a week) due to fraud-proof mechanisms.                                                                                                                                 | Fast finality (under 1 minute) via relay chain consensus.                                                                                                                                                                                                                                                                                           |
```

CrackTheCode016 (approved):
generally lgtm, it would be interesting to compare the notion of sovereign zk rollups and how they are relavent here in terms of finality.

filippoweb3:
thanks @josepot and @CrackTheCode016 for the review!

PR #6672, opened 2025-05-05
Edits sidebar

filippoweb3:
- added analytics
- moved dashboards
- added dashboard figure


CrackTheCode016 (approved):
LGTM, I think adding this as its own section makes sense and doesn't bury the dashboards below the surface.

PR #6674, opened 2025-05-06
New IA

filippoweb3:


CrackTheCode016 (approved):
Overall looks good to me from my end! 

I have one comment - namely "Decentralized Computer" as a main section title. Perhaps we could call this something less generic, that way it conveys the content? I.e., Decentralized Protocol, or just Protocol?

filippoweb3:
@CrackTheCode016 The idea is to put the architecture into the context of the decentralized computer narrative with virtual cores, kernel, etc. We can refine it with future PRs :)

Issue #6681, opened 2025-05-09
IA and Reorg Phase 2


filippoweb3:
Follows https://github.com/w3f/polkadot-wiki/issues/6655

- [x] Remove Archive folder and fix links
- [x] Reorg OpenGov page by removing Gov1 content
- [x] Target and remove unnecessary pages (example WASM page), eventually update redirects to the docs
- [x] Update contributing guidelines
--> prioritize static content
--> add content ONLY when released on Polkadot mainnet
--> avoid using "this new feature..." or "currently.." as this adds additional work for maintainers
- [ ] Work on the computer narrative in the architecture section
- [x] Update the Polkadot Vision section in "Getting Started"
--> Polkaodot 1.0, 2.0, JAM
- [ ] Update comparisons to add info in the graypaper (Solana and Cosmos)
- [x] Remove unused assets
- [x] Update learning resources in "Getting started" page
- [x] New landing page

CrackTheCode016:
I would also suggest that we begin refactoring each section. We should make sure of the following for each section of the Wiki: 

- Each section builds on the other, building a more chronological flow.
- Each page within the section should be formatted in the same manner:
  - A guide should have its own structure, and each guide formatted accordingly.
  - Each page in Architecture and Fundamentals also have their own structure and flow, ensuring the page is logically sound and does not contain any incomplete concepts.
  - Follows a style guide, i.e., the Google Style Guide (AI can be used to format without changing the content itself)

CrackTheCode016:
I'd also suggest that we move guides away from PolkadotJS and focus on mainstreaming them. This can be done by ensuring we use the tools in our ecosystem to replace the guides.

PR #6718, opened 2025-05-22
Update Landing Page

filippoweb3:
See https://github.com/w3f/polkadot-wiki/issues/6681


CrackTheCode016 (commented):
Looking good so far! I would say this is close to finished. There are two main things from preventing this from being merged IMO: 

1. I like the quad card design - but they should fall in the same theme. A few tweaks are needed, which I'll name:

   - Either remove border or make it black.
   - Ensure there is equal padding on all sides
   - Make text black and Unbounded
   - Smaller images
   - Ensure cards are square

2. For sections, should we display the whole sidebar for that section? I'm thinking it might help with empty space. Right now, when you click a section, it merely shows the top portion + the current section. 

Screenshot 2025-05-23 at 5 07 54 PM

material-overrides/home.html
```diff
@@ -0,0 +1,75 @@
...
+  <p>Learn about Polkadot's core architecture.</p>
+            </a>
+          </div>
+        </div>
+      
+        <div class="cta-section">
+          <div class="final-cta">
+            <div class="final-cta-content">
+            <h2>Polkadot-JS Guides</h2>
+            <p>Learn everything about the signature Polkadot developer tooling.</p>
```
CrackTheCode016:
```suggestion
            <p>Learn how to utilize Polkadot-JS to interact with Polkadot.</p>
```

CrackTheCode016 (approved):
LGTM! We can always make small adjustments later. Love the way it looks.

PR #6753, opened 2025-06-23
More info bags list

filippoweb3:
Specify the put in front of strategy


docs/learn/learn-staking-advanced.md
```diff
@@ -90,29 +88,34 @@ compounding the staking rewards), the position within a bag does not change auto
 scenario applies to a slashing event, i.e., when a nominator gets slashed, their position within a
```
michalisFr, line 88: > If one decides to send staking rewards to the stash account and automatically bond them (i.e.
compounding the staking rewards), the position within a bag does not change automatically.

This is incorrect. The position within the bag doesn't change automatically anyway. What doesn't change in this case is the bag, i.e. there is no automatic rebagging when compounding rewards or you get slashed. Rebagging happens automatically when the change in bonded is a result of an action, like bonding more or unbonding.

michalisFr: It's mentioned correctly below. You probably forgot to fix it in this sentence.

docs/learn/learn-staking-advanced.md
```diff
@@ -90,29 +88,34 @@ compounding the staking rewards), the position within a bag does not change auto
...
 (12500 on Kusama) come out as the electing nominators. See
 [Staking Election Stages](learn-nominator.md#staking-election-stages) section for more info.
 
 This means that only a portion of the nomination intents is kept. Once the nomination period ends,
-the NPoS election system takes all nomination intents and their associated votes as input, and it
+the NPoS election system takes all nomination intents and their associated votes as input and 
 outputs a set of validators. The bags are iterated from the most staked to the least staked. If the
-accounts are not appropriately sorted, this could leave the last touched bag to only be partially
-iterated. Thus, in some edge cases, the order of the members within a bag is important. Continuing
+accounts are not appropriately sorted, the last touched bag could only be partially
```
michalisFr: > If the accounts are not appropriately sorted,

I think you're trying to say something different, but it doesn't make much sense in this way. What you need to say here is what you describe in the example below: The algorithm will stop once it's selected 22500 nodes so the last touched bag will probably be only partially iterated. As such the order in the last bag matters.

docs/learn/learn-staking-advanced.md
```diff
@@ -90,29 +88,34 @@ compounding the staking rewards), the position within a bag does not change auto
...
-storage. In the current staking system configuration, at most 22500 nominators in the bags-list
+!!!info "Important Notes"
+      The `putInFrontOf` extrinsic does not check which account is the lightest (i.e., the account with the least tokens where you can move in front of), as opposed to `rebag`, which automatically puts you in the right bag. Suppose you have the account with 7 DOT and your bag configuration is 8 3 4 9 1 7 5. You can move in front of 1 or 4 (i.e., lighter accounts) or in front of 3 (i.e., lightest account). Ideally, you need to find the lightest account and give it as an argument to the extrinsic. This is what Polkadot-JS UI does for you: it finds the "lightest" account in the bag and puts you in front of it.
+
+      If you want to have the best placement in the bags list, the recommended order of actions is to first rebag yourself (if possible) and then issue the `putInFrontOf` extrinsic.
+
+      `Rebagging` and `putInFrontOf` only matter if you are in the last bag (i.e., close to the minimum active bond). You do not need to issue those extrinsics if you bond way above that value.
+
```
michalisFr, line 101: I would put the new section above, after the section below. I think it makes more sense in the reading order.

PR #6790, opened 2025-07-22
Update Offenses

filippoweb3:


michalisFr (changes requested):
Ok, that was simple. The only remark is that On-chain disablings also have priorities (as described in stage 4 [here](https://github.com/paritytech/polkadot-sdk/issues/4359))

Issue #6787, opened 2025-07-16
With v1.6.0 release of Polkadot, staking will use holds instead of locks


Ank4n:
In some places, in wiki we mention staking as frozen balance. 

As part of [these changes](https://github.com/paritytech/polkadot-sdk/pull/5501) that goes live on polkadot v1.6.0 runtime release (today), we will be migrating all staking accounts to use holds (shown as reserved in system account info) instead of locks (frozen section).

I found one [here](https://wiki.polkadot.network/learn/learn-account-balances/#:~:text=Frozen%20is%20the%20free%20balance%20locked%20for%20staking%2C%20governance%2C%20and%20vesting%20(also%20called%20locked%20balance)) but there might be other mentions as well.

Ank4n:
Linking the related forum post
https://forum.polkadot.network/t/upcoming-change-to-how-staked-funds-are-managed-locks-holds/13779

PR #6793, opened 2025-07-23
fix: Account Balances page

Imod7:
### Description
Related to this issue https://github.com/w3f/polkadot-wiki/issues/6787 

### Review
Review from @Ank4n would be very helpful! Thank you very much in advance!


docs/learn/learn-account-balances.md
```diff
@@ -12,14 +12,13 @@ remain frozen and unused due to an on-chain requirement.
 
 There are 5 types of account balances:
 
-- **Free** is the balance that can be used for on-chain activity like staking, participating in
-  governance etc. but is not necessarily spendable (or transferrable)
-- **Frozen** is the free balance locked for [staking](./learn-staking.md),
-  [governance](./learn-polkadot-opengov.md), and [vesting](./learn-transactions.md#vested-transfers)
+- **Free** is the balance that can be used for on-chain activity like participating in
```
Ank4n: This is complex to explain, but here is how I would explain it. I would still like to get more eyes on this to verify @filippoweb3 @DrW3RK 

- Free Balance: Balance that can be used for any on-chain activity (staking, governance, deposits) as long as your total balance (free + reserved) remains above the maximum of frozen balance and existential deposit.

- Frozen Balance (also called locks): This is a balance that overlaps across pallets . Example: If governance freezes 100 DOT and vesting freezes 120 DOT, total frozen = 120 DOT (not 220 DOT).

- Reserved Balance (also called holds): Balance removed from free and doesn't overlap. Used by Nomination Pools, Staking, etc. Can still be used for governance voting but not for transfers or fees.

- Spendable Balance: Portion of free balance available for transaction fees and creating new holds.

filippoweb3: I am on it, I need to verify a couple of things but there is already PR in place to update the balances page.

Imod7: @filippoweb3 Let me know if you would like to incorporate the changes reviewed by Ankan into your PR, and I can close mine. Otherwise, I am happy to add them to this PR.

Imod7: Just pushed the changes suggested from @Ank4n in this [commit](https://github.com/w3f/polkadot-wiki/pull/6793/commits/12491d7a5b548353d8d58bacb0430ddb2d0557e4)

docs/learn/learn-account-balances.md
```diff
@@ -12,16 +12,10 @@ remain frozen and unused due to an on-chain requirement.
 
 There are 5 types of account balances:
```
joepetrowski: 4

Imod7: Updated in this [commit](https://github.com/w3f/polkadot-wiki/pull/6793/commits/5bf5ce7d574ba3dfd4ec86f2a90deb9f94f6702b)

docs/learn/learn-account-balances.md
```diff
@@ -10,19 +10,12 @@ remain frozen and unused due to an on-chain requirement.
...
-- **Untouchable** is the portion of the free balance that cannot be moved (i.e., not spendable) but
-  can still be used for on-chain activity
+There are 4 types of account balances:
+
+- **Free** is the balance that can be used for any on-chain activity ([staking](./learn-staking.md), [governance](./learn-polkadot-opengov.md), deposits) as long as your total balance (free + reserved) remains above the maximum of frozen balance and existential deposit.
+- **Frozen** (also called locks) is a balance that overlaps across pallets. Example: If governance freezes 100 DOT and vesting freezes 120 DOT, total frozen = 120 DOT (not 220 DOT).
+- **Reserved** (also called holds) is the balance removed from free and doesn't overlap. Used by nomination pools, [staking](./learn-staking.md), etc. Can still be used for governance voting but not for transfers or fees.
+- **Spendable** is the portion of free balance available for transaction fees and creating new holds.
 
 The spendable balance is calculated as follows:
```
joepetrowski, line 27: This needs to be updated as part of this PR. E.g. `on_hold` is no longer defined. Does the formula need an update? @Ank4n

joepetrowski (changes requested):
There are still changes below (spendable calculation, examples) that need to be made.

docs/learn/learn-account-balances.md
Imod7: I updated the formula in this [commit](https://github.com/w3f/polkadot-wiki/pull/6793/commits/0db0d5b93375c7a1b1dcee6590ae6308116230f1). If there are any objections from Ankan I will edit again.

Imod7:
> There are still changes below (spendable calculation, examples) that need to be made.

- I updated the examples but the changes need to be double checked. 
- Once the example changes are confirmed, the photos will need to be aligned. I am not sure what tool to use to edit them, but I assume @filippoweb3 will know how to handle this. 
- Also, this section [Query Account Data in Polkadot-JS](https://wiki.polkadot.network/learn/learn-guides-accounts/#query-account-data-in-polkadot-js) was updated.
- `Untouchable` should either be added again in the balance types or be removed from the examples to maintain consistency.

filippoweb3:
@Imod7 I will have a look by EoW, will update the figures as well. I will push my edits, ask for review by @michalisFr and merge. Thank you so much for starting this PR!

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
+    - native [staking](./learn-staking.md) on the relay chain or via nomination pools
+    - deposits such as [governance](./learn-polkadot-opengov.md) decision and submission deposits, [identity](./learn-identity.md) deposits, and [proxy](./learn-proxies.md) and [multi-signature](./learn-account-multisig.md) accounts deposits. It cannot be used for transfers or paying fees.
+
+- **Frozen Balance** (also called locks, or locked balance) is a balance that overlays. The frozen balance can exceed the total balance. Frozen balance is used for:
+    - [vested transfers](./learn-transactions.md#vested-transfers)
+    - governance locks
+
+    Locks overlay with themselves and with holds, meaning that if staking reserves 60 DOT, voting for a governance proposal with 20 DOT will put a lock on 20 out of 60 reserved DOT. If a governance vote freezes 20 DOT and vesting freezes 120 DOT, the total frozen balance is 120 DOT (not 140 DOT).
+
+- **Spendable Balance** is the portion of free balance available for transaction fees and creating new holds.
```
michalisFr: **Spendable Balance** is the portion of free balance that can actually be transferred and is also available for transaction fees and creating new holds.

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
-Free: 60 DOT
-Frozen: 80 DOT
-Onhold: 40 DOT
-Spendable: 20 DOT
-Untouchable: 40 DOT
+Free: 20 DOT
+Frozen (locked): 20 DOT
+Reserved (held): 80 DOT
+Spendable: 19 DOT (Free - ED)
+Untouchable: 21 DOT (ED)
```
michalisFr: Untouchable will still be 1 DOT (just the ED) since it's "the frozen balance **in excess** of any holds"

This means the graph needs to be updated as well

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
 ```
 
 ![balance-example-4](../assets/balance-example-4.png)
 
-This update using the fungible trait allows the use of the frozen balance for on-chain activity like
-setting up proxies and identities. Note that
+Note how, through the fungible trait, the system uses the reserved balance. In other words, locks are
+subtracted from the free balance but overlap with the reserved balance, and with themselves (see below). The free and reserved portions remain 20 DOT and 80 DOT, respectively. We also have 20 DOT as a frozen balance because of the governance lock, and the untouchable balance would thus be 21 DOT.
```
michalisFr: > locks are subtracted from the free balance

They are not. They are imposed on the total balance, as shown in the numbers above and explained in the nest sentence: free isn't reduced. This probably needs to be reworded.

For example, with the 100 DOT free if you vote for governance with 20 DOT (and nothing else) you'd have:
Free: 100
Frozen: 20
Reserved: 0
Untouchable: 20 (not 21)
Spendable: 80

But I'd like to @joepetrowski or @Ank4n to confirm.

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
+A vested transfer of 50 DOT will
+grow the frozen balance to 50 DOT and the untouchable balance to 51 DOT, using the already frozen 20 DOT plus an additional 30 DOT from the reserved balance without touching the free and spendable balances.
+
+
+```
+Free: 20 DOT
+Frozen (locked): 50 DOT
+Reserved (held): 80 DOT
+Spendable: 19 DOT (Free - ED)
+Untouchable: 51 DOT (ED)
```
michalisFr: A vested transfer increases the total balance as well. At the end of the day when all locks and reserves are removed you should have 150 DOT transferable. Also, in this example you suddenly have more frozen than free, so you lose access to the 20 DOT you had before.

What would happen here is this:
Free: 70
Frozen: 50
Reserved: 80
Spendable: 69
Untouchable: 1

However, since Frozen overlaps with Reserved, as you can see you have that 50 extra DOT as spendable from day 1. But at the end of the day that's not a problem since if you remove all reserves and the governance lock, you'll still have 50 DOT frozen that will become spendable according to the vesting schedule. 

But again I'd like @joepetrowski or @Ank4n to confirm.

docs/learn/learn-account-balances.md
```diff
@@ -135,18 +142,18 @@ Locks overlap (in both amount and duration), and the general rule is that:
 - If you have multiple locks of the same amount of tokens, the lock with the longest duration
   decides when those tokens can be unlocked
 
-Let's take, for example, 80 DOT as a `frozen` balance. These 80 DOT are currently used in staking
+Let's take, for example, 80 DOT as a `frozen` balance. These 80 DOT are currently used in vesting
 and governance as follows:
 
-- 80 DOT Staking (just unbonded) -> lock 28 days
+- 80 DOT Vesting (just unbonded) -> lock 28 days
```
michalisFr: > - 80 DOT Vesting (just unbonded) -> lock 28 days

"- 80 DOT Vesting that become fully vested in 28 days"

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
-  and it is no longer free (also called reserved balance)
-- **Spendable** is the free balance that can be spent
-- **Untouchable** is the portion of the free balance that cannot be moved (i.e., not spendable) but
-  can still be used for on-chain activity
+In the Polkadot ecosystem, different types of balances depend on the account activity.
+Different balance types dictate whether your balance can be used for transfers, to pay fees, or must remain frozen and unused due to an on-chain requirement.
+
+There are four types of account balances:
+
+- **Free Balance** is a portion of an account's total balance that is not held (see below). It is the balance that can be used for any on-chain activity ([staking](./learn-staking.md), [governance](./learn-polkadot-opengov.md), and deposits) as long as the total balance (free + reserved) remains above the maximum of frozen balance and existential deposit.
```
Ank4n: I suggest we stick to just these three terms: free, frozen, and reserved, as reflected in the meta account info. In the frozen section, we can note that it’s sometimes referred to as locks, and in the reserved section, we can mention that it’s also called hold balance. Wdyt?

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
+
+- **Reserved Balance** (also called holds, or held balance) is the balance removed from free and does not overlay. It can be slashed, but only after all the free balance has been slashed. Reserved balance is used for:
```
Ank4n, line 13: >...but only after all the free balance has been slashed

Not sure what this means - shouldn’t only the reserved funds be subject to slashing? Why would the free balance be slashed?

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
+    - native [staking](./learn-staking.md) on the relay chain or via nomination pools
+    - deposits such as [governance](./learn-polkadot-opengov.md) decision and submission deposits, [identity](./learn-identity.md) deposits, and [proxy](./learn-proxies.md) and [multi-signature](./learn-account-multisig.md) accounts deposits. It cannot be used for transfers or paying fees.
+
+- **Frozen Balance** (also called locks, or locked balance) is a balance that overlays. The frozen balance can exceed the total balance. Frozen balance is used for:
```
Ank4n: > The frozen balance can exceed the total balance

That’s technically correct, but I’m not sure why it would ever happen in practice?

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
+    - [vested transfers](./learn-transactions.md#vested-transfers)
+    - governance locks
```
Ank4n:
```suggestion
    - governance votes
```

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
+
+    Locks overlay with themselves and with holds, meaning that if staking reserves 60 DOT, voting for a governance proposal with 20 DOT will put a lock on 20 out of 60 reserved DOT. If a governance vote freezes 20 DOT and vesting freezes 120 DOT, the total frozen balance is 120 DOT (not 140 DOT).
```
Ank4n: Just to not keep using multiple terminologies to say the same thing.
```suggestion
    Freezes overlay with themselves and with reserves, meaning that if staking reserves 60 DOT, voting for a governance proposal with 20 DOT will put a freeze on 20 out of 60 reserved DOT. If a governance vote freezes 20 DOT and vesting freezes 120 DOT, the total frozen balance is 120 DOT (not 140 DOT).
```

docs/learn/learn-account-balances.md
Ank4n: I would also note that, frozen balance is simply the minimum total balance the user needs to maintain (inclusive of reserves).

docs/learn/learn-account-balances.md
Ank4n: Formula looks good. 

Also aligns with https://github.com/w3f/polkadot-wiki/pull/6793/files#r2231235271

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
-**Wallet providers might show you the spendable, locked, and reserved balance.**
+**Wallet providers might show you the spendable, frozen, and reserved balance.**
 
 ## Example of Account Balance Types
 
 Below is an in-depth example of how an account balance composition changes depending on user actions
-once
-[the _fungible_ trait](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/index.html)
-is used for account balances. Let’s take, for example, an account with 100 DOT.
+**once [the _fungible_ trait](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/index.html) is implemented by all Substrate pallets**. Let’s take, for example, an account with 100 DOT.
```
Ank4n, line 54: Not sure why this detail is needed about the fungible migration?

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
 
 ```
 Free: 100 DOT
-Frozen: 0 DOT
-On hold: 0 DOT
+Frozen (locked): 0 DOT
+Reserved (held): 0 DOT
 Spendable: 99 DOT
 Untouchable: 1 DOT (ED)
```
Ank4n, line 61: Note for future: Post AHM, all these operations will be on AH and ED would reduce to 0.1 DOT.

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
 ```
 
 ![balance-example-1](../assets/balance-example-1.png)
 
-In this case, the existential deposit of 1 DOT is untouchable (meaning you can’t touch it if the
-account can’t or shouldn’t get reaped). If 80 DOT from the account is staked, we get the following
-balance structure:
+The untouchable balance is part of the free balance that cannot be spent due to ED or freezes. In this case, the existential deposit of 1 DOT is untouchable (meaning you can’t touch it if the
```
Ank4n, line 66: Not sure if it’s explained elsewhere, but I’d briefly mention that the ED acts as a deposit to cover storage and keep the account alive.

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,123 @@ title: Account Balances
...
-is used for account balances. Let’s take, for example, an account with 100 DOT.
+**once [the _fungible_ trait](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/index.html) is implemented by all Substrate pallets**. Let’s take, for example, an account with 100 DOT.
 
 ```
 Free: 100 DOT
-Frozen: 0 DOT
-On hold: 0 DOT
+Frozen (locked): 0 DOT
+Reserved (held): 0 DOT
 Spendable: 99 DOT
```
Ank4n, line 60: A subtle nuance to explain is that the spendable balance is 99 DOT if you intend to keep the account alive. However, if you’re fine with the account being reaped, you can transfer up to the full 100 DOT (minus transaction fees).

docs/learn/learn-account-balances.md
filippoweb3: added an example for a second gov lock as vesting can be a further example.

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,122 @@ title: Account Balances
...
+Reserved (held): 80 DOT
+Spendable: 19 DOT (Free - ED)
+Untouchable: 1 DOT (ED)
 ```
 
 ![balance-example-4](../assets/balance-example-4.png)
 
-This update using the fungible trait allows the use of the frozen balance for on-chain activity like
-setting up proxies and identities. Note that
+Note how, through the fungible trait, the system uses the reserved balance. Locks are
```
Ank4n:
```suggestion
Note how, through the fungible trait, the system uses the reserved balance. Freezes are
```

docs/learn/learn-account-balances.md
```diff
@@ -3,115 +3,122 @@ title: Account Balances
...
+imposed on the total balance, they overlap with the reserved balance, and with themselves (see below). The free and reserved portions remain 20 DOT and 80 DOT, respectively. We also have 20 DOT as a frozen balance because of the governance lock. The untouchable balance remains 1 DOT as it is defined as the frozen balance in excess of any holds (in this case frozen < holds, so only the ED is untouchable).
```
Ank4n:
```suggestion
imposed on the total balance, they overlap with the reserved balance, and with themselves (see below). The free and reserved portions remain 20 DOT and 80 DOT, respectively. We also have 20 DOT as a frozen balance because of the governance vote. The untouchable balance remains 1 DOT as it is defined as the frozen balance in excess of any reserved (in this case frozen < reserved, so only the ED is untouchable).
```

Ank4n (commented):
I added a few suggestions, but overall, great job.

docs/learn/learn-account-balances.md
filippoweb3: specified

docs/learn/learn-account-balances.md
ggwpez: I did not read through the context here, but i had this in my notes. Will compare against it:  
```pre
|__total__________________________________|
|__on_hold__|_____________free____________|
|__________frozen___________|
|__on_hold__|__ed__|        
            |__untouchable__|__spendable__|

untouchable = max(ed, frozen - on_hold)
```
from [here](https://github.com/paritytech/polkadot-sdk/issues/1833#issuecomment-1805764506).

docs/learn/learn-account-balances.md
filippoweb3: Not all pallets currently implement the fungible trait

docs/learn/learn-account-balances.md
filippoweb3: yeah, the underlying balance system will remain the same

docs/learn/learn-account-balances.md
filippoweb3: added a link to the Wiki page explaining it

docs/learn/learn-account-balances.md
michalisFr: This is taken from here @Ank4n 
https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/index.html#terminology

docs/learn/learn-account-balances.md
```diff
@@ -97,24 +97,23 @@ Free: 20 DOT
...
 
 ```
 Free: 20 DOT
-Frozen (locked): 50 DOT
+Frozen (locked): 85 DOT
 Reserved (held): 80 DOT
-Spendable: 19 DOT (Free - ED)
-Untouchable: 51 DOT (ED)
+Spendable: 14 DOT (Free - untouchable)
+Untouchable: 6 DOT (ED + frozen in excess of reserved)
```
michalisFr: It is:
Untouchable: 5
Spendable: 15

Remember the visual aid how frozen overlaps with ED and untouchable is only the excess from the holds.

docs/learn/learn-account-balances.md
ggwpez: I think this is sadly not correct, since a reserved balance creates a provider ref which makes the ED spendable. I will check again how it is exactly, but more complicated than the formula currently.

ggwpez: IMO it should be something like this. It is also not 100% correct though, see the NOTE. So not sure how to *easily* explain this. In practice its not easy.

```rust
untouchable = frozen - on_hold.min(frozen);

// NOTE: The `(on_hold > 0 || frozen > 0)` is an *approximation* for
// `(consumers != 0 && providers <= 1)`, which is the condition that actually matters.
if free > 0 && (on_hold > 0 || frozen > 0) {
	untouchable = untouchable.max(ed);
}

spendable = free - untouchable.min(free);
```

Ank4n: We have a [test case here](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/staking-async/src/tests/bonding.rs#L63) demonstrating that when a hold exists, the ED is not spendable.

Here’s the `AccountInfo` for 11 from that test (not printed in the test output, but retrieved by running it locally):

```
AccountInfo {
    nonce: 0,
    consumers: 1,
    providers: 1,
    sufficients: 0,
    data: AccountData {
        free: 1,
        reserved: 1000,
        frozen: 0,
        flags: ...
    }
}
```

docs/learn/learn-account-balances.md
Ank4n: Thanks for pointing out. This is not correct anymore. I think the original text was correct when staking was using locks, and it was part of free balance.

(I will make a fix in the fungible rustdoc)

Ank4n:
I had a question today about voting locks, and it wasn’t clear to the person asking — and probably not to many others either — that governance votes use the total balance, including non-staking holds (like proxy deposits).

Might be worth calling this out explicitly in the docs as well:
https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/conviction-voting/src/lib.rs#L425

Also, I think we should go ahead and merge this (post handing current open comments). It’s not perfect, but already a clear improvement over what’s currently in the docs. We can continue to iterate on it from here.
Approving this PR for that reason.

Ank4n (approved):

docs/learn/learn-account-balances.md
Ank4n: (pending review) [fix](https://github.com/paritytech/polkadot-sdk/pull/9347/files#diff-f7dca3bd40e6b09cd77c3a2ca9c9a44b43897987fcbfa7dd4abac722dac68c57L54)

Ank4n: @filippoweb3 This text should be fixed too. Totally misleading as it is now. You can refer the [text here](https://github.com/paritytech/polkadot-sdk/pull/9347/files#diff-f7dca3bd40e6b09cd77c3a2ca9c9a44b43897987fcbfa7dd4abac722dac68c57L54).

Ank4n:
@filippoweb3 Regarding the [balance examples](https://wiki.polkadot.com/learn/learn-account-balances/#example-of-account-balance-types), how are they currently being cross-verified? Should we add dedicated unit tests in the SDK to match each example one-to-one? Or any other idea?

filippoweb3:
@Ank4n we validate with Polkadot JS and the support team :) I would wait to build unit tests until those figures and examples are final, as the Wiki content is subject to changes, and this would add additional work to update test dependencies. Thanks!

PR #6820, opened 2025-08-02
Fix dark mode white flash and theme switching issues

### LINKED ISSUE #6811: Flickering Issue in dark mode
There is a slight transition to light mode when changing the menu items in dark mode.

ss-deshmukh:
Issue #6811 
- Implemented CSS first approach (Previously, the site would load in light mode by default, then JS would override to dark mode)
- Removed overriding and instead added separate CSS injections (Theme switch function now reloads page to start with a clean state to avoid interference. )
- New Pages opened in respective themes have smoother transition (Without flashes)


filippoweb3 (approved):
Awesome! Thank you for the fix! :D

PR #6850, opened 2025-08-24
Fix redirects

### LINKED ISSUE #6849: Check all redirects
There have been reports of old wiki links not working. We must test each redirect and ensure we do not get 404 errors.

filippoweb3:


CrackTheCode016 (commented):
It looks good, but we might have to add some catch alls for maintain, build, general, etc. like so:

```
- /*/maintain -> https://docs.polkadot.com/infrastructure
- /*/general ->  /general/web3-and-polkadot
```

other than that, imo looks good

PR #6905, opened 2025-10-20
AHM: Guides

### LINKED ISSUE #6901: Update guides to reflect AHM
In the guides, ensure any place the relay chain is mentioned (whether via Subscan) properly references the Hub instead.

CrackTheCode016:
This PR aims to ensure all guides are in-line with AHM.

Closes #6901

### DIFF
--- a/docs/learn/learn-guides-accounts.md
+++ b/docs/learn/learn-guides-accounts.md
@@ -18,7 +18,7 @@ description: Comprehensive guides for managing accounts, backups, and security u
 
 ## Account Address Format
 
-An account created on the relay chain can also be used on multiple chains in the ecosystem. More
+An account created on the Polkadot Hub can also be used on multiple chains in the ecosystem. More
 specifically, the account of a chain that uses the `*25519` account address format (the latest list
 can be accessed on the
 [ss58 registry repository](https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json))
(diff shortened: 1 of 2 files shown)


CrackTheCode016:
This PR is rather simple, and should be good to go, but we do need to check the following: Will PJS default to its respective RC (Kusama RC, Polkadot RC, etc) or will it go to its respective Asset Hub? Do we need to inform users of this when using PJS apps?

docs/learn/learn-guides-accounts.md
```diff
@@ -18,14 +18,14 @@ description: Comprehensive guides for managing accounts, backups, and security u
 
 ## Account Address Format
 
-An account created on the relay chain can also be used on multiple chains in the ecosystem. More
+An account created on the Polkadot Hub can also be used on multiple chains in the ecosystem. More
```
midegdugarova: I wonder if its Polkadot Hub or Polkadot. I think Polkadot would make more sense for new users?

midegdugarova (approved):
thanks! looks good, but as we discussed we will abstract "Asset Hub" away when we do the overhaul.

docs/learn/learn-guides-accounts.md
CrackTheCode016: I was wondering the same. We can just put Polkadot as it does eliminate confusion

CrackTheCode016: Removed it @midegdugarova

PR #6915, opened 2025-10-27
AHM: Proxies & Multisig

### LINKED ISSUE #6900: Update Proxies and multisig pages after AHM
FAQ: https://docs.google.com/document/d/1XR3vL2p4QV0wC7FrlC8eN-q62BqNFTFElbj21wEmMGg/edit?tab=t.6yd3zdf50r9d

CrackTheCode016:
This PR focuses on proxies, pure proxies, and multisig in relation to the AHM. 

- [x] Proxies - one minor change, which is that proxy announcements are **not** on the relay chain and not **not.** migrated to the AH. Does this mean they have to manually set them?
- [x] Multisig - ongoing transactions will not be migrated and need to be completed before migration, however, I feel this is unnecessary for the Wiki and needs to be a social media announcement.
- [x] Pure proxies - still waiting on final plan / verdict on them ?

Closes #6900


midegdugarova:
@CrackTheCode016 
My understanding is that only existing proxy announcements are not migrated  but deleted (although I do not think there are many existing at the time of migration, if any). The announcement usually have a short lifespan, a few hours to a few days - as long as the delay was defined. The new announcements will live on AH and will be callable the same way as before. 
But yeah, if there are any existing announcements, they need to be repeated manually.

What is with pure proxies?

docs/learn/learn-proxies-pure.md
```diff
@@ -42,6 +42,9 @@ proxy channels) within proxy calls via `proxy.proxy` extrinsic. [For example](./
 proxies_ within a multisig is possible. Using proxy calls, it is possible to use the _any_ proxy to call the
 _pure_ proxy, which in turn will do a multisig call.
 
+!!!info "Staking Migration to Asset Hub"
+    **Pure proxies** are automatically migrated to Asset Hub. Additionally, **"Any" type pure proxies** are kept on the Relay Chain to prevent them from becoming inaccessible. **Proxy announcements are not migrated** and will be deleted from the Relay Chain.
```
midegdugarova: relay chain

midegdugarova (approved):

PR #6933, opened 2025-11-10
Revert banner post-AHM

CrackTheCode016:
AHM has been completed; therefore, the banner is now defunct. Reverted back to the hackathon announcement


CrackTheCode016:
Appears to have been fixed already

PR #6964, opened 2025-12-15
Fix chain-state values post AHM

### LINKED ISSUE #6960: Update Chain State Values page
There are many issues with this page - https://wiki.polkadot.com/general/chain-state-values, likely due to the Asset Hub migration.

e.g. lots of values are set to 0.00 or default values because it's looking at the relay chain instead of AH

Example: "Polkadot's inactive issuance is 0.00 DOT in the era NOT_FOUND."

We also should include Asset Hub values here because those are now much more important than the relay chain for most account-based aspects (e.g. existential deposit).

CrackTheCode016:
Closes #6960 

Post-AHM, most of these pallets were moved to Asset Hub. The macro has been updated accordingly.

### DIFF
--- a/docs/general/chain-state-values.md
+++ b/docs/general/chain-state-values.md
@@ -21,7 +21,7 @@ description: Explore chain constants and storage values for Polkadot, Kusama, an
 
     #### Active Validator Count
 
-    The number of Polkadot validators in the active set is {{ rpc("polkadot-assethub", "Staking", "ValidatorCount", 297, false, false) }}.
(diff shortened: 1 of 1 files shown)


DrW3RK:
More changes have to be made. The scheduler needs to use the relay block for the testing to be successful. You can try running the tutorial code and check if it works.

CrackTheCode016:
Hi @DrW3RK, I believe you meant to comment on this PR? https://github.com/w3f/polkadot-wiki/pull/6965

And yes, that one is still in progress. I will tag for review once it is complete. Apologies if I misunderstood!

DrW3RK (approved):

DrW3RK:
My bad.  Let's merge this now

PR #7007, opened 2026-01-09
Remove ambassadors program

filippoweb3:
- This removes all programs
- add redirects
- rm internal links


filippoweb3:
@midegdugarova, we need to decide if we want to keep info about ambassadors in the wiki, I found the following latest info
- https://github.com/polkadot-phoenix/manifesto
- https://forum.polkadot.network/t/polkadot-ambassador-fellowship-paf-governance-critical-questions-before-on-chain-implementation/14997/10

It seems nothing is happening on-chain: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fdot-rpc.stakeworld.io%2Fcollectives#/ambassador

In my opinion, we should omit all this information unless there's some concrete development happening.

CrackTheCode016:
I would second @filippoweb3 here, I would only include info if it is: 

1. concrete / in place
2. actionable; has infrastructure to take action on it, and actually use it

filippoweb3:
@CrackTheCode016 then I would go ahead and merge this one. Old pages have been added to the archive repo.

CrackTheCode016 (approved):

PR #7040, opened 2026-02-17
Fix precise token formatting and add regression tests

afurm:
## Summary
This PR fixes a precision bug in token formatting used by RPC-backed wiki values and adds regression tests.

## Problem
`human_readable(..., rounded=False)` was not actually using full precision.
Because of that, `format(..., "precise_dot")` / `format(..., "precise_ksm")` returned truncated values (for example `1.23 DOT`) instead of full on-chain precision.

## Root Cause
In `macros/rpc.py`, the `rounded` argument was ignored when choosing `display_decimals`.

## Changes
- Updated `human_readable` in `macros/rpc.py` to use full precision (`decimals_amount`) when `rounded=False`.
- Kept existing rounded behavior unchanged (`2` decimals for DOT, `6` for KSM).
- Added regression tests in `tests/test_rpc.py` covering:
1. Default rounded DOT formatting.
2. Full-precision DOT formatting when `rounded=False`.
3. Full-precision small-value formatting (leading zeros case).
4. `format(..., "precise_dot")` output path.

## Validation
Ran:
`python3 -m unittest discover -s tests -v`

Result:
All tests pass.

## Impact
- Fixes accuracy for `precise_dot` / `precise_ksm` outputs in rendered docs.
- Low risk: change is isolated to formatting logic and covered by tests.

### DIFF
--- a/macros/rpc.py
+++ b/macros/rpc.py
@@ -49,11 +49,11 @@ def get_network_url(network):
         
 def human_readable(decimals_amount, number, rounded=True):
     balance_str = str(number)
-    
-    # Determine display decimals based on token type (derived from decimals_amount)
-    # For DOT (decimals_amount=10), show 2 decimals
-    # For KSM (decimals_amount=12), show 6 decimals
-    if decimals_amount == 10:  # DOT
+
+    # Keep full chain precision when rounded=False.
+    if not rounded:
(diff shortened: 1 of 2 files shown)


CrackTheCode016 (approved):
ty!

